PR #90, opened 2017-08-30
Patch fromverilog to handle 1-bit ports

leonardt:
@phanrahan can you explain why port.isinput()/isoutput() is flipped when querying the definition? My guess is that from the definition's perspective, an input to the module acts as an output (i.e. you can wire an input to an input of any instances). This is a slightly confusing for testing the module definition from verilog because we have to say:
```python
    if direction == "input":
        assert port.isoutput()
```

### DIFF
--- a/magma/fromverilog.py
+++ b/magma/fromverilog.py
@@ -40,7 +40,7 @@ def ParseVerilogModule(node):
         elif isinstance(io, Output):
             dir = Out
         else:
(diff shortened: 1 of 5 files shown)


phanrahan:
Right, inside the definition an output port appears as an input. That is because you need to wire up the output of an instance to it, and for this to work it must be an input. The situation is switched for input ports.

PR #94, opened 2017-08-30
Passes

phanrahan:
Start of a pass infrastructure. Seems to work. Not critical.


phanrahan:
All tests are passing. Pull it when you can.

PR #100, opened 2017-09-07
Updates for new clock types Reset and Enable.

phanrahan:
I added new clock types Reset and Enable to clock.py.

@leonardt could you check primitives.py and coreir.py. We should update these modules to use the new types.

Going forward, let's go with reset and enable. We can add set and clear latter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,5 @@ ENV/
 # Rope project settings
 .ropeproject
 tests/build
+
+.DS_Store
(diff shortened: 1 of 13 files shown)


leonardt:
Updated the simulator and primitives to use the new types. Coreir doesn't have a built-in enable/reset type so for now the coreir backend will just treat them as regular `Bit`s until that's implemented.

phanrahan:
I added a bunch of tests. Pull this when you are satisfied.

PR #105, opened 2017-09-21
Initial implementation of CircuitGenerator

leonardt:
Here's a first crack at a notion of a CircuitGenerator.
See `tests/test_circuit/test_circuit_generator.py` for example usage.

The basic pattern is to define a subclass of `CircuitGenerator` with:
* a class field `name` that is the prefix of the generated circuit's name.
* a method `definition` that creates a circuit definition. This definition can use `self.cached_name` to get a unique version of `name` based on the current arguments to the `definition` method.

For the example:
```python
class DefineAdd(CircuitGenerator):
    name = "Add"
    def generate(self, n, cin=False, cout=False):
        ...
```

`self.cached_name == "Add(n=8, cin=True, cout=False)"`

Thoughts?

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -499,15 +499,25 @@ GeneratorArguments = namedtuple('GeneratorArguments', ['args', 'kwargs'])
 class CircuitGenerator(object):
     @lru_cache(maxsize=None)
     def __new__(type_, *args, **kwargs):
+        # Create an instance of the generator
         inst = object.__new__(type_)
+
+        # Build a list of stringified parameters of the form "param=value"
         params = list(signature(inst.generate).parameters.keys())
         cached_args = []
         for value, param in zip(args, params):
             cached_args.append("{}={}".format(param, value))
         for key in sorted(kwargs):
             cached_args.append("{}={}".format(key, kwargs[key]))
-        inst.cached_name = "{}({})".format(inst.name, ", ".join(cached_args))
+
+        # Cached name is base_name + stringified parameters joined by commas
+        inst.cached_name = "{}({})".format(inst.base_name,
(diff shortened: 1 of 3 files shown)


leonardt:
It also stores `args` and `kwargs` in the generated definition in the field `__generator_arguments` as a namedtuple of the form `namedtuple('GeneratorArguments', ['args', 'kwargs'])`

leonardt:
Also added a reference to the original generator in the field `_generator`.

leonardt:
This was the best scheme I could come up with that allows us to automatically inject a `cached_name` by passing a reference to `self`.  Using a decorator would give us access to the arguments to generate a unique name, but we would either need to add an implicit argument like `cached_name` to every generator that isn't used when calling the generator. The subclass pattern gives us an implicit argument `self` that makes sense to Python programmers.

It's a little strange because to call a `CircuitGenerator` it looks like instancing an object, but the special `__new__` method actually returns the result of the `generate` function instead of the instance of `CircuitGenerator`, the usage pattern is identical to the old `def DefineCirc(...)` pattern, we just change the definition to be a subclass of CircuitGenerator.

phanrahan:
The coreir generator makes a lot of sense to me.

The second part where you use the generator to create a definition seems less clear to me. For example, you use two defines by one seems more like an instance, which is different than our current convention.

Another way of writing the code might be to subclass a generator that provides the interface declaration and have the subclass fill in the definition.

If I understand it, we will need to convert all the mantle code to this form. That's a lot of changes, so let's make sure its right so we don't have to do a major rewrite.

leonardt:
@phanrahan splitting up the interface declaration and definition makes sense. I will rework this and submit another PR with an updated implementation

PR #106, opened 2017-09-22
Circuit generator

leonardt:
Here is a revised implementation of circuit generators.

The pattern is now:
* subclass CircuitGenerator and implement an `interface` classmethod that generates the interface for a generated circuit given a specific set of generator arguments.
* subclass the subclass of CircuitGenerator and implement a `definition` method that generates the definition for a circuit given a specific set of generator arguments.

The current way to link a generator is pass a list of generator implementations (that have a `definition` method) to the compile command. It then finds any circuit instances that are generator instances and compiles a definition for them by looking up an implementation that is a subclass of the instance's original generator.

Perhaps some extensions would be:
* A generator could provide a default definition that can be overridden by subclasses, so the backend compile should check if there's a default implementation and go with that if there's no linked implementation. This would also enable instancing generator implementations directly.

Thoughts on this new interface?


leonardt:
I think a more viable solution for "linking generators" would be to have a pass that replaces instances of generated declarations with instances generated definitions.

PR #114, opened 2017-10-06
Change dot backend to try and use the names instances were declared with

shacklettbp:
I tested this with the pico40 processor, and it makes some circuits a bit easier to understand. Unfortunately some circuits (especially mantle ones), wind up with a bunch of LUT4's with names like "self" since they are never assigned directly to a variable, so the DebugName pass can't really find a meaningful name.

This changes the dot backend to print nodes with varname (magma name), for example: mycounter (inst0)

### DIFF
--- a/magma/backend/dot.py
+++ b/magma/backend/dot.py
@@ -1,5 +1,6 @@
 from ..array import ArrayKind, ArrayType
 from ..bit import BitType, VCC, GND
+from ..passes.debug_name import DebugNamePass
 from collections import OrderedDict
 
 from graphviz import Digraph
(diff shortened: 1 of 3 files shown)


phanrahan:
Yes, this name generation scheme is still not robust. Free beers to the first person who comes up with a better way to generate names!

leonardt:
@shacklettbp I actually started porting some of the simulator to passes and ran into the same python compatibility issues.

You can check out this implementation for 2/3 compatible use of inspect frames https://github.com/phanrahan/magma/blob/refactor-simulator-passes/magma/passes/add_debug_info.py

shacklettbp:
@leonardt Yes the only change I needed was indexing instead of by name.

PR #117, opened 2017-10-14
Improve naming to find arrays of instances and name accordingly

shacklettbp:


shacklettbp:
Based on my experiments with pico40, with this commit the remaining unnamed instances are the result of AnonymousCircuits, or instances which are never assigned to a variable (such as the pico40 InstructionDecoder).

PR #122, opened 2017-10-27
Migrate from `__eq__` to `is` for type equality

leonardt:
These changes remove the `__eq__` logic from magma `Type`s (Bit, Array, Bits, SInt, UInt) so that it can be overloaded by the mantle operators (i.e. `a == b` will create an `EQ` circuit). 

To enable this, we need to change `magma` logic to check Type equality using `is`. 

For `is` to work, we need to cache the type constructors so `(Array(2, Bit) is Array(2, Bit)) == True`. I've updated the magma logic/tests to reflect these changes. 

I wanted to announce these via a PR (a) so people are aware of this migration from `==` to `is` for type equality and (b) get feedback as to whether there's any other alternatives for supporting an overloaded `==` operator while still retaining the ability to check type equality. Let me know any comments/concerns, alternatives. One issue is that this makes checking type equality more brittle and prone to errors as I'm guessing the intuition is to use `==`, maybe a smart error message could help identify these cases.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -7,6 +7,11 @@ from .bit import Bit, BitOut, VCC, GND, BitType, BitKind
 from .bitutils import int2seq
 from .debug import debug_wire, get_callee_frame_info
 
+try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
+
 __all__  = ['ArrayType', 'ArrayKind', 'Array']
 
 #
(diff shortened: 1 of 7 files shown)


leonardt:
I think a better solution than this might be to have the user specify a block of code to have operators overloaded using a decorator or with block

i.e.
```
with mantle.operator_overload:
    ...
```
or
```
@mantle.operator_overload
def func(...):
    ...
```

PR #123, opened 2017-10-31
Miscellaneous Simulator and Frontend changes needed for CoreIRSimulator

shacklettbp:


shacklettbp:
I agree mdb really needs a test suite, since it was broken after the summer changes. I'll look into getting pytest to at least make sure it can successfully launch, but I also need to get the Hierarchical simulator finished since coreir issues have distracted from it.

PR #125, opened 2017-11-03
Frontend improvements

shacklettbp:
Adds the ability to access nested instances: counter.reg.O for example, as well as several other improvements and changes to be more consistent with python / coreir.

Also adds a simple frontend test, I will add more tests soon.

### DIFF
--- a/magma/simulator/mdb.py
+++ b/magma/simulator/mdb.py
@@ -19,8 +19,8 @@ __all__ = ['simulate']
 
 class DisplayExpr:
     idx = 0
-    def __init__(self, string, vars, scope):
-        self.object = eval(string, None, vars)
+    def __init__(self, bit, string, scope):
+        self.object = bit
         self.scope = scope
         self.string = string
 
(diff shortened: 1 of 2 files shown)


magma/simulator/mdb.py
```diff
@@ -30,6 +30,9 @@ def __init__(self, string, vars, scope):
     def display(self, value):
         print("\t{}: {} = {}".format(self.idx, self.string, value))
 
+def print_err(str):
+    print(str, file=sys.stderr)
+
```
leonardt: `magma` has a `logging` module that we're trying to move all the logging code to use (so that we can easily improve the logging infrastructure without having to change as much code). I think this can just use `logging.error`.

leonardt: Actually since `mdb`is a frontend that should own stdout, I think it's okay to have its own logging infrastructure if you think it will have different behavior.

shacklettbp: What about all the normal prints?

shacklettbp: I'll leave it for now then, we can always revisit. I'm not sure if there will be different behavior for the frontend in the future or not as far as error logging goes.

Issue #132, opened 2018-01-12
Issue with Counter Icestick Example


David-Durst:
A couple things have changed
First, in the first line of the second code box, instead of importing Adders, Add must be imported. 

Second, I think something has changed with the int2seq or m.wire since this was last run. When I run https://github.com/phanrahan/magma/blob/master/notebooks/icestick/counter.ipynb, I get the following error when compiling to Verilog. 

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-19a1998a7430> in <module>()
      1 from magma.backend.verilog import compile as compile_verilog
----> 2 print(compile_verilog(DefineCounter(4)))

<ipython-input-8-704a4853371c> in DefineCounter(n, has_ce)
     14 
     15     m.wire( reg.O, add.I0 )
---> 16     m.wire( int2seq(1, n), add.I1 )
     17 
     18     m.wire(add.O, reg.I)

/Users/durst/dev/W17-8/magma/magma/debug.py in wire(i, o, debug_info)
     22         if debug_info is None:
     23             debug_info = get_callee_frame_info()
---> 24         return fn(i, o, debug_info)
     25     return wire
     26 

/Users/durst/dev/W17-8/magma/magma/wire.py in wire(o, i, debug_info)
     29 
     30     # if o is an input
---> 31     if not isinstance(o, IntegerTypes) and o.isinput():
     32         # if i is not an input
     33         if isinstance(i, IntegerTypes) or not i.isinput():

AttributeError: 'list' object has no attribute 'isinput'

leonardt:
This should now be working with the latest versions of magma/mantle. You'll likely have to stash/nuke your changes to the notebook to get a fresh copy.

External support for `int2seq` has been dropped in favor of conversion commands such as `bits(...)` and `uint(...)`. I believe functions from `bitutils` are still being used internally in some places.

PR #135, opened 2018-01-14
fixing adder int2seq and simulator issues

David-Durst:
I fixed 2 issues with the adder icestick example:

1. I removed int2seq following the guidance of https://github.com/phanrahan/magma/issues/132
2. I'm importing python_simulator from the right location. It probably was moved at some point since this example was written.


leonardt:
Looks good to me

PR #139, opened 2018-03-01
Miscellaneous simulator and other fixes

shacklettbp:


shacklettbp:
Not really sure why this is failing, do these tests get run with the coreir dev branch or no?

leonardt:
Hm, this raises a bit of an issue with our current testing setup. Basically, the coreir-dev branch will get tested against coreir-dev, so that's why the `pr` build passes (because it runs the test on the target branch, which is coreir-dev), but since this branch isn't coreir-dev it will use master. At least we can tell that it passes on `coreir-dev` so I can merge it, but ideally there'd be some logic that ran any branches that are based off `coreir-dev` with the dev branch of coreir

PR #141, opened 2018-03-05
Convert mapN test to mapParallel (pull to coreir-dev)

David-Durst:


David-Durst:
This isn't going to work until I pull my changes in coreir's c++ component. I'll fix those.

PR #143, opened 2018-03-13
Add support for nested arrays to simulator frontend and CoreIRSimulator

shacklettbp:


leonardt:
Can't quite make sense of these changes, but @David-Durst can you confirm that this resolves your issue?

leonardt:
Ah I think I get it now, looks like general support for flattened nested arrays (as per your title!). You need to handle the renaming of the names ("_".join) as well as setting the value.

leonardt:
@David-Durst feel free to merge if you approve, looks good to me

David-Durst:
lgtm

PR #148, opened 2018-03-16
Magma Now Able To Use CoreIR Modules

David-Durst:
Please merge after merging https://github.com/leonardt/pycoreir/pull/35


magma/coreirModuleWrapper.py
```diff
@@ -0,0 +1,17 @@
...
+        name = coreirModule.name
+        IO = cirb.get_ports_as_list(cirb.get_ports(coreirModule.type))
+        wrappedModule = coreirModule
+    return ModuleWrapper
+
+def ModuleFromGeneratorWrapper(cirb: CoreIRBackend, namespace: str, generator: str,
+                               dependentNamespaces: list, genargs = {}, modargs = {}):
+    moduleToWrap = cirb.context.import_generator(namespace,generator)(**genargs)
+    cirb.context.run_passes(["rungenerators"], [namespace] + dependentNamespaces)
+    return DefineModuleWrapper(cirb, moduleToWrap)(**modargs)
```
leonardt: Minor nitpick, and this is probably annoying for your existing code, but can we change this module name to be `module_wrapper` (consistent with general Python style: https://www.python.org/dev/peps/pep-0008/#package-and-module-names) and can we maybe move it to a new submodule called `frontend`?

So using it would look something like
```
from magma.frontend.coreir import DefineModuleWrapper
```

David-Durst:
Moved module_wrapper code to the location requested.

magma/simulator/simulator.py
```diff
@@ -27,13 +27,21 @@ def set_value(self, bit, scope, newval):
         pass
 
     @abstractmethod
-    def advance(self, n):
+    def advance_cycle(self, cycles):
+        pass
```
leonardt: Adding this causes existing tests to fail since the PythonSimulator doesn't support these abstractmethods. I can update it to have them since it's simple, but in theory could we define these methods in terms of the abstract `advance` and `rewind` functions?

leonardt: I'll just merge and test this out myself

leonardt:
@David-Durst when you get a chance, can you test if https://github.com/phanrahan/magma/commit/67509bc47aef94725dc77061aa8e35938a74fb70 works for you?

David-Durst:
They changes work for me. Thanks for fixing that!

David-Durst:
Also, the reason that the changes from you and Brennan are being included in this pull request is that those changes had been made in master, but not in dev. This pull request is adding them to dev.

PR #150, opened 2018-03-16
Better CoreIR-Magma Conversions

David-Durst:
This was commented out. Was there a reason why? It makes debugging a lot easier.

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -23,10 +23,6 @@ class keydefaultdict(defaultdict):
             return ret
 
 def magma_port_to_coreir(port):
-    # rewrite here to use coreir port name that couldn't be used in python
-    if hasattr(port, "origPortName"):
-        port.name.name = port.origPortName
-
     select = repr(port)
 
     name = port.name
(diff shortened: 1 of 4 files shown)


leonardt:
Not sure, I see a reference to @osnr so he might remember. Also @phanrahan. Seems like a reasonable naming scheme to me, so unless those two have objections I'm okay with this change.

leonardt:
Including @shacklettbp too since this might affect naming in the simulator

David-Durst:
I've added some more stuff:

1. Better mapping from "in" coreir ports to "I" magma ports. previously, this didn't work on nested arrays. I've gone to the source of where the name string is generated to fix this.
2. Added a standard way to get a coreir module from a magma module (either compile or get the wrapped module)

David-Durst:
I've undone the instance naming change so that this can go through. Can we get these other changes in? While the better instance names is a nice to have (I'll push strongly for later), these other changes I need for aetherling to work.

leonardt:
@David-Durst that sounds good to me, you can submit a separate PR to track the instance name change specifically so we don't forget about it.

Merging this now.

PR #152, opened 2018-04-01
handling clk named type when converting coreir to magma modules

David-Durst:
I'm not sure how to handle named types in a general way. This hack handles the common case of clk. Any ideas for how to improve it?

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -3,13 +3,14 @@ import os
 from ..bit import VCC, GND, BitType, BitIn, BitOut
 from ..array import ArrayKind, ArrayType, Array
 from ..tuple import TupleKind, TupleType, Tuple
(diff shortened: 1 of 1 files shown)


leonardt:
Seems reasonable, I think something like this is the best we could do given that the types in coreir are just based on names.

Looking towards support for more named types, could use a dictionary of the form `"coreir_named_type": magma_type` and if it's a named type lookup the magma type in there.

David-Durst:
I added the dictionary. Also, I improved one error message. I hadn't wired up something and I was getting the error at https://github.com/phanrahan/magma/pull/152/files#diff-0c2fbe0c5bd0204f993fd3b6b97ee73eR201. I found this error to be difficult to understand, so I added more explanation

leonardt:
Thanks!

PR #163, opened 2018-04-09
Pass in BitVector rather than coreir.BitVector to match pycoreir change

shacklettbp:


shacklettbp:
Depends on https://github.com/leonardt/pycoreir/pull/42

PR #169, opened 2018-04-12
Make instances a property on a CircuitType

leonardt:
This prevents instances from inheriting the `instances` field as a class attribute
```python
>>> from mantle.coreir.arith import DefineAdd
>>> Add8 = DefineAdd(8)
>>> Add8.instances
[inst0 = coreir_add8()]
>>> add = Add8()
>>> add.instances
  File "<stdin>", line 1, in <module>
AttributeError: 'Add8' object has no attribute 'instances'
>>> type(add).instances
[inst0 = coreir_add8()]
```

The previous (undesired) behavior was
```python

>>> from mantle.coreir.arith import DefineAdd
>>> Add8 = DefineAdd(8)
>>> add = Add8()
>>> add.instances
[inst0 = coreir_add8()]
>>> type(add).instances
[inst0 = coreir_add8()]
```

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -423,14 +423,14 @@ class DefineCircuitKind(CircuitKind):
     #
     def place(cls, inst):
         if not inst.name:
-            inst.name = 'inst' + str(len(cls.instances))
+            inst.name = 'inst' + str(len(cls._instances))
             # osnr's suggested name
             #inst.name = 'inst' + str(len(cls.instances)) + '_' + inst.__class__.name
             #print('naming circuit instance', inst.name)
         #print('placing', inst, 'in', cls)
         inst.defn = cls
         inst.stack = inspect.stack()
-        cls.instances.append(inst)
+        cls._instances.append(inst)
 
 
 # Register graphviz repr if running in IPython.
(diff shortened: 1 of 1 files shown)


David-Durst:
Does that actually solve this? Doesn't it have to be a property of DefineCircuitKind? Since CircuitType is a superclass of the circuit definition, instances of the definition will get the field, right? By making it a property of DefineCircuitKind (not one attached to instances of DefineCircuitKind in the new call)

leonardt:
Is the new behavior shown above not what you expected?

This change is inside DefineCircuitKind, so i'm not sure why you're asking if it has to be property of DefineCircuitKind? (because it is)

Instances of the definition will get the `_instances` field, but by Python convention they shouldn't be accessing that because it is marked as an internal member
See https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles, particularly
> _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore.

We change it to use a double leading underscore to actually invoke name mangling to protect it

The instances do not inherit the property as shown by the above code example (in the original comment)

David-Durst:
Nevermind, I misunderstood what you did. I had thought you had directly made it a property of CircuitType. All good. I'm undoing the other change I made as it unnecessary, once this passes the build (with my change removed), all good

PR #174, opened 2018-04-13
Always Pull The Latest CoreIR Without Using Version Numbers

David-Durst:
It is annoying to have to update all my branches every time we increment the coreir version number. This fixes that.


David-Durst:
I based this off the wrong branch.

Issue #181, opened 2018-04-20
Feature Request: Simulator Nested Arrays Of Arrays


David-Durst:
What has to be done so that this is no longer true: https://github.com/phanrahan/magma/blob/9bc707d7d96a84b2e84aabaa9084dce66940f46e/magma/simulator/coreir_simulator.py#L181

David-Durst:
I've added a much simpler test case https://github.com/phanrahan/magma/blob/simulatorNestedArrays/tests/test_simulator/test_nested.py that also fails (in a different way) due to nested arrays in the simulator. This occurs in the simulator at https://github.com/phanrahan/magma/blob/master/magma/simulator/coreir_simulator.py#L158, after the compilation, so I think its a simulator issue and not an issue in my test. Teh 
```
numVertices(g)    = 24
ERROR: Cannot select I_0 From self
  Type: {'CLK':coreir.clk, 'I_0_0':Bit[8], 'I_0_1':Bit[8], 'I_0_2':Bit[8], 'I_1_0':Bit[8], 'I_1_1':Bit[8], 'I_1_2':Bit[8], 'I_2_0':Bit[8], 'I_2_1':Bit[8], 'I_2_2':Bit[8], 'I_3_0':Bit[8], 'I_3_1':Bit[8], 'I_3_2':Bit[8], 'O_0_0':BitIn[8], 'O_0_1':BitIn[8], 'O_0_2':BitIn[8], 'O_1_0':BitIn[8], 'O_1_1':BitIn[8], 'O_1_2':BitIn[8], 'O_2_0':BitIn[8], 'O_2_1':BitIn[8], 'O_2_2':BitIn[8], 'O_3_0':BitIn[8], 'O_3_1':BitIn[8], 'O_3_2':BitIn[8]}

0   libcoreir.dylib                     0x000000011043de42 _ZN6CoreIR8Wireable3selERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 290
1   libcoreir.dylib                     0x00000001103723bf _ZN6CoreIR9ModuleDef3selERKNSt3__15dequeINS1_12basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEENS6_IS8_EEEE + 895
2   libcoreir.dylib                     0x0000000110371c2f _ZN6CoreIR9ModuleDef3selERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 191
[...]

Issue #153, opened 2018-04-01
Detect magma/mantle branches based off coreir-dev in travis install script


leonardt:

leonardt:
We haven't been using this pattern.

PR #183, opened 2018-04-22
Fix CoreIR Simulator's handling of arrays nested more than one level deep

shacklettbp:
Contains a test from David


David-Durst:
This solves all the issues in https://github.com/phanrahan/magma/issues/181. @leonardt any idea why this is failing the build. It looks like more gold issues.

leonardt:
@David-Durst it's this issue: https://github.com/phanrahan/magma/issues/153 basically, the travis script doesn't realize this is a branch based off of `coreir-dev` so it should build/install coreir's dev branch. Instead it's using the release version

PR #190, opened 2018-05-16
Coreir linking minimal example

leonardt:
Summary of changes:
* coreir backend uses a default context for all instances of backends. This change should still work with any existing code that uses their own coreir backend objects that they pass around, but moves us towards a singleton coreir context for all of magma
* adds a hook to refresh the context, used for testing where each test should have a fresh coreir context (so no state is shared across tests), similar to our hook to clear the circuit cache for testing
* adds `DeclareCoreIRGenerator(lib : str, name : str, typegen = None)` which generates a `Define*(...)` function that acts like a normal python/magma generator. It uses `DefineCircuitFromGeneratorWrapper` by passing through `kwargs`.
* Adds stubs for `coreir_typegen` which will enable linking in generators which we don't currently have access to (so we can't run them to get the generated module and create a wrapper)

This travis push build will fail since it's not being run on the coreir-dev branch, but hopefully the PR build will pass since it will build on the coreir-dev branch

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,3 +9,4 @@ def magma_test():
     magma.config.set_compile_dir('callee_file_dir')
     magma_clear_circuit_cache()
     clear_cachedFunctions()
+    magma.backend.coreir_.__reset_context()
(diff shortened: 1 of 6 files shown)


leonardt:
@David-Durst one annoying change for you is that I reordered the `uniqueName` and `dependentNamespaces` parameters so I could make `dependentNamespaces` have a default value (of empty list).

PR #221, opened 2018-07-24
Begin migration of testing code to fault

leonardt:


leonardt:
Nope, they're used in the mantle test suite and i've updated those invocations.

PR #224, opened 2018-07-28
Preliminary support for if statements and ternary expressions

leonardt:
Supports basic nesting. Includes tests that compile some `Mux` declarations to verilog.

Things that aren't supported:
* Using anything other than an assignment statement in the if/else body
* Assigning to a variable only once in the if or else body (not both). We could support this if the variable is already defined in the enclosing scope, for example using a default value
    ```
    x = 3
    if S:
       x = 4
    ```
    This brings up another issue, which is that it doesn't support a default value. (So the above code would break even if `x` was assigned in the else block.
* If without an else (for the same reason as the above)

These changes introduce a decorator `@m.circuit_def`, which is replaces `@classmethod` by first doing a rewrite pass. This provides hooks for doing other syntactic rewrites.

Right now it uses `=` syntax to bind the result of the `Mux` that is generated for the statements. We could explore binding directly as an alternative to wiring (or by using `<=` syntax to bind to a port), but the existing mechanism required the least amount of changes (we bind a variable to a magma value, just like any other magma code, which is then wired up to a port).

Thoughts/feedback? Should I continue to widen support for the varying cases mentioned above before we merge? Or should we get this in so people can start playing around with it.


leonardt:
Added some documentation in `doc/circuit_definitions.md`. It also includes a note that the test can be any expression that evaluates to a magma Bit value, so something like `io.S` or `io.S[0]` or `eq(io.I0, io.I1)`. Basically, the rewrite rules change the expression into an input to the `Mux`, at which point the standard magma typechecker will kick in (enforcing this constraint). We could do a syntax level analysis to do a check, but this would require type inferencing rules. Another approach might be to catch an error in running the magma code, and a trace it back to the original if statement.

phanrahan:
Cool!

A couple of comments 

1. I suggest we name Mux2xn to just Mux2 if width is None.

2. I believe you suggested m.circuit.combinational for the decorator when
we met. I can imagine we will have more than one decorator, so it might be
worth doing this now.

leonardt:
Done

phanrahan:
Can you handle multiple assignments inside a block?

leonardt:
It takes the last assignment, and prints out a warning saying that you've assigned multiple times within the same block. It's a simple change to make it an error if we want.

PR #235, opened 2018-08-03
Support arrays of clocks

David-Durst:
Previously, when I did a map over an op with a clock input, it would fail as the clocks are all unwired. CoreIR is ok with wiring up arrays of clocks. However, Magma would fail unnecessarily. This fixes that so Magma is now also ok with arrays of unconnected clocks prior to a wireclocks pass.


leonardt:
Can we get a test for this case? Something that compiles a simple circuit with nested clocks to a coreir json would be fine.

leonardt:
I'll merge it so I can get your fix in and unblock you, pull the test in if you get a chance

PR #261, opened 2018-08-23
Fix wiring error

rsetaluri:
Fixed error message for arrays with different lengths.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -84,7 +84,7 @@ class ArrayType(Type):
             return
 
         if i.N != o.N:
-            report_wiring_error(f'Cannot wire {o.debug_name} (type={type(o)}, len={i.N}) to {i.debug_name} (type={type(i)}, len={o.N}) because the arrays do not have the same length', debug_info)  # noqa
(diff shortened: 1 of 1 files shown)


magma/array.py
```diff
@@ -84,7 +84,7 @@ def wire(i, o, debug_info):
             return
 
         if i.N != o.N:
-            report_wiring_error(f'Cannot wire {o.debug_name} (type={type(o)}, len={i.N}) to {i.debug_name} (type={type(i)}, len={o.N}) because the arrays do not have the same length', debug_info)  # noqa
+            report_wiring_error(f'Cannot wire {o.debug_name} (type={o.T}, len={o.N}) to {i.debug_name} (type={i.T}, len={i.N}) because the arrays do not have the same length', debug_info)  # noqa
```
leonardt: Good catch, I think perhaps it makes the most sense to return `type(o)` and `type(i)` instead of `.T` and `.N` since those will be both contained in the `type` string.

rsetaluri:
Done.

Issue #241, opened 2018-08-07
Add support for magma.Tuple to be a top level port


rsetaluri:
```python
import magma as m


class Foo(m.Circuit):
    IO = ["IFC", m.Tuple(I=m.In(m.Bit), O=m.Out(m.Bit))]

    @classmethod
    def definition(io):
        pass


if __name__ == "__main__":
    m.compile("foo", Foo, output="coreir")
    print (open("foo.json").read())
```

The above fails in the `compile()` call. The same holds for `output=verilog` but that's simply because `magma.Tuple` is simply not allowed in a verilog module declaration.

rsetaluri:
The underlying issue seems to be that magma top level ports need to have `isinput() or isoutput() == True` in the core ir backend.

phanrahan:
Currently magma modules can only have type Bit or Bits. Tuples or more general arrays are not supported.

leonardt:
We should be able to support this in the coreir backend. The verilog backend doesn't have support for basic Tuples yet, so this definitely won't work for that.

rsetaluri:
```python
import magma as m


class Foo(m.Circuit):
    IO = ["I", m.In(m.Tuple(a=m.Bit, b=m.Bit)), "O", m.Out(m.Tuple(a=m.Bit, b=m.Bit))]

    @classmethod
    def definition(io):
        pass


if __name__ == "__main__":
    m.compile("foo", Foo, output="coreir")
    print (open("foo.json").read())
```

To be clear the above **does work** because the top level types are In/Out.

rsetaluri:
@phanrahan: Is this true? I think Tuples and Arrays seem to be mostly working for the CoreIR backend. If it is the case, what is the purpose of these types? Just for internal use?

@leonardt: understood.

leonardt:
Support for more complex arrays (e.g. nested) and tuples in the verilog backend require flattening the types to normal verilog types (1-D arrays, wire). We haven't added that yet since we wanted to be careful about how to do this right since it requires munging names. Coreir has built-in support for those types so we can support it, and the corier-verilog backend has implemented name munging for generating those types.

rsetaluri:
To provide some more context:

```python
import magma as m

foo = m.DefineCircuit("Foo", "ifc", m.Tuple(I=m.In(m.Bit), O=m.Out(m.Bit)))
m.wire(foo.ifc.I, foo.ifc.O)
m.EndCircuit()

m.compile("foo", foo, output="coreir")
```

[...]

PR #271, opened 2018-08-30
Support multi-directional tuples in coreir backend

leonardt:
Fixes https://github.com/phanrahan/magma/issues/241

@rsetaluri check out the test in tests/test_coreir/test_coreir_tuple.py which checks compiling and also inspecting the interface/IO, the output files are
* tests/test_coreir/gold/test_multi_direction_tuple.json
* tests/test_coreir/build/test_multi_direction_tuple_instance.json (checks that instances with these multi directional tuples works as well, not just in the definition signature)

The issue was that the `parse` code in Interface was not properly flipping the tuple type. It assumes that the top level type is either an input or an output, and flips it for definitions. I changed the code to just use the `.flip()` method, which works for tuples with multi directional elements.

I also had to update the coreir backend to properly handle IO ports that may not be inputs or outputs at the top level.

### DIFF
--- a/tests/test_coreir/test_coreir_tuple.py
+++ b/tests/test_coreir/test_coreir_tuple.py
@@ -33,5 +33,17 @@ def test_multi_direction_tuple_instance():
     assert check_files_equal(__file__, f"build/test_multi_direction_tuple_instance.json",
                              f"gold/test_multi_direction_tuple_instance.json")
 
+
+def test_multi_direction_tuple_instance_bulk():
+    bar = m.DefineCircuit("bar", "ifc", m.Tuple(I=m.In(m.Bit), O=m.Out(m.Bit)))
+    foo_inst = def_foo()()
+    m.wire(bar.ifc, foo_inst.ifc)
+    m.EndCircuit()
+
+    m.compile("build/test_multi_direction_tuple_instance_bulk", bar, output="coreir")
+    # NOTE: Should be the same as previous test, so we use that as a check
+    assert check_files_equal(__file__, f"build/test_multi_direction_tuple_instance_bulk.json",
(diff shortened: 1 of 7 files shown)


tests/test_coreir/test_coreir_tuple.py
```diff
@@ -0,0 +1,37 @@
...
+                             f"gold/test_multi_direction_tuple.json")
+
+    assert [T.isinput() for T in foo.interface.ports["ifc"].Ts] == \
+           [False, True]
+
+    assert [T.isinput() for T in foo.IO.ports["ifc"].Ts] == \
+           [True, False]
+
+
+def test_multi_direction_tuple_instance():
```
rsetaluri: would another test like:

```python
    bar = m.DefineCircuit("bar", "ifc", m.Tuple(I=m.In(m.Bit), O=m.Out(m.Bit)))
    foo_inst = def_foo()()
    m.wire(bar.ifc, foo_inst.ifc)
    m.EndCircuit()
     m.compile("build/test_multi_direction_tuple_instance", bar, output="coreir")
    assert check_files_equal(__file__, f"build/test_multi_direction_tuple_instance.json",
                             f"gold/test_multi_direction_tuple_instance.json")
```

be helpful? (Note the difference of wiring `ifc.I` and `ifc.O` vs `ifc` itself.) Perhaps this is/should be tested elsewhere.

leonardt: I think I might as well add this as another test just in case (or for future refactors)

leonardt:
The way the code was written was intended to work with arbitrary nesting and mixing (since it's recursive), but I'll also add another test with 2-deep nesting just to make sure.

PR #272, opened 2018-08-30
Add option to split output files

rsetaluri:
Adding the option "split" in the magma compile command, which will pass
an option to coreir that splits the output into multiple files. The
usage is 'split=<output_dir>'.

Also refactored some of the magma.compile() command. Rather than
accepting a specific 'coreir_args' argument, we accept a general set of
options using kwargs. Also, removed the 'origin' argument is it was
unused.

The refactoring will make adding backend-agnostic args easier, e.g.

  if opts.get("inline", True):
    <some inline logic>

### DIFF
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -58,14 +58,40 @@ def check_definitions_are_unique(circuit):
     CheckDefinitionUniquenessPass(circuit).run()
 
 
-def compile(basename, main, output='verilog', origin=None, coreir_args=None):
-    if coreir_args is None:
-        # Make a fresh dict every time so we don't keep manipulating the
-        # singleton default argument dict
-        coreir_args = {}
+def __compile_to_coreir(main, file_name, opts):
+    # Underscore so our coreir module doesn't conflict with coreir bindings
+    # package.
+    from .backend import coreir_
+    backend = coreir_.CoreIRBackend()
+    backend.compile(main)
+    if opts.get("passes", False):
+        backend.context.run_passes(opts["passes"], ["global"])
+
+    backend.modules[main.coreir_name].save_to_file(file_name + ".json")
+    if opts.get("output_verilog", False):
+        # TODO(rsetaluri): Expose compilation to verilog in pycoreir rather than
+        # calling the binary from the command line.
+        lib_arg = ""
+        if backend.libs_used:
+            lib_arg = f"-l {','.join(backend.libs_used)}"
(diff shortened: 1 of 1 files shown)


rsetaluri:
Depends on https://github.com/rdaly525/coreir/pull/612

magma/compile.py
```diff
@@ -58,14 +58,40 @@ def check_definitions_are_unique(circuit):
...
+    backend.modules[main.coreir_name].save_to_file(file_name + ".json")
+    if opts.get("output_verilog", False):
+        # TODO(rsetaluri): Expose compilation to verilog in pycoreir rather than
+        # calling the binary from the command line.
+        lib_arg = ""
+        if backend.libs_used:
+            lib_arg = f"-l {','.join(backend.libs_used)}"
+        cmd = f"coreir {lib_arg} -i {file_name}.json"
+        if opts.get("split", ""):
+            split = opts["split"]
```
rsetaluri: I think it’s more natural to specify `split=True, output_dir=...` but that requires 2 extra opts rather than 1.

leonardt: Maybe support `split=True` and automatically generate an `output_dir` if `split` is a bool. To customize `output_dir`, then the user could pass the string to `split`. This is kind've the environment variable pattern (sometimes just set vs unset, sometimes the value is meaningful). Kinda odd though.

Another alternative that might be better is two opts: `split` and `split_dir`. If `split_dir` is set, it implies `split == True`. If `split` is set, the directory is automatically generated (or defaults to output_dir).

However, I think it's fine as it is, so no need to do this for now. Just throwing out some other options.

magma/compile.py
```diff
@@ -58,14 +58,40 @@ def check_definitions_are_unique(circuit):
...
+    from .backend import coreir_
+    backend = coreir_.CoreIRBackend()
+    backend.compile(main)
+    if opts.get("passes", False):
+        backend.context.run_passes(opts["passes"], ["global"])
+
+    backend.modules[main.coreir_name].save_to_file(file_name + ".json")
+    if opts.get("output_verilog", False):
+        # TODO(rsetaluri): Expose compilation to verilog in pycoreir rather than
+        # calling the binary from the command line.
```
leonardt: Opened an issue to track this here https://github.com/rdaly525/coreir/issues/613

rsetaluri:
Depends on rdaly525/coreir#612 has now passed, so merging.

PR #273, opened 2018-09-06
Escape dir name when splitting files

rsetaluri:
This fixes an issue, where if the directory where split files should go
already exists and contains verilog files, then those are picked up in
bash glob, rather than passing the string `"<dir>/*.v"` directly. Now,
when issuing the CoreIR command we escape this string.


leonardt:
We should consider adding some basic tests for this feature so it's covered.

PR #276, opened 2018-09-06
Fix 'wired to None' error message

rsetaluri:


magma/backend/coreir_.py
```diff
@@ -307,7 +307,8 @@ def is_clock_or_nested_clock(p):
         if value is None and is_clock_or_nested_clock(port):
             return
         elif value is None:
-            raise Exception("Got None for port: {}, is it connected to anything?".format(port))
+            raise Exception(f"Got None for port '{port.debug_name}', is it "
```
rsetaluri: we should probably consolidate all error reporting in the backends. In some places we use Exceptions, in others `error()`.

PR #279, opened 2018-09-07
Verilog inline

rsetaluri:
Make inlining verilog (through coreir) an option that can be specified.


magma/compile.py
```diff
@@ -80,6 +80,8 @@ def __compile_to_coreir(main, file_name, opts):
             cmd += f" -o \"{split}/*.v\" -s"
         else:
             cmd += f" -o {file_name}.v"
+        if "inline" in opts and opts["inline"]:
```
leonardt: How about `opts.get("inline", False)`

Issue #277, opened 2018-09-06
Add mechanism to convert bits to clocks when parsing verilog 


rsetaluri:
CoreIR type checks clock wiring; it does not allow wiring Bit to Clock. When parsing verilog everything is Bit/s. We either need to add an option `clocks=[]` to Declare/DefineFromVerilog which dictates that certain port names should be Clock not Bit (with appropriate checks that this conversion makes sense). Or we could add a lifting pass that allows conversion from Bit to Clock in general.

rdaly525:
Presumably when you are wiring to the lifted verilog, you know exactly where the clock is. Why not just explicitly convert from the clock to a bit while wiring?

rsetaluri:
That is an option that will not error, but is not identical to what I’m suggesting. Primarily because we get the extra typing of the wire being a clock (then we could even do the automatic clock wiring pass). Maybe those benefits are not worth it, but it seems like the right thing to do is convert the Bit to Clock, not the other way around.

rdaly525:
Yeah, it sounds like a good solution is then to pass in what are clock ports to the Declare/DefineFromVerilog in magma, then when you construct the wrapper coreir module, just use a clock port(s) in the type.

PR #280, opened 2018-09-07
Define from verilog clocks

rsetaluri:


rsetaluri:
See #277

magma/fromverilog.py
```diff
@@ -33,7 +34,16 @@ def visit_ModuleDef(self, node):
         self.nodes.append(node)
         return node
 
-def get_type(io):
+def convert(input_type, target_type):
+    if isinstance(input_type, BitKind) and \
+       isinstance(target_type, ClockKind) and \
+       input_type.direction == target_type.direction:
```
leonardt: A more general check might be that they're both instances of `_Bitkind`? That way, any sub type of `_Bit` can be converted to another sub type of `_Bit`.

leonardt: Also, we could do similar logic for array -> bits/uint/sint (check that they're both sub types of array)

rsetaluri: Done and tested.

PR #289, opened 2018-09-18
Add check for clock type in CoreIR compilation

rsetaluri:
Similar to what we do for AysncReset types, we make a special case to be
able to wire constants to clocks.

### DIFF
--- a/tests/test_type/test_clock.py
+++ b/tests/test_type/test_clock.py
@@ -2,8 +2,8 @@ import pytest
 import tempfile
 from magma import In, Out, Flip, \
     Clock, ClockType, ClockKind, \
-    Reset, ResetType, ResetKind, reset \
-    Enable, EnableType, EnableKind, enable \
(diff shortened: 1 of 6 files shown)


leonardt:
Looks good, any chance we can get a little test for this? I think the code in the issue should work.

rsetaluri:
Done

tests/test_type/test_clock.py
```diff
@@ -1,7 +1,12 @@
+import pytest
+import tempfile
 from magma import In, Out, Flip, \
     Clock, ClockType, ClockKind, \
-    Reset, ResetType, ResetKind, \
-    Enable, EnableType, EnableKind
+    Reset, ResetType, ResetKind, reset \
+    Enable, EnableType, EnableKind, enable \
```
leonardt: Missing end commas here I think

leonardt: Made the change via the github web ui, i'll wait for the build

PR #291, opened 2018-09-19
Fix zext

rsetaluri:


tests/test_type/test_conversions.py
```diff
@@ -105,3 +108,33 @@ def test_tuple():
...
+    else:
+        assert int(out) == value
+
+
+@pytest.mark.parametrize("value", [-5, 0, 10])
+def test_sext(value):
+    in_ = sint(value, 16)
+    # TODO(rsetaluri): Ideally, zext(sint) should return an object of type
+    # SintType, instead it returns an object of type ArrayType. For now, we wrap
+    # the result of zext() in sint().
```
leonardt: Can we open an issue on this? Basically the extension operators should preserve the input type.

magma/conversions.py
```diff
@@ -193,12 +193,12 @@ def zext(value, n):
...
     elif isinstance(value, BitsType):
         zeros = bits(0, n)
-    return concat(zeros, value)
+    return concat(value, zeros)
 
 
 def sext(value, n):
     assert isinstance(value, SIntType)
-    return sint(concat(array(value[-1], n), array(value)))
+    return sint(concat(array(value), array(value[-1], n)))
```
leonardt: @phanrahan can you review these changes? I think they look okay to me, but would be good to get a second set of eyes since these bit related ops can be tricky. From my understanding, `value[-1]` (so the last element in the array of bits) contains the sign bit, so to properly extend it, we need to concat the sign value on the end of the array. The previous implementation was concatenating it to the front of the array. This assumes that the concat logic matches the indexing logic, where the end values of concat are actually prepended (if you consider the most significant bits being in the front).

There's some wonkiness here, we'll want to review that the indexing logic and the concatenation logic are consistent

phanrahan:
Yes, is looks like there was a bug in the previous version. The bit should be extended in the most significant elements, which would be at the end of the array.

While we are making this change, perhaps we should change the semantics. I recall there was a request to interpret `n` as the final length, not the number of bits to extend.

rsetaluri:
That was my initial guess as to what the semantics were before testing/looking at the code. I think it is more natural to have `n` be the final length and not the number of bits to extend. Not sure what would break with that change, though.

rsetaluri:
Also, I noted CoreIR has a zext primitive. We could just make zext a mantle circuit and offload the semantics to CoreIR (thought this is a bit clunky syntactically)

rsetaluri:
In either case, I would argue getting in the minimal fix as this issue is blocking other work. I can create an issue tracking the semantics.

phanrahan:
I would vote for metaprogramming these operators.

Also, I didn't see zero_extend or sign_extend in coreir_prims.v.

rsetaluri:
I see a zext generator: https://github.com/rdaly525/coreir/blob/e2cd7be195d30eb3a1599e77b93f4abefc768c50/src/ir/headers/core.hpp#L65

And verilog expansion for zext: https://github.com/rdaly525/coreir/blob/e2cd7be195d30eb3a1599e77b93f4abefc768c50/include/coreir/definitions/coreVerilog.hpp#L46

https://github.com/rdaly525/coreir/blob/e2cd7be195d30eb3a1599e77b93f4abefc768c50/include/coreir/definitions/coreVerilog.hpp#L46

Not sure what this means in terms of placing these in magma/mantle however...

By metaprogramming do you mean something different from the function-like syntax that currently exists?

leonardt:
The current interface is done as metaprogramming. It does the extension logic in Python, so the magma circuit (graph) has a statically realized `zext` done in terms of wiring (and constants). Contrast this with instantiating a module to do the extension (so the extension logic is deferred and is abstracted within the graph).

leonardt:
I'm going to merge this to unblock Raj, I'll open an issue to adjust the semantics

rsetaluri:
@leonardt got it

PR #297, opened 2018-09-21
Emit verilog with coreir backend when using coreir output with coreir mantle target

leonardt:
This avoids unexpected behavior when trying to compile with the default output (verilog) and the default mantle target (coreir). If the user requests verilog when the mantle target is coreir, it will now use coreir to output verilog

### DIFF
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -1,3 +1,4 @@
+import sys
 import os
 import inspect
 import subprocess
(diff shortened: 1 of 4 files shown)


magma/compile.py
```diff
@@ -88,9 +90,13 @@ def __compile_to_coreir(main, file_name, opts):
 def compile(basename, main, output='verilog', **kwargs):
     opts = kwargs.copy()
 
+    mantle_imported = "mantle" in sys.modules
+
     # Rather than having separate logic for 'coreir-verilog' mode, we defer to
     # 'coreir' mode with the 'output_verilog' option set to True.
-    if output == 'coreir-verilog':
+    if output == 'coreir-verilog' or \
```
rsetaluri: It would be good to have some logging here that we are doing this implicit transformation of `output`. Also maybe we can move this logic outside of this `if` and above do:

```python
if output == "verilog" and m.mantle_target == "coreir" and mantle_imported:
    output = "coreir-verilog"
```

PR #307, opened 2018-10-19
Allow specifications of which modules to parse

rsetaluri:


magma/fromverilog.py
```diff
@@ -130,23 +130,17 @@ def FromVerilog(source, func, type_map, module=None):
...
-            if module is not None:
-                assert node.name == module
-                return circuit
             modules.append(circuit)
         except Exception as e:
             logger.warning(f"Could not parse module {node.name} ({e}), "
                            f"skipping")
-    if module is not None:
-        raise Exception(f"Could not find module {module}")
-
```
leonardt: This changes the behavior to return an empty list if there is no module found. Seems like reasonable behavior, but before it would raise an error if it couldn't find the module.

Either behavior seems like it could work, but maybe raising an exception if `target_modules` is not empty but no modules were found, might be helpful for the user (forcing them to handle this case, rather than returning an empty list which might make their code advance and cause a bug later down the line that may be harder to understand).

magma/fromverilog.py
```diff
@@ -107,7 +107,7 @@ def ParseVerilogModule(node, type_map):
 
     return node.name, args
 
-def FromVerilog(source, func, type_map, module=None):
+def FromVerilog(source, func, type_map, target_modules=None):
```
leonardt: Is it worth having both a single module argument and a list of modules argument? It would probably be convenient, but I'm not sure if there's an elegant way to do this without having to introduce some complexity in the code (handling the various cases). Maybe we can have a separate function that defines a single module from verilog (and then wraps the module in a list of one element and passes it to this function).

rsetaluri: Seems like reducing the complexity here and forcing the client to pass in a list and take the 0-th element of the list is preferable, esp. since it is likely not the most common case. This change also moves in the direction of thinking about bigger/general/more complex design scenarios where people might implement several modules in the same file etc.

magma/fromverilog.py
rsetaluri: Understood. Perhaps a warning would be best? The information is ultimately conveyed one way or another when an empty list is returned.

magma/fromverilog.py
leonardt: Makes sense, I agree.

magma/fromverilog.py
leonardt: warning sounds good, so if there is some issue caused, they can at least inspect the log and see it which might give them a clue as to where the issue is coming from

PR #308, opened 2018-10-22
Move isdefinition into separate file

rsetaluri:


leonardt:
Looks like github is having some issues related to webhooks, which probably is why we're not seeing a build: https://status.github.com/messages

rsetaluri:
Ah I see. I'll wait for tests to run again before merging.

PR #310, opened 2018-11-01
Attach verilog source when compiling to CoreIR

rsetaluri:
TODO(rsetaluri): Add tests.

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -298,6 +298,15 @@ class CoreIRBackend:
         if get_codegen_debug_info() and definition.debug_info:
             coreir_module.add_metadata("filename", json.dumps(make_relative(definition.debug_info.filename)))
             coreir_module.add_metadata("lineno", json.dumps(str(definition.debug_info.lineno)))
+
+        # If this module was imported from verilog, do not go through the
+        # general module construction flow. Instead just attach the verilog
+        # source as metadata and return the module.
+        if hasattr(definition, "verilog_source"):
+            verilog_metadata = {"verilog_string": definition.verilog_source}
+            coreir_module.add_metadata("verilog", json.dumps(verilog_metadata))
+            return coreir_module
+
         module_definition = coreir_module.new_definition()
         self.compile_definition_to_module_definition(definition, module_definition)
         coreir_module.definition = module_definition
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -302,8 +302,8 @@ class CoreIRBackend:
         # If this module was imported from verilog, do not go through the
         # general module construction flow. Instead just attach the verilog
         # source as metadata and return the module.
-        if hasattr(definition, "verilog_source"):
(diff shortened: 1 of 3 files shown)


rsetaluri:
@leonardt a few questions:
1. Should we treat `DefineFromVerilog` and `DeclareFromVerilog` differently? Only do this if it's using `Define`?

rsetaluri:
Note: depends on https://github.com/leonardt/pycoreir/pull/61 and https://github.com/rdaly525/coreir/pulls

magma/backend/coreir_.py
```diff
@@ -290,6 +291,15 @@ def compile_definition(self, definition):
         self.check_interface(definition)
         module_type = self.convert_interface_to_module_type(definition.interface)
         coreir_module = self.context.global_namespace.new_module(definition.coreir_name, module_type)
+
+        # If this module was imported from verilog, do not go through the
+        # general module construction flow. Instead just attach the verilog
+        # source as metadata and return the module.
+        if hasattr(definition, "verilog_source"):
+            verilog_metadata = {"verilog_string": definition.verilog_source}
```
leonardt: Noting that `verilog_source` was introduced as a temporary workaround for Priyanka to get her templated verilog files. Right now, the verilog backend expects a field `verilogFile`, which is only set on `DefineFromVerilog` calls (this is how it differentiates between `DefineFromVerilog` and `DeclareFromVerilog`. However, before we change this, we should resolve whether we want to differentiate the two.

leonardt:
The verilog backend does treat them differently. It will only emit the verilog if `DefineFromVerilog` is used. It seems that the use case for using Declare would be that the user would want to merge the verilog file in at a later stage in some different fashion (e.g. some build script). This was useful for when we couldn't pass verilog through the coreir backend (we use a declare, which allows us to compile the verilog referencing the external verilog module, then we splice the verilog file in using a build script). 

Now that we have that support, we should think about whether there's a use case for `DeclareFromVerilog` rather than `DefineFromVerilog`. In some sense, having it not include the verilog and emit just a reference to it in the final verilog might be useful for some more flexibility for the user. I can't come up with a good use case now, but maybe you want to be able to incrementally swap in multiple verilog files matching the declaration? This is similar to linking. Maybe your verilog file only has a declaration of the module (and no body), so you could use this to get a declaration to the module and have the definition filled in later for whatever reason.

rsetaluri:
Good points. I actually think there is still a good use case for `Declare`, e.g. if you have an existing flow that assumes there is a separate file and then just port one module from verilog to magma (and instance some other module parsed from verilog), then you just want to Declare it and not change your flow. I'll instead check for `verilogFile` rather than `verilogSource`

leonardt:
Looks like a minor conflict in coreir_.py introduced after the merge today, I think we can take both code snippets. Can you rebase and resolve it?

I'll merge this once we get a simple test and the functionality is merged into coreir master/release.

rsetaluri:
Added a test, and now everything passes.

PR #314, opened 2018-11-03
Verilog Operator Docs

leonardt:
These changes add a table to `docs/operators.md` which provides a reference for Verilog user. It lists all the operators detailed in the IEEE SystemVerilog specification and shows their verilog counterpart. 

Naturally, magma's operators are not feature complete, so I've attempted to note with TODOs where we could add missing operators (or corresponding mantle functions for Python operators that we can't overload), and other operators which I don't think we should plan to support (e.g. `dist` for constraints which is a verification feature).

This is mainly a documentation change, but it would be good to get a review on the TODO items and discuss what our plans for that are. If we decide to add them, we can open a separate github issue to track and discuss design.


phanrahan:
This looks good. I haven't had a chance to review all the TODOs. But here are some comments.

- Might want to show the type hierarchy. That is, make it clear that all the operators for Bits apply to SInt and Uint, for example.

- Array slicing with constants is allowed, but that corresponds to just bit selection.

- Do we need to mention the conversion functions. And when if any type restrictions there are on operator. Magma is different than verilog here.

PR #316, opened 2018-11-09
Fix coerir add_metadata calls

rsetaluri:
The new add_metadata API in CoreIR expects data parseable as json. We
now pass everything through json.dumps().


rsetaluri:
Note: https://github.com/phanrahan/magma/pull/317 needs to be merged before tests pass. Merging this PR first.

PR #322, opened 2018-11-29
Uniquification work

rsetaluri:


rsetaluri:
This is a first pass at a universal and robust way to do uniquification. Treat this more as a design proposal rather than  a final implementation -- let's iterate on this.

The semantics are as follows:
There is a static/global database of circuits per invocation of magma. Every circuit defined in that lifetime gets added to that database. If any 2 circuits have the same "name" then we compare the circuits to see if they are equal (right now we do this by comparing hashes of compiled coreir json -- we could theoretically use any equality mechanism that is appropriate). If they are "equal" then we map them to the same name. Otherwise they are guaranteed to have different names.

Here are some concerns I have with this method:
-- Not all valid magma circuits can be compiled to CoreIR (currently the important case is if outputs are not wired, this is a valid magma circuit but will not compile to CoreIR, though this is a general problem). For now, if something does not compile to CoreIR then we assume it is a new unique instance.
-- Renaming verilog is hard. We can either rely on rewriting the AST (more robust, less general) or do text replacement in the verilog source (not robust at all, general).
-- How does renaming interact with the user/debugging?

magma/backend/coreir_.py
```diff
@@ -414,7 +416,8 @@ def compile(self, defn_or_declaration):
         if defn_or_declaration.is_definition:
             self.compile_dependencies(defn_or_declaration)
             # don't try to compile if already have definition
-            if hasattr(defn_or_declaration, 'wrappedModule'):
+            if hasattr(defn_or_declaration, 'wrappedModule') and \
+               defn_or_declaration.wrappedModule.context == self.context:
```
leonardt: I think we might want to update (generalize) `wrappedModule` to be a map from context to coreir module. So in the current context, it should see that there is no wrapped module and recompile it.

magma/circuit.py
```diff
@@ -138,6 +144,27 @@ def __repr__(cls):
...
+        # <new_name>" existing anywhere else, most likely in comments etc. The
+        # more robust way to do this would to modify the AST directly and
+        # generate the new verilog code.
+        if cls.verilogFile:
+            find_str = f"module {old_name}"
+            replace_str = f"module {new_name}"
+            assert cls.verilogFile.find(find_str) != -1
+            cls.verilogFile = cls.verilogFile.replace(find_str, replace_str, 1)
+            # TODO(rsetaluri): Do we need this?
+            cls.verilog_source = cls.verilogFile
```
leonardt: `verilog_source` is set when declaring a circuit from template verilog. See https://github.com/phanrahan/magma/blob/7fd81ac42985e5f17ef447a100558bfa35de7754/magma/fromverilog.py#L136

`verilog_source` is set on circuits that are declared from verilog. I think the use case was a template generated verilog file, which was used to create a magma circuit declaration, then use the `verilog_source` field to dump the generated verilog to a file, to be merged in as part of the build script. This was before we could compile wrapped verilog definitions to coreir, so we could drop support for this attribute if the user can just use a DefineFromVerilog instead (removes build script complexity)

rsetaluri: We should be able to do this now with the way that we wrap verilog thru CoreIR.

magma/backend/coreir_.py
rsetaluri: It seems like this would only make sense if that map would contain more than one entry. Is that the case?

leonardt: It could be the case if we have multiple coreir contexts being used to compile in the same python interactive session. So in theory, yes. In practice, I'm not sure. We may want to redesign the use of context to avoid this issue entirely (I think it's simplest if everything uses the same context).

rsetaluri:
@leonardt what do you think about this change? I think some solution for uniquification will be useful to a lot of projects

PR #326, opened 2018-12-15
Fix issue with list of port declarations

leonardt:
This is a hotfix for an issue with importing verilog that uses a list of declarations of the following format
```
input clk_in, clk_en, reset, config_read, config_write, config_en,
       config_en_linebuf, wen_in, ren_in, chain_wen_in, flush;
```

In the old code, we were assuming that a declaration of the form `input <port_name>` only had one port, but in Verilog it's possible to list the ports as shown above. This patch updates the behavior to traverse all the children of the declaration AST node (part of the logic used for getting the types of ports listed in the module declaration).


alexcarsello:
This fix worked for me... thanks!

Issue #333, opened 2019-01-18
Issue with uniquifying circuits with mantle40 DefineSub


splhack:
This simple test code doesn't compile due to `KeyError: 'Could not find key: Invert2'` for ice40 board.

    import magma as m
    from mantle import *
    class TestSub(m.Circuit):
        IO = ['I', m.In(m.SInt(2)), 'O', m.Out(m.Bit)]

        @classmethod
        def definition(io):
            o = mux([m.bit(0), m.bit(1)], io.I - m.sint(0, 2) <= m.sint(0, 2))
            m.wire(o, io.O)

The error only happens with mantle40. So pure coreir and coreir-verilog work just fine.

    $ magma -o coreir -t TestSub testsub.py
    (no error)
    $ magma -o coreir-verilog -t TestSub testsub.py
    (no error)

With mantle40, it causes the error.

    $ magma -b tinyfpga -o coreir -t TestSub testsub.py
    magma:WARNING:Could not compile circuit: ''Could not find key: Invert2''. Uniquifying anyway.
    ...
        raise KeyError("Could not find key: {}".format(key))
    KeyError: 'Could not find key: Invert2'

The problem is generating `_Sub` class twice here https://github.com/phanrahan/mantle/blob/cf6c792a0dab0e6ba0245b04904bed8c36a0bdb1/mantle/lattice/mantle40/arith.py#L80-L99

One is for subtract by `io.I - m.sint(0, 2)`, the other one is for signed comparison `<= m.sint(0, 2)`.

- https://github.com/phanrahan/mantle/blob/cf6c792a0dab0e6ba0245b04904bed8c36a0bdb1/mantle/common/arith.py#L11
- https://github.com/phanrahan/mantle/blob/cf6c792a0dab0e6ba0245b04904bed8c36a0bdb1/mantle/lattice/mantle40/compare.py#L171

Both `_Sub` (`Sub2`) classes look identical, but they are not the same. They have different id and different hash values. And eventually goes to `BuildInstanceGraphPass` `self.graph[]` https://github.com/phanrahan/magma/blob/ce970a2a1930e224b865af6c040e2e80b94a96c3/magma/passes/passes.py#L70 and they will exist as different keys in the dictionary.

`tsort` doesn't work well in this case, one `Sub2` key has no edges. `Sub2` is populated before `Invert2`. This is the reason of `Could not find key: Invert2` error.

splhack:
[...]

PR #336, opened 2019-01-25
Fixes issues in the new uniquification logic

leonardt:
Major change is to add new `__hash__` and `__eq__` methods for circuits.
Given the new uniquification naming scheme, we can assume that names are
unique and use them for hashing and equiality checks.

Another major change is to use the default `magma_coreir_context`
instead of using a new context for the uniquification compiling logic.
The issue here was the use of wrapped circuits which were already
defined w.r.t. the default magma context.  There were issues related to
compiling multiple times within the same context, but they have been
fixed with https://github.com/rdaly525/coreir/pull/656 (the issue was
that the JSON pass not being re-run because the graph was not marked as
modified).

Updates the coreir backend to use proper syntax of logging.

Adds the ability to set the database backend (for mantle's verilog
target, it makes more sense to use the magma verilog backend, this
shouldn't affect most user code since the mantle verilog target isn't in
heavy use).

Update the transform logic to end the circuit definition at the end.
Before, it was calling EndCircuit right away (but then populating the
definition later) which was causing issues in the uniquification logic
(can't compile when EndCircuit was called because the definition wasn't
ready).

### DIFF
--- a/magma/circuit_database.py
+++ b/magma/circuit_database.py
@@ -4,6 +4,8 @@ import uuid
 from .compile import compile
 import coreir
 from .logging import warning
+from .config import get_database_hash_backend
+from magma.backend.coreir_ import CoreIRBackend
 
 
 class CircuitDatabaseInterface(ABC):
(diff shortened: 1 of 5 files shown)


leonardt:
Fixes #333

magma/circuit_database.py
```diff
@@ -35,12 +38,42 @@ def __init__(self, name):
...
+                            if key.coreir_lib:
+                                deps.add(key.coreir_lib)
+                            elif hasattr(key, 'wrappedModule'):
+                                deps |= key.coreir_wrapped_modules_libs_used
+                        backend.context.run_passes(["markdirty"], ["global"] + list(deps))
+                        string = open(tempdir + "/circuit.json").read()
+                    elif backend == "verilog":
+                        compile(tempdir + "/circuit", circuit, backend)
+                        string = open(tempdir + "/circuit.v").read()
+                        print(string)
```
rsetaluri: nit: let's get rid of this print

PR #355, opened 2019-02-14
Fix hash

rsetaluri:
Addresses https://github.com/phanrahan/mantle/issues/131


tests/test_uniquify.py
```diff
@@ -57,9 +57,46 @@ def test_uniquify_unequal():
...
+    assert hash(repr(foo)) != hash(repr(bar))
 
     m.compile("build/uniquify_unequal", top, output="coreir")
     assert check_files_equal(__file__,
                              "build/uniquify_unequal.json",
                              "gold/uniquify_unequal.json")
+
+
+def test_key_error():
+    import mantle
```
rsetaluri: For some reason I'm not able to reproduce the error this test targets without these mantle elements -- I don't think the issue (and therefore this test) really has anything to do with mantle, but it is sufficiently nuanced that I couldn't reproduce in a hand-crafted way w/o mantle.

@leonardt thoughts?

leonardt: The circuit combinational tests also depend on a `Mux` (for testing the generated muxes from if statements). What I did was stub out the `DefineMux` logic without a definition. Would that be sufficient for this test? Or does it need the actual implementation? Either way, I think it might be okay to copy the `Mux` definition from mantle and keep it the magma test suite as a common circuit. It's not essential to keep this up to date with mantle, it's more to serve as an example of a non-trivial circuit definition.

leonardt: Here's a link to the stub logic: https://github.com/phanrahan/magma/blob/master/tests/test_circuit_def.py#L24-L26

rsetaluri: Thanks for the pointers @lenny. It would be good to be able to mock mantle in general.

For this test I just pretty much copied from: https://github.com/phanrahan/mantle/blob/29ec32d76b2faf189525fcb5b5a845d817dcd9b2/mantle/coreir/MUX.py#L46-L119.

rsetaluri: I verified that the test is "valid" in that it causes the proper failure without this change.

PR #357, opened 2019-02-15
Add option for namespaces in magma.compile

rsetaluri:
This option is used when running passes in the CoreIR backend. The
default value is ["global"], when namespaces is unspecified.


David-Durst:
lgtm

PR #359, opened 2019-02-22
Keep original name for keyword argument

splhack:
I'm not sure whether there is a reason or not, but with this small example

    @m.circuit.combinational
    def test(a: m.Bit) -> m.Bit:
      return ~a

the argument name turns into with `_0` suffix, after recent ssa change.

    class test(m.Circuit):
        IO = ['a_0', m.In(m.Bit), 'O', m.Out(m.Bit)]

I assume I can use positional argument to call the combinational logic like this.

    b = test(io.a)

and also keyword argument like this.

    b = test(a=io.a)

but this keyword argument doesn't work due to `_0` suffix. I needed to change it into `test(a_0=io.a)`.

### DIFF
--- a/magma/ssa/ssa.py
+++ b/magma/ssa/ssa.py
@@ -39,8 +39,8 @@ class SSAVisitor(ast.NodeTransformer):
     def visit_FunctionDef(self, node):
         for a in node.args.args:
             self.args.add(a.arg)
-            self.last_name[a.arg] = f"{a.arg}_0"
-            a.arg = f"{a.arg}_0"
+            self.last_name[a.arg] = f"{a.arg}"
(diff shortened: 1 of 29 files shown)


leonardt:
The reason for this change was that it was the most elegant solution to handle the case where a variable in the definition matched a variable in the argument list.

There's two ways to address this issue: One is to update the ssa algorithm to appropriately handle inputs. The other is to have circuit.combinational "wrap" the generated ssa circuit with the original port names wired up to the ssa'ed port. The second version is an easier fix that addresses this issue (at the cost of one extra "wrapper" layer introduced in the generated code, I will work on it.

splhack:
fixed by #361

PR #361, opened 2019-02-22
Hotfix issue with circuit.combinitational renaming inputs ports

leonardt:
Addresses the issue in https://github.com/phanrahan/magma/pull/359 by using the `renamed_ports` class attribute to map the original port names to the ssa'ed port names


leonardt:
@splhack when you get a chance, can you try the fix on this branch? The circuit definition generated now has the original names, but this doesn't change the ssa algorithm so it should still properly handle the case when an input name matches a variable name in the body.

splhack:
confirmed this fixed the issue. thank you for working on this quickly!

PR #366, opened 2019-03-11
m.circuit.sequential compiles pe1

leonardt:
Still need to factor the tests in a way that doesn't depend on peak/mantle.


leonardt:
Closing in favor of lassen branch

PR #369, opened 2019-03-19
Initial lassen support

leonardt:
Changes required to compile lassen pe


.travis.yml
```diff
@@ -11,9 +12,23 @@ addons:
...
 - source .travis/install_coreir.sh
 install:
+# install conda for py 3.7
+- wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+- chmod +x miniconda.sh
+- ./miniconda.sh -b -p $TRAVIS_BUILD_DIR/miniconda
+- export PATH=$TRAVIS_BUILD_DIR/miniconda/bin:$PATH
+- hash -r
+- conda config --set always_yes yes --set changeps1 no
+- conda update -q conda
```
rsetaluri: nit: whitespace @ end of line

tests/test_syntax/_test_seq_pe1.py
```diff
@@ -0,0 +1,405 @@
+import magma as m
```
rsetaluri: Could we move this to peak and get rid of the fault dependency? If there's not a specific reason to leave this in magma, I think it would just mechanically (CI, environment) be easier to not have the magma -> fault dependency.

leonardt: Yes, that's why it's named `_test_seq_pe1`. The `_` marks it to be skipped by pytest. The full test is now part of lassen. I was keeping this file around as a reference, but we can remove it.

PR #381, opened 2019-04-18
Initial implementation of new style class definition syntax

rsetaluri:
This is a first pass attempt at a "new-style" class definition
syntax for magma circuits.  This removes the requirement that the user
specify the definition inside a `definition` classmethod. Instead,
the user constructs an `IO` object with the interface and then performs
the desired wiring operators to define the circuit.

The `IO` object constructs a set of anonymous values to be used for the
ports.  The circuit definition is defined in terms of the anonymous
values.  These changes add a final stage to the metaclass `__new__`
logic defined in the method `process_new_style_definition`.

The `process_new_style_definition` method constructs an "old-style" IO
list (e.g. `["I0", m.In(m.Bit), ...]`) and passes this to the old logic
for setting up interfaces (`DeclareInterface`).  This required the
minimal amount of changes to the existing code, although we may consider
rearchitecting the entire interface pipeline given this opportunity.

Because we use the old-style interface setup logic, we must traverse the
definition graph defined on the temporary anonymous values setup by the
`IO` object and replace the anonymous interface values with the concrete
values from the old-style interface setup logic.

`process_new_style_definition` also traverses the definition graph and
"places" the instances (by using `self.place`) within the current
definition.

### DIFF
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -70,6 +70,8 @@ from .product import Product
 
 from hwtypes.bit_vector_abc import TypeFamily
 
+from .generator import Generator
+
 
 BitVector = Bits
 UIntVector = UInt
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -561,6 +561,8 @@ class DefineCircuitKind(CircuitKind):
                 # For instance references, we place the instance in the current
                 # definition, then add the driver of its inputs to the queue
                 inst = curr.name.inst
+                if inst.defn:
+                    inst.defn.__unplace(inst)
                 self.place(inst)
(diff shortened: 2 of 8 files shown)


rsetaluri:
Issues:
- Ensuring new style and old style do not bleed into each other. I.e. it should be clear to the user that they are completely in mode, or the other
- Making sure instances are *not* placed twice (they are placed once correctly during the work queue algorithm, and once incorrectly at instantiation time in the normal placement flow)
- Thinking about adding ports to `IO` after creation.

rsetaluri:
Closing in favor of #576

PR #385, opened 2019-04-27
Sort instances during repr

leonardt:
This fixes an issue related to uniquification where two definitions with the same instance/connection structure are treated as two unique circuits because the order in which the instances are placed is different.

The workaround was to sort the instance list before generating the `__repr__` string, so both these circuits have the same `__repr__` result (so they hash to the same value and aren't uniquified). The other option is to use a separate "sorted repr" code path that uses this logic just for uniquification (if we want to preserve definition order in the repr string).

Basically, we can either (a) have repr use sorted instance ordering (by name) or (b) hash on a repr string with sorted instance ordering, but preserve definition ordering for `__repr__`.

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -118,12 +118,13 @@ class CircuitKind(type):
         if hasattr(cls,"instances"):
             s = '{} = DefineCircuit("{}", {})\n'.format(name, name, args)
 
+            sorted_instances = sorted(cls.instances, key=lambda x : x.name)
             # emit instances
-            for instance in cls.instances:
+            for instance in sorted_instances:
                 s += repr(instance) + '\n'
 
             # emit wires from instances
-            for instance in cls.instances:
+            for instance in sorted_instances:
                 s += repr(instance.interface)
 
             # for input in cls.interface.inputs():
--- a/tests/test_circuit/test_inspect.py
+++ b/tests/test_circuit/test_inspect.py
@@ -14,12 +14,12 @@ def test_str_repr():
     print(repr(Logic2))
     assert repr(Logic2) == """\
 Logic2 = DefineCircuit("Logic2", "I0", In(Bit), "I1", In(Bit), "O", Out(Bit))
-XOr2_inst0 = XOr2()
 And2_inst0 = And2()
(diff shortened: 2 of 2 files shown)


leonardt:
From the garnet issue:
> Waiting for feedback on a longer term solution, but that magma branch should generate a single PE tile, the issue was that two definitions of `SB_ID0_5TRACKS_B1_irq_res_p` had the same instances and wiring, but the instances were created in different orders. This means the repr string used for hashing had a different line ordering. The workaround avoids this issue by sorting the instances by name before generating the repr string (but this means the the repr string no longer has definition ordering). We might want to preserve definition order for `repr` (I don't know if anyone cares or uses this currently), if so we can change it to only use the sorted instances logic in the code path specifically for hashing/uniquification.

Kuree:
I don't think there is any code in `garnet/canal` that relies on the ordering of `__repr__`.

rsetaluri:
Do we have a sense of the performance impact of this?

Kuree:
Here is my experiment:
On branch `default-nodebug`
```
$ time python garnet.py -v --width 8 --height 4
real	0m58.168s
user	0m57.208s
sys	0m1.079s
```
Then
```
$ git cherry-pick 1e87228ce9cd390f6502b68ebd765b617208a609
$ time python garnet.py -v --width 8 --height 4
real	0m54.524s
user	0m53.724s
sys	0m0.984s
```

Notice that it's slightly faster. It's either because the variance in CPU time, or the unification actually makes fewer unique instances and thus faster. Without this fix, each PE tile is a unique module, i.e. `PE_Tile_unq` everywhere.

Given amount of time that PR's on `magma`/`peak` need to be merged in, I can write a git/bash script that pulls commits from different repo and branches and patch everything up so Alex can continue work on his PD flow.

leonardt:
@rsetaluri do we want to block this merge until we optimize for performance? or do you think it's okay to just merge for now given our garnet example actually runs faster

rsetaluri:
I think it's good to merge for now. If perf issues show up, I think we can think about how to solve it.

Kuree:
Can we merge it now?

PR #397, opened 2019-05-19
2x (almost) speed up for magma

Kuree:
`cProfile` comparison for `garnet` when compiling 32x16
Before this PR:
image
After this PR:
image

What happened here:
1. Lots of unnecessary conversation from set and back
2. Lots of unnecessary comparisons.

Solution:
- More efficient code to implement the same logic.

I think there are opportunities to achieve even more speed ups, e.g. another 2x at least. I will dig around more.

### DIFF
--- a/magma/port.py
+++ b/magma/port.py
@@ -1,3 +1,4 @@
+from magma.config import get_debug_mode
 from .logging import error, warning, get_source_line
 from .backend.util import make_relative
 from .ref import DefnRef, InstRef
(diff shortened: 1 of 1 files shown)


magma/port.py
```diff
@@ -59,6 +60,18 @@ def mergewires(new, old, debug_info):
         o.wires = new
 
 
+def fast_mergewires(w, i, o):
+    assert len(w.outputs) + len(o.wires.outputs) <= 2
+    w.inputs = i.wires.inputs + o.wires.inputs
+    w.outputs = i.wires.outputs + o.wires.outputs
+    w.inputs = list(set(w.inputs))
+    w.outputs = list(set(w.outputs))
+    assert len(w.outputs) <= 1
```
leonardt: Can we change this to something roughly similar to the old error message?

Basically something like `report_wiring_error(f"Connecting more than one output ({w.outputs}) to an input `{i.bit.debug_name}`")` (note we may have to augment report wiring error to have debug_info be optional).

The motivation to have this be an error message vs an assert is that it won't crash the compiler (so we can see other errors that may occur later on). This is similar to the behavior of a standard C compiler where it will show multiple error messages on different lines (rather than stopping at the first line with an error)

Kuree: Done.

PR #399, opened 2019-05-20
add plus to get_value

THofstee:


magma/fromverilog.py
```diff
@@ -81,15 +87,11 @@ def get_type(io, type_map):
...
     ports = []
-    for port in node.portlist.ports:
-        if isinstance(port, Ioport):
-            io = port.first
-            args.append(io.name)
-            args.append(get_type(io, type_map))
-        elif isinstance(port, Port):
-            ports.append(port.name)
-        else:
-            raise NotImplementedError(type(port))
```
leonardt: Was this port related code supposed to be removed? My suspicion as to why the tests are failing are because we still need to have `ports` and `args` populated

THofstee: oops.

THofstee:
Hmm, Travis isn't building?

THofstee:
@leonardt pytest passes locally for me, not sure what's up with Travis.

PR #403, opened 2019-05-23
Add ports

rsetaluri:
TODO: add tests

### DIFF
--- a/tests/test_editing.py
+++ b/tests/test_editing.py
@@ -1,0 +1,27 @@
+import magma as m
+
+
+def test_add_ports_basic():
+    Foo = m.DefineCircuit("Foo", "I", m.In(m.Bit), "O", m.Out(m.Bit))
+    m.EndCircuit()
+
+    Foo.add_port("I2", m.In(m.Bit))
+    assert isinstance(Foo.I2, m._BitType)
+
+    # Instance Foo *after* adding a port.
+    foo = Foo()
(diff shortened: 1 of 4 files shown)


tests/test_editing.py
```diff
@@ -0,0 +1,27 @@
...
+    m.EndCircuit()
+
+    # Instacne Foo *before* adding a port.
+    foo = Foo()
+    assert not hasattr(foo, "I2")
+    
+    Foo.add_port("I2", m.In(m.Bit))
+    assert isinstance(Foo.I2, m._BitType)
+
+    assert isinstance(Foo.I2, m._BitType)
```
leonardt: Could we add a compile test to make sure this logic interacts correctly with the backends? (e.g. there's not a data structure that the backend is using that we're forgetting to update)

rsetaluri: Done.

leonardt:
@rsetaluri what do you think about improving the logic to use weakref? I thought about it more, and this is mainly an issue when we add the notion of deleting instances (because then the class will hold the remaining reference), but in theory it could still manifest in the current code when a Circuit object goes out of scope (it will get destructed, releasing it's reference to it's placed instances, however those placed instances won't get cleaned up because their definition class still maintains a reference to them).

rsetaluri:
This is stale. We will revisit this feature in the future properly.

Issue #407, opened 2019-06-11
Reusing CoreIR Instance Across Unit Tests Gives Wrong CoreIR Module Signature


David-Durst:
This issue is based on the magma branch https://github.com/phanrahan/magma/tree/cache_coreir_context and mantle branch https://github.com/phanrahan/mantle/tree/cache_coreir_context where I use a singleton CoreIR context and CoreIR backend.

The singletons are causing issues.

When I individually run each of the tests in https://github.com/David-Durst/aetherling/blob/master/tests/test_delayed_buffer.py, they pass.

However, when I run them both at once with the command `pytest -s tests/test_delayed_buffer.py`, they fail. The error log is below.

The issue is that the singleton CoreIRBackend and CoreIR context are returning the incorrect module when [import generator is called](https://github.com/phanrahan/magma/blob/cache_coreir_context/magma/frontend/coreir_.py#L32). I've added code at https://github.com/phanrahan/magma/blob/cache_coreir_context/magma/frontend/coreir_.py#L34-L38 to compare what the singletons return compared to a fresh CoreIRBackend and CoreIR context.

The line in the below error log shows the difference. 'in' is the correct name, not 'in_0'. For the CoreIR module that is being wrapped, it is impossible for any of the ports to have the name 'in_0'.
```
Cached cirb IO: dict_items([('in_0', <coreir.type.Type object at 0x10cb3f7b8>), ('out', <coreir.type.Type object at 0x10cb3f390>)])
Fresh cirb IO: dict_items([('in', <coreir.type.Type object at 0x10cfd9630>), ('out', <coreir.type.Type object at 0x10d1ab2b0>)])
F
```

The error log is
```
Testing started at 13:05 ...
/Users/durst/anaconda/envs/aetherling/bin/python "/Users/durst/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/183.4886.43/PyCharm.app/Contents/helpers/pycharm/_jb_pytest_runner.py" --path /Users/durst/dev/W17-8/aetherling/tests/test_delayed_buffer.py -- -s
Launching pytest with arguments -s /Users/durst/dev/W17-8/aetherling/tests/test_delayed_buffer.py in /Users/durst/dev/W17-8/aetherling/tests

[...]

PR #408, opened 2019-06-12
CoreIR Context And Backend Singletons

David-Durst:
1. Added methods in the CoreIR frontend file for getting a singleton CoreIR context and backend
2. Replaced the global magma_coreir_context with a singleton object that wraps the context
    1. This fixed the issue with __reset_context noted https://github.com/phanrahan/magma/issues/407#issuecomment-501019630 and https://github.com/phanrahan/magma/issues/407#issuecomment-501040969. __reset_context() wasn't replacing all references to magma_coreir_context. The singleton object handles resets correctly. I know it handles resets correctly because I'm using the singleton CoreIR context and backend in Aetherling's unit tests.

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -18,6 +18,7 @@ from ..interface import InterfaceKind
 import inspect
 import copy
 import json
+from .. import singleton
 
 from collections import defaultdict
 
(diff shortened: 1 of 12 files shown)


magma/frontend/coreir_.py
```diff
@@ -18,8 +27,9 @@ def definition(cls):
 
 def DefineCircuitFromGeneratorWrapper(cirb: CoreIRBackend, namespace: str, generator: str,
                                       uniqueName: str, dependentNamespaces: list = [],
-                                      genargs: dict = {}, runGenerators = True):
+                                      genargs: dict = {}, runGenerators = True, print_error=False):
```
leonardt: Doesn't seem like this new `print_error` argument is used, can we drop this change?

magma/backend/coreir_.py
```diff
@@ -69,21 +70,26 @@ def magma_port_to_coreir(port):
...
+    def __init__(self):
+        self.__instance = coreir.Context()
+CoreIRContextSingleton()
 
 class CoreIRBackend:
     context_to_modules_map = {}
     def __init__(self, context=None):
         if context is None:
-            context = magma_coreir_context
+            context = CoreIRContextSingleton().get_instance()
```
leonardt: Since we're phasing out the notion of different contexts, perhaps we can add an assert here that the passed in context is the same as the singleton context? Alternatively, we could do a check and issue a warning (not sure how much code out there is relying on the ability to use their own coreir context)

magma/frontend/coreir_.py
David-Durst: Removed `print_error`.

magma/backend/coreir_.py
David-Durst: I think it's best to avoid an argument that must have one value at all times. I don't want to crash if the context is not the singleton context. There also may be uses for a non-singleton CoreIR context. 

I've added in a warning if using the non-singleton context and a flag for disabling the warning.

David-Durst:
@leonardt. I had to change tests/test_circuit/test_define.py and now the gold file tests are failing. Is there an automated way to update the gold files?

leonardt:
There's no fully automated way, but you could repurpose this script to get the jobs done: https://github.com/phanrahan/magma/blob/master/tests/update_golds.py (i.e. add the files you want to update to the list of pairs

PR #418, opened 2019-06-25
Rework logic for flattenning name

leonardt:
@David-Durst I was seeing other aetherling tests fail with the recent patch to the naming logic, so I rewrote the logic based on the fault implementation and I think it makes more sense, can you review it?

I think the issue is that the coreir simulator does not flatten array of bits (but it does flatten a tuple of bits), so the we needed to rework the check for when to flatten the name.

Also let me know if you're running into other/related test failures.

### DIFF
--- a/magma/simulator/coreir_simulator.py
+++ b/magma/simulator/coreir_simulator.py
@@ -8,6 +8,7 @@ from ..scope import Scope
 from ..ref import DefnRef, ArrayRef, TupleRef
 from ..array import ArrayType
 from ..tuple import TupleType
+from ..bit import BitKind
 from ..bitutils import int2seq
 from ..clock import ClockType
 from ..transforms import setup_clocks, flatten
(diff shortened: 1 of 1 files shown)


David-Durst:
Which Aetherling tests are you referring to? Are you running Aetherling locally?

 https://travis-ci.com/David-Durst/aetherling/jobs/210898011#L649 are failing because my mantle changes haven't been released yet. I had gotten my Magma changes in but hadn't gotten around to my Mantle ones.

David-Durst:
I just merged https://github.com/phanrahan/mantle/pull/151. Will that automatically do a new mantle release?

leonardt:
I'll trigger a release now, in the future you can trigger one by bumping the version number in setup.py and pushing a tagged commit

leonardt:
I was running `tests/test_native_2d_line_buffer.py ` on aetherling master locally using the aforementioned mantle test (and need these changes for those tests to pass)

PR #426, opened 2019-07-23
Named phi

cdonovick:
Adds utility function `gen_free_name`

changes `ssa`:
* `ssa` now takes an optional argument `phi` which can be either a function or a string
  * default value `"phi"`
  * if `phi` is a string then `ssa` will assume that such a name exists in the environment of the decorated function
  * if `phi` is a function then `ssa` will generate a new name and associate `phi` with the name in the generated code
* moves the stripping of decorators from `convert_tree_to_ssa` to `ssa`

### DIFF
--- a/magma/ast_utils.py
+++ b/magma/ast_utils.py
@@ -73,6 +73,34 @@ def inspect_enclosing_env(fn):
         return fn(enclosing_env, *args, **kwargs)
     return wrapped
 
+class NameCollector(ast.NodeVisitor):
+    def __init__(self):
+        self.names = set()
+
+    def visit_Name(self, node: ast.Name):
+        self.names.add(node.id)
+
+    def visit_FunctionDef(self, node: ast.FunctionDef):
+        self.names.add(node.name)
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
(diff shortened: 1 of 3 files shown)


magma/ast_utils.py
```diff
@@ -73,9 +73,44 @@ def wrapped(*args, **kwargs):
...
+def gen_free_name(tree: ast.AST, defn_env: dict, prefix: str = '__auto_name_'):
+    visitor = NameCollector()
+    visitor.visit(tree)
+    used_names = visitor.names | defn_env.keys()
+    f_str = prefix+'{}'
+    c = 0
+    name = f_str.format(c)
+    while name in used_names:
+        c += 1
+        name = f_str.format(c)
```
leonardt: Should we add a timeout (e.g. `if c >= 1000: raise Exception`). Unlikely to come up and would incur a performance cost, but you never know?

cdonovick: As python ints don't overflow don't see the need to give up.

leonardt: I just realized we can assume that the space of names is finite, so we will eventually find a name.

leonardt: (because the length of the program must be finite)

PR #433, opened 2019-08-22
Migrate to ast_tools ssa

leonardt:
Depends on https://github.com/leonardt/ast_tools/pull/11

Also, there's still some cleanup to do w.r.t. the handling of the environment/names (mainly that the combinational and sequential passes rely on injecting some names in to the environment, we should make sure we're being safe about how we're doing this or use a different strategy).


leonardt:
Still working through some issues in the peak/lassen flows so we should wait to pull until those repos are working with this branch.

magma/syntax/sequential.py
```diff
@@ -171,21 +174,19 @@ def __call__(self, I):
...
+class {circuit_name}(m.Circuit):
+    IO = {io_list}
+
+    @classmethod
+    def definition(io):
+        {register_instances}
+        @ast_tools.passes.end_rewrite()
+        @m.circuit.combinational()
+        @ast_tools.passes.begin_rewrite()
+        def {circuit_name}_comb({circuit_combinational_args}) -> ({circuit_combinational_output_type}):
```
cdonovick: Could we build the AST and directly call `m.circuit.combinational()`?  This would avoid needing to inject ast_tools into env.

magma/syntax/sequential.py
```diff
@@ -364,50 +369,28 @@ def _sequential(
...
-        return wrapped
-    else:
-        assert cls is not None
-        wrapped_sequential = ast_utils.inspect_enclosing_env(
-                _sequential,
-                decorators=[sequential]
-        )
-        combinational = m.circuit.combinational(decorators=[sequential])
-        return wrapped_sequential(True, cls, combinational)
+    if "DefineRegister" not in env:
```
cdonovick: I know this is a bit paranoid but all of these should should be injecting with a free name then passing the generated names around.  e.g.:
```python
def gen_register_instances(def_reg_name, initial_value_map, async_reset):
    register_instances = []
    for name, (value, type_, eval_type, eval_value) in initial_value_map.items():
        if isinstance(eval_type, m.Kind):
            if isinstance(eval_type, m._BitKind):
                n = None
                if isinstance(eval_value, (bool, int)):
                    init = bool(eval_value)
                else:
                    assert eval_value.name.name in ["GND", "VCC"], eval_value.name
                    init = 0 if eval_value.name.name == "GND" else 1
            else:
                n = len(eval_type)
                init = int(eval_value)
            register_instances.append(f"{name} = {def_reg_name}({n}, init={init}, has_async_reset={async_reset})()")  # noqa
        else:
            value = astor.to_source(value).rstrip()
            register_instances.append(f"{name} = {value}")
    return register_instances
```

magma/syntax/combinational.py
```diff
@@ -210,69 +120,41 @@ def visit_Name(self, node):
...
-        def wrapped(fn):
-            exec(_SKIP_FRAME_DEBUG_STMT)
-            nonlocal decorators
-            decorators = list(itertools.chain(decorators, [wrapped]))
-            wrapped_combinational = ast_utils.inspect_enclosing_env(
-                    _combinational,
-                    decorators=decorators)
-            return wrapped_combinational(fn)
-        return wrapped
+        phi_name = ast_tools.gen_free_name(tree, env,
```
cdonovick: Unnecessary see below

magma/syntax/combinational.py
```diff
@@ -210,69 +120,41 @@ def visit_Name(self, node):
...
+            mux_name = "mux"
+
+        def phi(cond, arg0, arg1):
+            if all(isinstance(arg, tuple) for arg in (arg0, arg1)):
+                return tuple(phi(cond, *zipped) for zipped in zip(arg0,
+                                                                       arg1))
+            return env[mux_name]([arg1, arg0], cond)
+
+        env.locals[phi_name] = phi
+        tree, env, metadata = ast_tools.passes.if_to_phi(phi_name).rewrite(
```
cdonovick: `if_to_phi` will generate a name for you with an optional prefix with the same logic that you have.  Just call:
```ast_tools.passes.if_to_phi(phi, 'magma_combinational_phi').rewrite(...)```

https://github.com/leonardt/ast_tools/blob/master/tests/test_if_to_phi.py#L23

https://github.com/leonardt/ast_tools/blob/master/ast_tools/passes/if_to_phi.py#L42

leonardt:
Closing in favor of https://github.com/phanrahan/magma/pull/668

PR #435, opened 2019-09-06
added memory to list of preloaded coreir libraries

rdaly525:
Fix to David's issue https://github.com/David-Durst/aetherling/issues/23
@David-Durst, could you add a test to make sure that you can access "memory.rom" ?


David-Durst:
I also needed to make this pull request: https://github.com/leonardt/pycoreir/pull/107

Why did I need both?

PR #443, opened 2019-10-03
Add basic for loop to combinational

leonardt:
It turns out, because for loops are ignored and passed through to the Python magma circuit definition, using them in some basic forms works, see the example.

There are some issues, for example you can't do something different based on the loop index, for example
```python
    n = 4
    @m.circuit.combinational
    def logic(a: m.Bits[n]) -> m.Bits[n]:
        O = []
        for i in range(n):
            b = a[n - 1 - i]
            if i % 2:
                b = Not()(b)
            O.append(b)
        return m.bits(O, n)
```
will actually get translated into
```python
class logic(m.Circuit):
    IO = ['a', m.In(m.Bits[n]), 'O', m.Out(m.Bits[n])]

    @classmethod
    def definition(io):
        O_0 = []
        for i_0 in range(n):
            b_0 = io.a[n - 1 - i_0]
            b_1 = Not()(b_0)
            b_2 = phi([b_0, b_1], i_0 % 2)
            O_0.append(b_2)
        __magma_ssa_return_value_0 = m.bits(O_0, n)
        O = __magma_ssa_return_value_0
        m.wire(O, io.O)
```
Notice that `Not()` is called regardless if the `if` statement is taken, because the SSA transformation doesn't preserve behavior w.r.t. to side effects. I think the solution here is to do a type inference pass, determine that `if` statement conditions are python or magma values. If it is a python value, don't do an ssa transformation and leave it as a python value (or the inverse, only do ssa on magma values).

We could use an explicit escape pattern, such as (the curly braces):
```python
    n = 4
    @m.circuit.combinational
    def logic(a: m.Bits[n]) -> m.Bits[n]:
        O = []
        for i in range(n):
            b = a[n - 1 - i]
            if {i % 2}:
                b = Not()(b)
            O.append(b)
        return m.bits(O, n)
```

Here's another version that explicitly escapes the for loop (curly braces quote magma code)

```python
    n = 4
    @m.circuit.combinational
    def logic(a: m.Bits[n]) -> m.Bits[n]:
        O = []
        with syntax as python:
            for i in range(n):
                {b = a[n - 1 - i]}
                if i % 2:
                    {b = Not()(b)}
                {O.append(b)}
            {return m.bits(O, n)}
```

### DIFF
--- a/docs/circuit_definitions.md
+++ b/docs/circuit_definitions.md
@@ -91,6 +91,31 @@ def return_magma_named_tuple(I: m.Bits(2)) -> m.Tuple(x=m.Bit, y=m.Bit):
     return m.namedtuple(x=I[0], y=I[1])
 ```
 
+## Basic for loops
+Basic for loops ranging over integer values are supported (they are ignored by
+the syntax transformations so they will be executed as standard Python/magma
+code). However, `if` statements are subject to an `ssa` transformation, so
+they use the above described form (you can only assign to values inside an if
+statement).
+
+This code:
+```python
+n = 4
+@m.circuit.combinational
+def logic(a: m.Bits[n]) -> m.Bits[n]:
+    O = []
+    for i in range(n):
+        O.append(a[n - 1 - i])
+    return m.bits(O, n)
+```
+
+produces this verilog
+```python
+module logic (input [3:0] a, output [3:0] O);
+assign O = {a[0],a[1],a[2],a[3]};
+endmodule
+```
+
 # Sequential Circuit Definition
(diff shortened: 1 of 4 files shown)


cdonovick:
I like the idea of curly braces to quote however `{return m.bits(O, n)}` is a SyntaxError.

I do however have a secret project to enable python grammar extensions which I have been working on while on flights.  I figure 2 more flights and it should probably in a functionally state.  If this is high priority I could I actually dedicate a work day to getting it finished.

leonardt:
Closing in favor of https://github.com/phanrahan/magma/pull/504

PR #461, opened 2019-10-18
Hwtypes

leonardt:
Finally, the rewrite of magma's type system to hwtypes is almost finshed.

I figured I'd announce this today with this PR since the test suite is now passing locally for me.  There's still quite a bit of work to cleanup the code, and we may want to move the array metaclass to hwtypes.

Sorry for the huge change set, but there was no way to do this incrementally since it was basically a rewrite of the entire type system.  I'm going to do a pass to make sure everything is as consistent as possible with the old way, hopefully no interfaces have been changed, but some things may be deprecated.  I'll also run fault and mantle test suites to see what coverage is missing in the magma test suite.

Some early feedback on the structure of the class/type system would be good, as there is some room for changes.

One major issue I've run into is the typing relation between parametrized types and their concrete instances.  For example, we used to say `isinstance(Array[3, Bit], ArrayKind)`. Now we could say `issubclass(Array[3, Bit], Array)` but I'm not sure if this is exactly what we want. The other option is to revert to the old `Kind` nomenclature, which might be best for preserving compatibility with existing code and being explicit (I originally made this change since we are using the 'Meta' naming scheme for Kinds, but we could just alias the metaclasses to kind for magma.  I don't feel strongly either way, so if I don't get input, I'll probably revert to the old `Kind` pattern.

I'll try to notify when this PR is ready for final review before merging, but until then, any time/eyes I can get on the code would be great. There was a lot of code to change, so I likely made mistakes, so please point them out!


magma/__init__.py
```diff
@@ -73,15 +75,13 @@ def set_mantle_target(t):
...
-SIntVector = SInt
-_Family_ = TypeFamily(Bit, BitVector, UIntVector, SIntVector)
-def get_family():
-    return _Family_
+# BitVector = Bits
+# UIntVector = UInt
+# SIntVector = SInt
+# _Family_ = TypeFamily(Bit, BitVector, UIntVector, SIntVector)
+# def get_family():
+#     return _Family_
```
leonardt: Still need to unify the family logic with hwtypes

leonardt:
Also, I took this as an opportunity to rename `_Bit` (parent of Bit, Clock, Reset, ...) to `Digital`, open to other names too (Could use Wire, but there's already a Wire object in the magma IR, or Signal)

cdonovick:
I'm out this weekend but will review Tuesday.

magma/bits.py
```diff
@@ -115,236 +78,135 @@ def wire(i, o, debug_info):
...
-        from .conversions import bits
-        result = super().__getitem__(key)
-        if isinstance(key, slice):
-            return bits(result)
-        return result
+    def __setitem__(self, index : int, value : AbstractBit):
+        raise NotImplementedError()
 
+    def bvnot(self) -> 'AbstractBitVector':
+        raise NotImplementedError()
```
leonardt: Still need to implement the operators as magma primitives. The plan is to use circuit declarations for them, and "link" in the definitions from a particular mantle implementation.

rsetaluri:
I kind of like moving away from using `Kind` objects as a way to test type relationships, and opting for the hwtypes style of doing this (using subclass and the abstract base classes). If it is possible to have this the pattern going forward and still have Kind's for backwards compatibility that seems optimal (not entirely sure what's required, but seems doable).

Also, does `Tuple` move to `Product`?

phanrahan:
I agree let's move array from Kind. I only did that because I did not know how to do proper subclass.

```
issubclass(Array[3, Bit], Array)
```
seems much nicer to meet.

phanrahan:
Changing `_Bit` to Digital seems good.  I think we discussed using `Signal` as the base class, with `Analog` and `Digital` as subclasses.

leonardt:
@rsetaluri Tuple is for "anonymous" tuples (no key/value mapping), Product is for key/value mapping (this matches hwtypes distinction). It's similar to tuple vs namedtuple in Python. Before, magma used Tuple for both anon and named variants

leonardt:
yes

leonardt:
hwtype has `__setitem__` as part of the abstract BV interface. Can we support this in magma?

Would we allow setting an item on an input? This could mean replacing the Bit wire/value with a new wire/value, e.g. `io.I[0] = 1` would change that index to a constant 1 (ignoring the value of the driver).  What about an output? Would that just be wiring up to that index `io.O[0] = 1`. I'm a little hesitant to support this as it seems to be inconsistent with our `=` for Python values and `<=` for wiring pattern.

We could support setting items for anonymous values, where we basically return a new anonymous value with the indices changed. I think this makes the most sense. Perhaps the sanest design is to only support this on anonymous values?

phanrahan:
I would love to get ride of `__setitem__`. 

Better is to encourage a functional view of arrays of bits. You can construct new arrays from other values, but modifying values after they have been formed into arrays has lots of issues.

rsetaluri:
I agree with @phanrahan, for his reason above but also because it becomes very confusing with the python = and magma <= semantics that are pretty clear in all other cases (as you mention).

leonardt:
Okay, I've added initial implementation/tests of the various operators.

The only missing operators are rotate left/right (rol/ror). This is a pretty complex circuit, are we expecting magma to generate a generic circular shifter for this operator? Alternatively, we could simply support rotating by a static amount, which is pretty simple (just slice the bits and concat them based on the rotate amount).

leonardt:
Also, currently `sext` is only defined on SInt, should we define it on `Bits`?

phanrahan:
I suggest `sext` only be defined on `SInt`.

phanrahan:
WRT to shifting by arbitrary amounts, that is a barrel shifter. I agree it is a complex circuit. I know there is an implementation somewhere.

I suggest we proceed without it, and then add it later (at least until I find my implementation).

cdonovick:
> The only missing operators are rotate left/right (rol/ror). This is a pretty complex circuit, are we expecting magma to generate a generic circular shifter for this operator? Alternatively, we could simply support rotating by a static amount, which is pretty simple (just slice the bits and concat them based on the rotate amount).

This would break from other implementations of BitVector.   I'd rather remove the methods from `AbstractBitVector` than have partial implementations in magma.  They are there because they are part of SMT lib but they are in SMT lib as a convenience functions so I see no issue in removing them.

phanrahan:
I found the implementation of the barrel shifter. Let's add the rotates and shifts for completeness.

cdonovick:
> I suggest `sext` only be defined on `SInt`.

I somewhat disagree.  `sext` is well defined in an unsigned context and there are bit twiddling use cases for sext for example:
```python
def mux(sel, b0, b1):
    n_sel = ~sel
    size = b0.size
    return (sel.sext(size - 1) & b1) | (n_sel.sext(size - 1) & b0)

File: magma/array.py [hunk 3]
```diff
@@ -1,124 +1,163 @@
...
-        return t
+        orig_name = cls.__name__
+        class_name = '{}[{}]'.format(cls.__name__, index)
+        type_ = mcs(class_name, bases, {"orig_name": orig_name}, info=(cls, ) + index)
+        type_.__module__ = cls.__module__
+        mcs._class_cache[cls, index] = type_
+        return type_
+
+    @property
+    def abstract_t(cls) -> 'BitMeta':
```
cdonovick: Type comment is wrong I assume

magma/array.py
```diff
@@ -1,124 +1,163 @@
...
-                (issubclass(cls.T, Array) and cls.N == 1):
-            arg = args[0]
-            if len(arg) < len(result):
-                from .conversions import zext
-                arg = zext(arg, len(result) - len(arg))
-            result(arg)
-        return result
+    def __eq__(cls, rhs):
+        if not isinstance(rhs, ArrayMeta):
+            return False
```
cdonovick: Should return `NotImplemented`.

rdaly525:
I agree with Caleb about sext. SMT defines a sign extend operation on bitvectors and we have that in AbstractBitVector.

magma/array.py
```diff
@@ -1,124 +1,163 @@
...
+                    for elem in args[0]:
+                        if isinstance(elem, int):
+                            self.ts.append(VCC if elem else GND)
+                        else:
+                            self.ts.append(elem)
+                elif isinstance(args[0], Array):
+                    if len(args[0]) != len(self):
+                        raise TypeError(f"Will not do implicit conversion of arrays")
+                    self.ts = args[0].ts[:]
+                elif isinstance(args[0], int):
```
cdonovick: This constructor seems to assume that we are defining `Array[N, Bit]`

magma/bits.py
```diff
@@ -106,245 +85,522 @@ def __int__(self):
...
+    def sext(self, other) -> 'AbstractBitVector':
+        ext = int(other)
+        if ext < 0:
+            raise ValueError()
+
+        T = type(self).unsized_t
+        return self.concat(T[ext]([self[-1] for _ in range(ext)]))
+
+
+class BFloat(Bits):
```
cdonovick: Is this still needed?

magma/t.py
```diff
@@ -1,28 +1,40 @@
...
+# From http://code.activestate.com/recipes/391367-deprecated/
+def deprecated(func):
+    """This is a decorator which can be used to mark functions
+    as deprecated. It will result in a warning being emmitted
+    when the function is used."""
+    def newFunc(*args, **kwargs):
+        warnings.warn("Call to deprecated function %s." % func.__name__,
+                      category=DeprecationWarning)
+        return func(*args, **kwargs)
+    newFunc.__name__ = func.__name__
```
cdonovick: Should just use `functools.wraps`

magma/t.py
```diff
@@ -1,28 +1,40 @@
...
-__all__  = ['Type', 'Kind']
-__all__ += ['In', 'Out', 'InOut', 'Flip']
+
+# From http://code.activestate.com/recipes/391367-deprecated/
+def deprecated(func):
+    """This is a decorator which can be used to mark functions
+    as deprecated. It will result in a warning being emmitted
+    when the function is used."""
+    def newFunc(*args, **kwargs):
+        warnings.warn("Call to deprecated function %s." % func.__name__,
```
cdonovick: Probably want `warnings.warn(msg, category, stacklevel=2)` which makes the warning appear at the call site.

magma/tuple.py
```diff
@@ -1,14 +1,14 @@
 from collections.abc import Sequence, Mapping
```
cdonovick: We should discuss the proper way for Magma to interact with ADT types.  In my mind subclassing is not correct.

magma/tuple.py
```diff
@@ -168,58 +189,145 @@ def const(self):
...
+    @classmethod
+    def keys(cls):
+        return cls.field_dict.keys()
+
+    @classmethod
+    def types(cls):
+        return cls.fields
+
+
+# class TupleKind(Kind):
```
cdonovick: Remove commented code

phanrahan: I'm ok with `sext` on all bit vectors.

magma/bits.py
rsetaluri: I think it is still used, but IMO should go in some external files which extend the type system.

magma/array.py
leonardt: Fixed

magma/array.py
leonardt: Fixed

magma/array.py
leonardt: Fixed

magma/t.py
leonardt: Fixed

magma/t.py
leonardt: Fixed

magma/tuple.py
leonardt: Fixed

magma/bits.py
leonardt: @rsetaluri do you mean that the user code defines this subclass?  Or just to move this to a separate file?

The logic for the BFloat operators is currently handled in the mantle implementations (e.g. the DefineAdd generator dispatches on the type, if it sees a BFloat, it generates a coreir float.add), I think this operator definition will now be moved to this type definition, but could either live in core magma or be moved to user code.

magma/tuple.py
leonardt: We have discussed various methods for rearchitecting the magma internals such that subclassing the ADT types will not be necessary, I will be trying these out to see if it is possible.

magma/bits.py
rsetaluri: at very least it should go to `bfloat.py`...think it's ok to live in core magma then (not sure where else it could go)

leonardt:
@cdonovick these changes have been updated so that subclassing the ADT types is no longer necessary.  I still need to work on other parts of the ecosystem (mainly mantle and fault) to migrate them to this branch, and I also need to write a migration guide as this will require changes to user code.  It would be helpful if you could try using it for your needs (e.g. see if it can compile your peak programs removing the family closure, try integrating the assembled ADT logic).  There was a lot of code touched so we should test this thoroughly before getting it into the mainline

rdaly525:
@leonardt if I were able to show that Peak RTL generation works with this branch, what is your time estimate for integrating this branch into master?

leonardt:
That’s promising, very soon. I’d probably need a day or two to push it through the mantle and fault test suites then I’d be okay to merge

leonardt:
https://github.com/phanrahan/mantle/pull/161 contains the changes for mantle tests to pass with this branch, fault is next but should be fairly quick.

rdaly525:
@leonardt is it possible to use hwtypes.Enum instead of the magma Enum?

leonardt:
Not yet, looking at the code, I think it might be tricky since Enum values don't have a type.

@cdonovick is it possible to adjust Enum values to be an instance of BitVector? That would simplify magma support.  Basically, referencing enum in the body of a definition should work fine, e.g. `io.I0 == MyEnum.A`, but in order to use an Enum as an input type, we need to know what it corresponds to, so a `In(MyEnum)` would ideally be regarded as a `Bits` input of the appropriate width.  As a temporary workaround, the input type be described as BitVector, or perhaps we could add support for an assembled enum

leonardt:
Closing in favor of the `hwtypes2` branch which reverts to the old logic before introducing the Wire object (as per discussion with @cdonovick where we will be using a different approach for supporting peak rtl)

PR #462, opened 2019-10-19
Parse n-d array types in verilog ports

rsetaluri:


rsetaluri:
To actually work and pass tests, this depends on https://github.com/PyHDI/Pyverilog/pull/51 (however, the code is guarded to be backwards compatible)

leonardt:
Just saw your follow up comment, not sure how fast we can expect that PR to get merged, one option is to add a requirements.txt back to magma that points to our fork of pyverilog

rsetaluri:
@leonardt I've added back requirements.txt to point to leonardt/pyverilog@add-nd-arrays

rsetaluri:
There is an issue of packed arrays vs. unpacked arrays in verilog. See http://www.verilogpro.com/systemverilog-arrays-synthesizable/. However, in magma we do not distinguish between these. @rdaly525 any thoughts on how to deal with this? Both from the parsing and generation side.

PR #464, opened 2019-10-23
Prototype logic for automatically collecting instance name

leonardt:
This PR shows where we can insert logic to automatically determine the name of a magma instance using `inspect`.

Basically, when placing an instance in a definition, before naming it checks if we have debug mode enabled.  If so, we know that we have access to the filename/lineno where the instance was created. Then, we can parse the line to try to determine the name of the instance.  This works well for simple cases of the form `<name> = <instance>()`.  It also handles the case when a name is reused (by simply appending a counter value to the end of the name), so this should work when instances are created in a loop.

More complex forms might be supportable, e.g. `(<name_0>, <name_1> = <instance_0>(), <instance_1>)`.

One issue is how to handle when multiple instances are chained in the same line, e.g. `x = F()(G()(io.I))`. In this case, the instances of F and G would be named `x_1` and `x_0`. This should still work, but we may consider doing an SSA transformation on the original line, and introduce temporary variables/instance names for the "anonymous" instances, so it's more clear that these were unnamed instances.

CC @Kuree

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -489,35 +489,38 @@ class DefineCircuitKind(CircuitKind):
     def instances(self):
         return self._instances
 
+    def inspect_name(cls, inst):
+        # Try to fetch instance name
+        with open(inst.debug_info.filename, "r") as f:
+            line = f.read().splitlines()[inst.debug_info.lineno - 1]
+            tree = ast.parse(textwrap.dedent(line)).body[0]
+            # Simple case when <Name> = <Instance>()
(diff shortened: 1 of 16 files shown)


leonardt:
It's not more expensive than the existing debug logic which requires inspecting the stack to get the calling frame during instancing/wiring. One flag could be to enable the inspect logic on instancing, but not wiring. If the stack information has been inspected, we just parse the line pointed to by filename/lineno, which should be a marginal cost assuming the complexity of the line of code is sane.

So the cost is getting the filename/lineno of the instance location, which is currently guarded by debug mode.

PR #466, opened 2019-10-24
update reset logic to support asyncresetn

leonardt:
This extends the features added by @jameshegarty to include AsyncResetN in the clock interface logic, automatic clock wiring logic, and coreir backend logic.

* Adds a parameter `has_async_resetn` to `m.ClockInterface` (enables us to use consistent interfaces for asyncresetn)
* Adds support for automatically wiring asyncresetn types (similar to how we wire clocks/resets/enables automatically)
* Adds support for compiling asyncresetn to coreir (coreir uses a namedtype for reset/asyncreset which we have to convert to when compiling magma's reset types)

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -182,7 +182,8 @@ ClockTypes = (ClockType, ResetType, AsyncResetType, EnableType)
 
 
 def ClockInterface(has_enable=False, has_reset=False, has_set=False,
-                   has_ce=False, has_async_reset=False):
+                   has_ce=False, has_async_reset=False,
+                   has_async_resetn=False):
     args = ['CLK', In(Clock)]
     has_enable |= has_ce
     if has_enable:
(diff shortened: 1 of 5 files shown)


magma/clock.py
```diff
@@ -182,7 +182,8 @@ class EnableType(_BitType):
 
 
 def ClockInterface(has_enable=False, has_reset=False, has_set=False,
-                   has_ce=False, has_async_reset=False):
+                   has_ce=False, has_async_reset=False,
```
rsetaluri: I think it would be good to decouple `has_reset` from `reset_type` (those could be the exact parameters to this function). Seems like you wouldn't want to have multiple resets anyway (perhaps I'm missing some use case). Also, this would change the interface of this function, so current users may be affected.

phanrahan:
Do we have any tests for these types?

@rsetaluri's comment is a good one. Are there use cases where you have more than one reset type? @rdaly525?

leonardt:
This change set adds tests for the async reset n type that were missing (tests the code path that was fixed by this).  I agree that an overhaul of this to simplify the interface would be good, but let's do that in a separate change set (this is a hotfix for existing user code).  I think we should be able to make the change while supporting existing user code (e.g. we can deprecate the has_async_reset and has_async_resetn parameters, but just have them set the `reset_type` parameter for now).

PR #468, opened 2019-10-28
Add documentation for values

leonardt:
This initializes the cheat sheet doc (inspired by https://github.com/freechipsproject/chisel-cheatsheet) with documentation on magma values.  Note that I discovered a weird issue with how values currently work, which is included in the documentation.  This should be fixed by the `hwtypes` branch, so instead of working on a patch (which requires a non trivial change to how values work), I think effort is best put into getting the `hwtypes` branch in which will fundamentally change how values work.

### DIFF
--- a/docs/cheat_sheet.md
+++ b/docs/cheat_sheet.md
@@ -1,0 +1,52 @@
+**NOTE** This is a work in progress, please let us know via issue/gitter/email
+if you'd like to see anything added to this.
+
+This is inspired by https://github.com/freechipsproject/chisel-cheatsheet and
+will be rendered in a similar single page layout soon.
+
+# Basic Magma Constructs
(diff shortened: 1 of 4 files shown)


docs/cheat_sheet.md
```diff
@@ -0,0 +1,52 @@
...
+    def definition(io):
+        # Allocate `x` as a value of type `Bits`
+        x = Bits[5]()
+        # Wire io.I to x
+        x <= io.I
+        # Wire x to io.O
+        io.O <= x
+```
+
+**NOTE** Currently magma only supports wiring two intermediate temporary values
```
rsetaluri: btw, is this a `<=` issue or a more general `wire()` issue?

PR #470, opened 2019-10-28
Check for unconnected ports at definition time

rsetaluri:


rsetaluri:
One question is how customizable to make this warning. I think the two most reasonable options are either: (a) **always** issue a **warning**, or (b) allow user to set config flag to select between nothing, warning, or error.

thoughts @jameshegarty @leonardt ?

magma/config.py
```diff
@@ -22,3 +22,18 @@ def set_debug_mode(value=True):
 
 def get_debug_mode():
     return __DEBUG_MODE
+
+
+class _MagmaDebugSection:
```
rsetaluri: Perhaps this utility can go in a different file

magma/port.py
```diff
@@ -208,7 +208,7 @@ def trace(self):
             if len(self.wires.inputs) < 1:
                 # print('Warning:', str(self), 'is not connected to an input')
                 return None
-            assert len(self.wires.inputs) == 1
+            #assert len(self.wires.inputs) == 1
```
leonardt: Is this necessary for this change set? This logic is a bit weird, and will be cleared up in the `hwtypes` branch.  I guess technically this shouldn't break any existing code (since it would've triggered an assertion failure otherwise), but there may be other internal code that depends on this assumption (this is just my assumption though, I'm not sure if that's true)

magma/config.py
leonardt: Nice, maybe in `testing`? One thing that I never got to was that this code should run even if the test fails (otherwise if the test fails, it leaves the debug mode on for other tests since it never hits the exit condition.

magma/port.py
rsetaluri: That's right, uncommented it.

(This might be a spurious assertion though, calling `wired()` on an input that drives > 1 wire doesn't work due to this assertion...can take a look later.)

rsetaluri:
Makes sense, switched from warning to error.

I'll investigate how to track those errors and using the atexit handler.

magma/config.py
rsetaluri: I can move to `testing`.

I am seeing that `__exit__` runs even if the test fails (assuming it fails in a normal way). See:

```python
class _WithBlock:
    def __enter__(self):
        print ("Enter")

    def __exit__(self, typ, value, traceback):
        print ("Exit")

def test():
    with _WithBlock():
        assert False, "My assertion"

if __name__ == "__main__":
    test()
```

PR #482, opened 2019-11-01
allow passing in of enviroment into comb and seq

rdaly525:
Also requires a PR for ast_tools

### DIFF
--- a/magma/syntax/combinational.py
+++ b/magma/syntax/combinational.py
@@ -15,7 +15,7 @@ from magma.ssa import convert_tree_to_ssa
 from magma.config import get_debug_mode
 import itertools
 import typing
-from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT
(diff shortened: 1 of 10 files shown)


rdaly525:
https://github.com/leonardt/ast_tools/pull/22

magma/syntax/combinational.py
```diff
@@ -15,7 +15,7 @@
 from magma.config import get_debug_mode
 import itertools
 import typing
-from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT
+from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT, SymbolTable
```
rsetaluri: nit: if something is exported from a module (i.e. being imported by someone else) then it shouldn't have a leading `_` :)

rdaly525: I am indifferent, but this is not my code.

rdaly525:
@leonardt, could you republish ast_tools? That is needed to get this passing

PR #490, opened 2019-11-08
Fix bug in tuple iswhole logic for bulk connections

leonardt:
This fixes the case where the compiler improperly tries to do a bulk
connection.

The desired behavior is to check:
is the value connected to the current tuple "whole", which means
that a tuple of the same type is connected to the current tuple,
in this case we can  connect the two tuples together, rather than
their components.

However, this check was improperly indicating a bulk connection in the
case when the driving value satisfied all the properties for a whole
tuple, except that it doesn't contain all the keys of the parent tuple.

This change strengthens the check to ensure that the driver is only
considered whole when the values refer to all the values corresponding
to the parent tuple.

### DIFF
--- a/magma/tuple.py
+++ b/magma/tuple.py
@@ -131,6 +131,10 @@ class TupleType(Type):
             if ts[i].name.index != self.Ks[i]:
                 return False
 
+        if len(ts) != len(ts[0].name.tuple):
+            # elements should refer to a whole tuple
+            return False
+
         return True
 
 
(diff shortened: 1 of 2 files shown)


magma/tuple.py
```diff
@@ -131,6 +131,10 @@ def iswhole(self, ts):
             if ts[i].name.index != self.Ks[i]:
                 return False
 
+        if len(ts) != len(ts[0].name.tuple):
```
rsetaluri: It seems like this "exclusion" method is not robust (e.g. we found another case that indicated non-whole-ness)...probably would be better to write down the exact condition we want and check for it.

But if this fixes the immediate issue, we can fix ^ separately.

leonardt: Agreed

Issue #492, opened 2019-11-13
Can't cast to async reset


jameshegarty:
Casting operator to async reset looks broken?

>class AsyncResetTest(m.Circuit):
>  IO = ['I', m.BitIn, 'O', m.AsyncResetOut ]
>  @classmethod
>   def definition(io):
>   io.O <= m.asyncreset(io.I)

produces:

>ERROR: AsyncResetTest: Cannot wire together
> self.I : Bit
> self.O : coreir.arstIn
>I AM DYING!
>python3: ../src/ir/context.cpp:112: void CoreIR::Context::die(): Assertion `0' failed.
>Aborted (core dumped)`

leonardt:
That looks like a bug, will investigate

PR #493, opened 2019-11-14
Fix handling of asyncreset cast in coreir backend

leonardt:
Fixes https://github.com/phanrahan/magma/issues/492

Coreir uses a special named type to distinguish async reset, the only way to wire up a bit to this type is to insert a `wrap` module that converts from the bit to the named type (effectively an explicit cast using an instance of a special primitive).  Magma, on the other hand, handles the cast by just wiring up the underlying values (so there's no explicit instance between them).

These changes introduce a pass that is run before compilation that insert the coreir wrap primitive instances to handle the conversion to the named type.  For now it only handles AsyncRest and AsyncResetN, but we could generalize this later to handle arbitrary named types that need this logic.

In order to do this pass, I needed to include logic to allow unwiring a port (so I could unwire a bit connected to an asyncreset, then rewire them to the intermediate wrap instance).

TODO: Generalize support for nested types (right now the pass only works on ports that are asyncreset(n), but not if they're nested inside a tuple or array)

Depends on https://github.com/rdaly525/coreir/pull/816

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -182,6 +182,10 @@ class Array(Type, metaclass=ArrayKind):
         for k in range(len(i)):
             i[k].wire(o[k], debug_info)
 
+    def unwire(i, o):
+        for k in range(len(i)):
+            i[k].unwire(o[k])
+
     def driven(self):
         for t in self.ts:
             if not t.driven():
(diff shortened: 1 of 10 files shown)


leonardt:
Also TODO: add logic for instances

leonardt:
Instances and nested types now tested

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/818 and https://github.com/rdaly525/coreir/pull/817 too

PR #496, opened 2019-11-15
Enforce Unique Names

jameshegarty:
I suggest we change magma to enforce that module names are unique at definition time. Currently, if the user forgets to memoize a generator, their code will create thousands of duplicate module definitions, which the compiler will silently accept. This is causing us to have hour long compile times, and it's not possible to debug this if we don't raise an error.

I added a quick fix to enforce this behavior, but (if we agree this is the behavior we want), I will need help squashing all the duplicate name bugs. It looks like a large number of generators in the repo that are not being memoized.

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -596,8 +596,12 @@ def DefineCircuit(name, *decl, **args):
     currentDefinition = DefineCircuitKind( name, (Circuit,), dct)
     return currentDefinition
 
+__seenCircuitNames = {}
 def EndDefine():
     if currentDefinition:
+        if currentDefinition.name in __seenCircuitNames:
+            raise Exception(f"The circuit name '{currentDefinition.name}' was already used. Module names must be unique!")
+        __seenCircuitNames[currentDefinition.name] = 1
         debug_info = get_callee_frame_info()
         currentDefinition.end_circuit_filename = debug_info[0]
         currentDefinition.end_circuit_lineno   = debug_info[1]
--- a/magma/syntax/combinational.py
+++ b/magma/syntax/combinational.py
@@ -275,7 +275,6 @@ def combinational(
     else:
         wrapped_combinational = ast_utils.inspect_enclosing_env(
                 _combinational,
-                decorators=[combinational],
-                st=env
+                decorators=[combinational]
         )
         return wrapped_combinational(fn)
--- a/tests/test_circuit/test_define.py
+++ b/tests/test_circuit/test_define.py
@@ -54,7 +54,7 @@ def test_for_loop_def(target, suffix):
     And2 = m.DeclareCircuit('And2', "I0", m.In(m.Bit), "I1", m.In(m.Bit),
                             "O", m.Out(m.Bit))
 
-    main = m.DefineCircuit("main", "I", m.In(m.Bits[2]), "O", m.Out(m.Bit))
+    main = m.DefineCircuit("main2", "I", m.In(m.Bits[2]), "O", m.Out(m.Bit))
 
     and2_prev = None
     for i in range(0, 4):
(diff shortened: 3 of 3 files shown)


jameshegarty:
Also @leonardt there was a breaking change to the combinational file on my machine which I fixed (?) here.

magma/syntax/combinational.py
```diff
@@ -275,7 +275,6 @@ def wrapped(fn):
     else:
         wrapped_combinational = ast_utils.inspect_enclosing_env(
                 _combinational,
-                decorators=[combinational],
-                st=env
+                decorators=[combinational]
```
leonardt: Can you try updating to the latest version of `ast_tools` and reverting this change? (`pip install --upgrade ast_tools`), the `st` argument was introduced in the latest release.

magma/circuit.py
```diff
@@ -596,8 +596,12 @@ def DefineCircuit(name, *decl, **args):
     currentDefinition = DefineCircuitKind( name, (Circuit,), dct)
     return currentDefinition
 
+__seenCircuitNames = {}
 def EndDefine():
     if currentDefinition:
+        if currentDefinition.name in __seenCircuitNames:
+            raise Exception(f"The circuit name '{currentDefinition.name}' was already used. Module names must be unique!")
+        __seenCircuitNames[currentDefinition.name] = 1
```
leonardt: For now, could we add this as a configurable option? That way we can merge it into the mainline release ASAP but then figure out how exactly we want to handle it and what the default behavior should be (I expect there to be some discussion).

We could add a simple API like `set`/`get_debug_mode`  https://github.com/phanrahan/magma/blob/master/magma/config.py#L14-L24 to enable this?

magma/syntax/combinational.py
jameshegarty: Ok I guess I didn't realize I had to run pip upgrade, but that makes sense.

magma/circuit.py
jameshegarty: IMO I would rather have the discussion and decide to fix the issue or not. We don't need to have this merged in immediately. What do people think? Are there any arguments on the other side?

rsetaluri:
This issue has ping-ponged a bit in magma previously. I think this condition used to be enforced, but we changed it because not enforcing it was more general and robust. In particular, if you have a top level design that's using components from disparate places, they may not have unique names. And in fact they may not be modifiable (imagine you're importing some third party magma code or something).

Plus, the issue of not caching generators is separate and should be solved "directly". I think one reasonable thing would be to make auto-uniquification *off* by default in the compiler; i.e. this error is raised at *compile* time rather than define time. This mode already exists, it just has to be enabled.

tests/test_circuit/test_define.py
```diff
@@ -86,7 +86,7 @@ def test_interleaved_instance_wiring(target, suffix):
     And2 = m.DeclareCircuit('And2', "I0", m.In(m.Bit), "I1", m.In(m.Bit),
                             "O", m.Out(m.Bit))
 
-    main = m.DefineCircuit("main", "I", m.In(m.Bits[2]), "O", m.Out(m.Bit))
+    main = m.DefineCircuit("main3", "I", m.In(m.Bits[2]), "O", m.Out(m.Bit))
```
rsetaluri:
```suggestion
    main = m.DefineCircuit("main3", "I", m.In(m.Bits[2]), "O", m.Out(m.Bit))
```

leonardt:
@rsetaluri I think that would help with performance (avoiding the repr uniquification time), another incremental option might be to allow selective whitelist of names to uniquify (so if there's a commonly used name such as `Util` or something, uniquification can be enabled for those circuits, while off by default.

leonardt:
Another improvement would be to add automatic name generation in the `Generator` class, right now it caches so it avoids regenerating multiple instances, but I think if the use doesn't provide a name to the generated Circuit, it will still trigger uniquification downstream for each instance of the generator

PR #502, opened 2019-12-03
Make pycodestyle run by defualt for pytest

rsetaluri:
Through setup.cfg, by default (i.e. unless over-ridden) all calls to
pytest will invoke pycodestyle. Note that currently all files are
blacklisted such that no errors are raised. All new files should be
linted before being checked in.


leonardt:
perhaps we can submit a style PR with an automated tool like YAPF to get us started with existing files?

PR #504, opened 2019-12-04
Loop unroll

leonardt:


docs/circuit_definitions.md
```diff
@@ -157,6 +157,42 @@ class Foo(m.Circuit):
...
+## Statically Elaborated For Loops
+Statically elaborated for loops are supported using the [ast_tools loop
+unrolling macro](https://github.com/leonardt/ast_tools#loop-unrolling).
+Here's an example:
+```python
+from ast_tools.passes import begin_rewrite, loop_unroll, end_rewrite
+
+n = 4
+@m.circuit.combinational
+@end_rewrite()
```
rsetaluri: Just so I understand, looks like 4 things are needed here: `begin_rewrite`, `loop_unroll`, `end_rewrite` decorators, and wrapping the iterator in `ast_tools.macros.unroll()`. Just wondering why all of these are necessary and what the alternative syntaxes could be? (Figured you've done due diligence, just unclear to me. The biggest concern is communicating to the user if they forget to do one of these things, or what even happens in that case.)

cdonovick: `begin_rewrite` collects the ast and environment and returns them.  `loop_unroll` unrolls iterators passed to `ast_tools.macros.unroll()`, and returns the modified ast /env.  `end_rewrite` strips decorators between the innermost `begin_rewrite` and `end_rewrite` (inclusive) from the ast, then returns results of execing the ast in the env.

cdonovick: The begin / end rewrite syntax is becoming more and more established so if you don't like I would get over to ast_tools and file an issue.

PR #505, opened 2019-12-06
Refactor backend

rsetaluri:


rsetaluri:
@leonardt I understand this may have some conflict with `hwtypes` PR (though I think its mostly simple, e.g. `isinput` -> `is_input` and `issubclass(ArrayType)` -> `isinstance(Array)`). I can insure you by promising to own those merge conflicts when you rebase :)

magma/backend/coreir_.py
```diff
@@ -81,445 +32,55 @@ def get_instance(self):
...
+
+
 CoreIRContextSingleton()
 
+_context_to_modules = {}
+
+
 class CoreIRBackend:
-    context_to_modules_map = {}
-    def __init__(self, context=None, check_context_is_default=True):
```
leonardt: IIRC removing `check_context_is_default=True` will break aetherling code since there's some code that creates these singleton contexts and uses this flag to override the warning (disabling it).  Perhaps we can add this back just for backwards compatibility?

magma/backend/coreir_.py
```diff
@@ -81,445 +32,55 @@ def get_instance(self):
...
-                for (k, t) in zip(port.Ks, port.Ts):
-                    check_type(t, errorMessage.format("Tuple({}:{})".format(k, "{}")))
-            elif isinstance(port, (BitKind, ClockKind, EnableKind, ResetKind, AsyncResetKind, ResetNKind, AsyncResetNKind)):
-                return
-            else:
-                raise CoreIRBackendError(errorMessage.format(str(port)))
-        for name, port in definition.interface.ports.items():
-            check_type(type(port), 'Error: Argument {} must be comprised only of Bit, Array, or Tuple')
-
-    def get_type(self, port):
```
leonardt: I also recall @David-Durst 's code using this method (and possible others) from the corer backend directly, so either we should try to preserve the old interface so we don't break his code, or work with him on a plan to update it.

leonardt: Actually, reading through these changes, perhaps the usage I remember was restricted to the coreir frontend? (Which has been updated, so should be okay) @David-Durst can you comment on whether these changes to the coreir backend will affect your code?

magma/backend/coreir_.py
David-Durst: My old tests used to create coreir contexts. My new tests don't they just need access to `CoreIRContextSingleton().reset_instance()`. See https://github.com/David-Durst/aetherling/blob/master/conftest.py

magma/backend/coreir_.py
David-Durst: I need to port renaming preserved (converting between CoreIR's .I ports and magma's .in names) Otherwise, I think the I'm mainly using https://github.com/phanrahan/magma/blob/master/magma/frontend/coreir_.py directly. Unfortunately, CI won't be useful in debugging these issues since I'm still on a CoreIR branch.

magma/backend/coreir_.py
rsetaluri: Got it. In either case it was just a warning (so unless you were checking the output for a warning, it shouldn't affect the result of tests). @David-Durst confirming this change is ok?

magma/backend/coreir_.py
rsetaluri: @David-Durst no functionality has changed. Most likely the same functionality is now available to in `frontend/coreir_.py`. Otherwise, it may have moved to `backend/coreir_utils.py`, in which case I can expose it for you and you'll just have to modify the call-sites marginally.

David-Durst: Ok, then this should work for me. I'll make changes to my code if it breaks. lgtm.

rsetaluri:
Checked with @David-Durst offline, looks like there *may* be some changes required to Aetherling as a result of this change, but should be simple/superficial. Plan is to deal with those when they come up when Aetherling points to master.

PR #509, opened 2019-12-10
Refactor and clean-up passes

rsetaluri:
-- Clean up passes/{clock,passes}.py:
   Style and refactoring changes only.
-- Remove unused file passes/ir.py
-- Clean up passes/debug_name.py
-- Clean up passes/tsort.py:
   Also moved tsort test to tests/test_tsort.py.
-- Un-blacklist magma/__init__.py


magma/passes/ir.py
```diff
@@ -1,12 +0,0 @@
...
-
-class IRPass(DefinitionPass):
-    def __init__(self, main):
-        super(IRPass, self).__init__(main)
-        self.code = ''
-
-        WireClockPass(main).run()
-
-    def __call__(self, definition):
-        self.code += repr(definition) + '\n\n'
```
leonardt: Is there a reason this was removed?  I think it could be useful to print out the entire representation of a circuit (including dependencies)

phanrahan:
I use that IRPass in some of my dev directories.

rsetaluri:
@leonardt @phanrahan, was just confused by the name and didn't seem to be used anywhere inside magma. Seems more like a user-specific pass then a general one that should be in magma. Can put it back for now.

PR #515, opened 2019-12-12
Hwtypes2

leonardt:
Mantle branch: https://github.com/phanrahan/mantle/pull/165
Fault branch: https://github.com/leonardt/fault/pull/187

Migration guide is in `docs/magma_2_migration_guide.md` (we'll probably need to add to this)


splhack:
https://github.com/phanrahan/magma/blob/e6a341601fa91f12ef7a60f588d66a8af4aeb1d8/magma/array.py#L218

looks like this `array` was forgotten to replace? causing undefined error.

> NameError: name 'array' is not defined

probably supposed to be like this?

    return type(self)[len(res_bits), self.T](res_bits)

magma/digital.py
```diff
@@ -0,0 +1,198 @@
...
+    def const(self):
+        return self is VCC or self is GND
+
+    def unwire(i, o):
+        i.port.unwire(o.port)
+
+    @classmethod
+    def flat_length(cls):
+        return 1
+
```
splhack: probably loam was supposed to be updated to sync with magma hwtypes2, but in the meantime, these two methods were required to build with loam. 

    def getinst(self):
        t = self.trace()
        return t.name.inst if t is not None else None

    def getgpio(self):
        return self.getinst()

- https://github.com/phanrahan/loam/search?q=getinst&unscoped_q=getinst
- https://github.com/phanrahan/loam/search?q=getgpio&unscoped_q=getgpio

setup.py
```diff
@@ -6,7 +6,7 @@
 
 setup(
```
splhack: It would be nice to have `dataclasses` in `install_requires` for `pip install -e .`

    @@ -26,6 +26,7 @@ setup(
         install_requires=[
             "colorlog",
             "astor",
    +        "dataclasses",
             "six",
             "mako",
             "pyverilog",

phanrahan:
Reviewing the code.

One question: Why is Flip a direction?

It would also be nice if test_types included tests similar to hwtypes. In particular, for the new issubclass and isinstance aspects of types (I didn't find those).

setup.py
leonardt: Done

magma/digital.py
leonardt: Thanks, I've added these in for compatibility, although we should revisit this interface as it seems pretty loam specific so perhaps they should not be part of the magma core.

leonardt:
Flip is a direction since it made it simpler to implement flip using qualify, although it would be trivial to change it so `flip` logic internally (rather than relying on qualify).

splhack:
I noticed some regression(?)

master

    sint_value = m.sint(1, 1)
    print(isinstance(sint_value, m.Bits)) -> True
    print(isinstance(sint_value, m.SInt)) -> True
    print(isinstance(sint_value, m.UInt)) -> False

hwtypes2

    sint_value = m.sint(1, 1)
    print(isinstance(sint_value, m.Bits)) -> True
    print(isinstance(sint_value, m.SInt)) -> True
    print(isinstance(sint_value, m.UInt)) -> True  <- ??

phanrahan:
We should make sure we have tests for all the rules. And the implementation should pass the tests.

phanrahan:
@leonardt still don't understand the implementation of flip. In the old implementation, the flip changed the qualification. That meant it was immediately applied. Is that how it still is intended to work? In that case, flip is really not a valid direction.

splhack:
Added tests https://github.com/phanrahan/magma/pull/525

leonardt:
Question on the relationship between directed/undirected types.

Right now, a directed type is not related to an undirected type, e.g.
```python
>>> isinstance(m.In(m.Bit)(), m.Bit)
False
>>> isinstance(m.In(m.Bit), m.BitKind)
True
```

This works because we use instances of the meta class to check the type of the type, so the type of directed types are the same (they are instances of BitKind)

However, now we are planning to use `issubclass(T0, T1)` to check type relation.

For this to work, we need `m.In(m.Bit)` to be a subclass of `m.Bit`, so we can check
```
>>> issubclass(m.In(m.Bit), m.Bit)
True
```

But that also means that a directed instance is an instance of an undirected instance
```
>>> isinstance(m.In(m.Bit)(), m.Bit)
True
```

Which is a departure from the old behavior.

I don't think this should affect existing code since usually in this case we were using the meta class relationship.

The question is: is having an instance of an directed type also an instance of an undirected type a reasonable design? Otherwise we may have to revert to the `isinstance` of meta class relationship we have been using before.

phanrahan:
I think the old way was broken, and the new way is better. That is, a directed type should be a subclass of an undirected type.

leonardt:
Good to hear :), I discussed this offline with James H. and Raj and they both agree.

leonardt:
I've added tests for the various typing relations, I'll merge this in tomorrow (alongside mantle/fault) unless there are any more objections.  Peak/lassen/garnet will need to be updated

phanrahan:
Looks great. Thanks for adding all these tests.

splhack:
This code works with master

    from magma import *
    from mantle import *

    def test_namedtuple():
        class A(Product):
            a0 = Bit
            a1 = SInt[8]

        @circuit.sequential(async_reset=False)
        class TestNamedTuple:
            def __init__(self):
                self.a0: Bit = bit(0)
                self.a1: SInt[8] = 0

            def __call__(self, a: A, b: Bit) -> A:
                if b:
                    new_a = a
                else:
                    new_a = namedtuple(a0=self.a0, a1=self.a1)

                self.a0 = new_a.a0
                self.a1 = new_a.a1

                return new_a

    test_namedtuple()

but got error with hwtypes2

      File "magma/magma/ast_utils.py", line 54, in compile_function_to_file
        exec(code, defn_env)
      File ".magma/TestNamedTuple_comb.py", line 5, in <module>
        class TestNamedTuple_comb(m.Circuit):
      File "magma/magma/circuit.py", line 441, in __new__
        self.definition()
      File ".magma/TestNamedTuple_comb.py", line 13, in definition
        new_a_2 = phi([new_a_1, new_a_0], io.b)
      File "mantle/mantle/common/operator.py", line 74, in wrapper
        retval = fn(*args, **kwargs)
      File "mantle/mantle/common/operator.py", line 296, in mux
        return Mux(len(I), T=T, **kwargs)(*I, S)
      File "mantle/mantle/coreir/MUX.py", line 126, in Mux
        return DefineMux(height, width, T=T)(**kwargs)
      File "mantle/mantle/coreir/MUX.py", line 89, in DefineMux
        class _Mux(Circuit):
      File "magma/magma/circuit.py", line 417, in __new__
        self = CircuitKind.__new__(metacls, name, bases, dct)
      File "magma/magma/circuit.py", line 101, in __new__
        renamed_ports=dct["renamed_ports"])
      File "magma/magma/interface.py", line 185, in __init__
        port = port.flip()
      File "magma/magma/tuple.py", line 374, in flip
        cls.__name__, (base, ), {})
      File "hwtypes/hwtypes/adt_meta.py", line 128, in _cache_handler
        t = mcs._from_fields(fields, name, bases, ns, **kwargs)
      File "magma/magma/tuple.py", line 345, in _from_fields
        t = TupleKind.__new__(mcs, name, bases, ns, **kwargs)
      File "magma/magma/tuple.py", line 41, in __new__
        raise TypeError("Can't inherit from multiple different bound_types")
    TypeError: Can't inherit from multiple different bound_types

looks like due to `In` vs `Out`?

    bound_types=(Out(Bit), Out(SInt[8]))
    base.fields=(In(Bit), In(SInt[8]))


--
By the way, is there a way to create instance from `class A(Product)` aside from namedtuple?
something like

    a = A(a0=self.a0, a1=self.a1)

leonardt:
@splhack thanks, fixed by https://github.com/phanrahan/magma/pull/515/commits/c3fa956f82ff6251f15e3bcbf960d7293b56e028

PR #516, opened 2019-12-12
Remove unnec. rename in uniq. pass

rsetaluri:
Previously we were calling rename with the same name (i.e. an
unnecessary rename). For normal magma circuits this wasn't an issue,
since the rename was a no-op, but it was surfacing when doing so for
verilog-wrapped circuits, since they can't be renamed.

Added a test for this issue.

### DIFF
--- a/magma/uniquification.py
+++ b/magma/uniquification.py
@@ -36,8 +36,7 @@ def _hash(definition):
 
 class UniquificationPass(DefinitionPass):
     def __init__(self, main, mode):
-        super(UniquificationPass, self).__init__(main)
-        self.definitions = {}
+        super().__init__(main)
         self.mode = mode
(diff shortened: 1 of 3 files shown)


magma/uniquification.py
```diff
@@ -56,27 +56,10 @@ def __call__(self, definition):
         else:
             if self.mode is not UniquificationMode.UNIQUIFY:
                 assert self.seen[name][key][0].name == definition.name
-            else:
-                type(definition).rename(definition, self.seen[name][key][0].name)
             self.seen[name][key].append(definition)
 
-    def _run(self, definition):
```
rsetaluri: This was also unnec. code, so removed. Not really related to the mentioned issue.

PR #517, opened 2019-12-12
Add separate debug option in *_to_verilog functions

Kuree:
This additional parameter `debug` allows users to only debug on the circuit they're interested in, instead of the entire design. It will also speed up the debugging performance.

Also refactor the debug information extraction code to the magma package so that fault can use it to generate correct debugging logic.

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -23,7 +23,8 @@ from magma.syntax.combinational import combinational
 from magma.syntax.sequential import sequential
 from magma.syntax.combinational import combinational
 from magma.syntax.sequential import sequential
-from magma.syntax.verilog import combinational_to_verilog, sequential_to_verilog
+from magma.syntax.verilog import combinational_to_verilog, \
+    sequential_to_verilog, build_kratos_debug_info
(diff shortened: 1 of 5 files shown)


magma/circuit.py
```diff
@@ -23,7 +23,8 @@
 from magma.syntax.sequential import sequential
 from magma.syntax.combinational import combinational
 from magma.syntax.sequential import sequential
-from magma.syntax.verilog import combinational_to_verilog, sequential_to_verilog
+from magma.syntax.verilog import combinational_to_verilog, \
+    sequential_to_verilog, build_kratos_debug_info
```
rsetaluri: why is `build_kratos_debug_info` exposed in `magma.circuit`?

Kuree: I have no idea where to expose.

rsetaluri: Probably leaving it in `magma.syntax` or `magma.syntax.verilog`

Kuree: I think `magma` doesn't have `syntax` imported and that's why I put in a different place. I will add `syntax` to the top level.

PR #518, opened 2019-12-20
Inline Verilog (assertions)

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/823

This prototypes a feature to allow the use of inline verilog for writing assertion properties.  The main issue is interpolating the verilog name for port references (can be hierarchical sub instances) into the assertion string.

Eventually we'd like to do this in a more principled fashion with a Python syntax and a coreir representation, but this is an initial step to enable the definition of inline properties.  The view logic is a refactor of the `fault` logic for handling internal signals.  We should migrate the fault code to use this view logic so the code is shared, and eventually this logic should be abstracted and moved into coreir (since that's where the management of conversion from names to verilog happens).


tests/test_verilog/test_inline.py
```diff
@@ -0,0 +1,97 @@
+import magma as m
```
rsetaluri: Think these tests should just check json instead of verilog

leonardt: But checking the verilog verifies the translation through coreir (namely, the coreir backend reading the metadata correctly)

leonardt: I guess we could assume that the coreir test suite covers this

PR #520, opened 2019-12-20
Add refactoring tool for new wire syntax

cdonovick:
usage: `./tools/wire_refactor.py <src_file> <dst_file>`

Replaces old wire statements `x <= y` with `x @= y`

old wire statements are identified as ast nodes of the form `Expr(Compare(..., [LtE(), ...], ...))`

A bit hacky because I can't use an `ast.NodeTransformer` because I want to maintain whitespace / comments etc...

Tested on `examples/fifo.py` and a hand crafted set of examples.  I don't know how we want to handle testing of this.

Should handle `;` delimited statements:
`x <= y; z <= x` becomes `x @= y; z @= x`

Can handle 'tricky' lines:
`bv[c <= d] <= x` becomes `bv[c <= d] @= x`

Will fix buggy code such as:
`x <= y < z` to  be `x @= y < z`
This technically changes the semantics of the code, the correct translation would be something like `(x @= y) and (y < z)` but this is a syntax error.  Also, I understand this a source of complaints so I figured better to try fix the code.

This will break code if for some reason there are random `Expr` that look like wire statements.  For example `1 <= 2` (a perfectly valid line of python, which is effectively a NOP) will become `1 @= 2` which will error.  Or worse the meaning of a line will change if the `__imatmul__` or `__matmul__` is defined on operands of a `LtE`.   So definitely check the diff.

### DIFF
--- a/tools/wire_refactor.py
+++ b/tools/wire_refactor.py
@@ -139,8 +139,8 @@ if __name__ == '__main__':
     diff = difflib.unified_diff(
         src_txt.splitlines(),
         new_src_txt.splitlines(),
-        fromfile=sys.argv[0],
-        tofile=sys.argv[1] + ' (pending)',
+        fromfile=sys.argv[1],
+        tofile=sys.argv[2] + ' (pending)',
         lineterm=''
     )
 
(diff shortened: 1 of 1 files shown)


leonardt:
LGTM, can we add a simple test/documentation showing how to use this on a simple file?

cdonovick:
I will add documentation and an example file

cdonovick:
@leonardt  `examples/fifo.py` works well as an example.  Do you want a more minimal example?

PR #521, opened 2020-01-01
Replace non-ascii in setup.cfg

splhack:
(probably better to update Python itself but) To fix pip install with Python 3.6

    $ pip3 install -e .
    Obtaining file:///...
        Complete output from command python setup.py egg_info:
    ...
          File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
            return codecs.ascii_decode(input, self.errors)[0]
        UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 71: ordinal not in range(128)

        ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in .../magma/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [pycodestyle]
 max-line-length = 80
 
-# E741: do not use variables named ‘l’, ‘O’, or ‘I’
+# E741: do not use variables named 'l', 'O', or 'I'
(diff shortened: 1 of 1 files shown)


splhack:
same here https://github.com/leonardt/fault/blob/master/setup.cfg#L4

leonardt:
Thanks!

PR #524, opened 2020-01-07
Allow `CLK` argument in sequential __call__

splhack:
I really like combinational and sequential, it can reduce a lot of boilerplate code but a current blocker is clock interface, if I understand correctly. A good example would be this FIFO. This could be turned into sequential if the generated combinational function had CLK for the memory.
https://github.com/phanrahan/mantle/blob/master/mantle/experimental/fifo.py

I can put CLK to input if it was combinational function.

    @m.circuit.combinational
    def FIFO(enq_val: m.Bit, ... CLK: m.Clock) -> OutType:

But I can't find a way to put CLK to the generated combination function by sequential. So, first, adding arguments to combinational decorator, similar with the current async_reset argument of sequential decorator. The arguments (async_reset, async_resetn, ce, clock, reset) are almost same as ClockInterface. The benefit is, no need to write CLK in the input with `clock=True`.

    @m.circuit.combinational(clock=True)
    def FIFO(enq_val: m.Bit, ...) -> OutType:

Second, sequential decorator arguments. Just putting the same arguments to sequential makes conflict with the current async_reset. sequential __init__ requires async_reset for Registers, but combinational may not use async_reset. Thus, adding `comb_` prefix to the arguments.

    @m.circuit.sequential(comb_clock=True)
    class FIFO:
        def __init__(self):
            ...
        def __call__(self, enq_val: m.Bit, ...) -> OutType:

and generated combinational by sequential code looks like

    class FIFO_comb(m.Circuit):
        IO = [..., 'CLK', m.In(m.Clock)]

Any thoughts? I can add unit tests if it goes well.


phanrahan:
Thanks for the comment. We have had several other requests for enhancements to the clock interface for sequential. For example, a FIFO could need two clocks, one for the read point and one for the write port.

Let's use this thread to discuss different designs. Adding @leonardt and @rdaly525.

splhack:
A short term solution would be allowing to put CLK to '__call__', modify sequential code to skip add CLK if input has CLK :)

    def __call__(self, enq_val: m.Bit, ..., CLK: m.Clock) -> OutType:

phanrahan:
I agree with that.

We should make clocks behave like magma circuits defined not using sequential. You have a good default behavior, but total control over how the clocks are declared and wired up.

rdaly525:
> A short term solution would be allowing to put CLK to '**call**', modify sequential code to skip add CLK if input has CLK :)
> 
> ```
> def __call__(self, enq_val: m.Bit, ..., CLK: m.Clock) -> OutType:
> ```

I think this solution gives the most flexibility, clearly shows user intent, and is likely the easiest to implement. This could handle custom clock gating, and allows naming of the clock port(s). It does somewhat clash with the use of decorators to specify async_reset. If we go this route, we should also allow the same behavior with async_reset signals (They have their own type right?). 

It also does not handle multiple clocks (which is what I initially thought this issue was about regarding the fifo), and it might be useful to think of a solution for this that is also general for multiple clocks. 

The issue of multiple clocks can be shown in the following example

```
class A:
    def __initi__(self):
        self.reg1 : m.Bit = m.Bit(1)
        self.mem : DualPorted2ClkRam = DualPorted2ClkRam(...)

   def __call__(self, ..., read_clk : m.Clock, write_clk, m.Clock) -> ...
       #Passing in multiple clocks works great
       self.mem(read_clk, write_clk, ...)

       #Which clock is used here?
       if ...:
           self.reg1 = m.Bit(0)
       else:
           self.reg1 = self.reg1

```

cdonovick:
I don't know enough about the nuances of clock crossing to give a really educated thought but maybe we could declare asynchronous state separately and insert synchronization logic automatically.

```Python
@sequential(synchronizer=DoubleFlop)
class A:
    def __init__(self):
        self.reg1 : m.Bit = m.Bit(1)

    def asynch_state(self):
        self.mem : DualPorted2ClkRam = DualPorted2ClkRam(...)

    def __call__(self, ..., read_clk : m.Clock, write_clk: m.Clock) -> ...
        self.mem(read_clk, write_clk, ...) 
        # Wire directly to mem, assumes anything passed to mem
        # is synchronized or that mem will have its own synchronizer

        if self.mem.rdy: # Gaurd mem.rdy with a double flop synchronizer. 
            self.reg1 = m.Bit(0) #writes on implicit clock
        else:
            self.reg1 = self.reg1
```

splhack:
`__call__` now allows `CLK` arg and other `m.Clock` args safely by this PR.

And, for the `__init__`  Registers clock, technically whatever clock can be connected to the default `CLK` in this way, isn't it?

    @sequential(async_reset=True)
    class A:
        def __init__(self)
            self.reg1 : m.Bit = m.Bit(1)  # will be connected with the_third_clk

       def __call__(self, ..., read_clk : m.Clock, write_clk, m.Clock) -> ...
           mem = DualPorted2ClkRam(..., read_clk, write_clk)
           if ...:
               self.reg1 = m.Bit(0)
           else:
               self.reg1 = self.reg1

    a = A(..., read_clk=read_clk, write_clk=write_clk)
    a.CLK <= the_third_clk

rdaly525:
> And, for the `__init__` Registers clock, technically whatever clock can be connected to the default `CLK` in this way, isn't it?
> 

That is a possible solution, but it would prevent the ability to specify that a passed in clock gets hooked up to 'reg1' and I would argue would not be obvious to users. I guess one could always wrap 'reg1' in a different class that has a single clock and pass in the appropriate clock, but this seems too restrictive. 

Perhaps the ability to specify which clock will be the 'default' clock in a decorator?

leonardt:
Clarification question:
Is the issue here that'd you like `CLK` to be available inside the __call__ and @combinational methods so that you can use it to wire up to an instance that's created inside the body of the combinational function? (e.g. a memory)

Another option that we've been considering is to allow arbitrary magma code to be used inside the `__init__` method of the sequential class.

So, if you wanted to define a fifo with an internal memory instance, you could declare in the __init__

```
@sequential
class FIFO:
    def __init__(self):
         self.memory = DefineMemory(width, height)
```

This still leaves the issue of how to declare and wire up the clock.  I like the idea of allowing the user to explicitly declare it, perhaps we could use some syntax similar to how `m.Circuit` works, e.g.

```
@sequential
class FIFO:
    ClockInterface = {
        "READ_CLK": m.In(m.Clock),
        "WRITE_CLK": m.In(m.Clock)
    }

    def __init__(self):
         self.memory = DefineReadWriteMemory(width, height)
         # wire clocks up
         self.memory.READ_CLK @= self.READ_CLK
         self.memory.WRITE_CLK @= self.WRITE_CLK
```

then, inside the `__call__` method, you could refer to ports (read outputs, write to inputs).

However, it seems in this case, you'd need to have two methods, one for each read/write clk (similar to two posedge CLK always blocks in verilog).

So, I think there's two issues her:
1) How do we declare clocks explicitly? And how do we wire them up? I'm in favor of doing them as structural magma rather than using combinational since we typically aren't doing logic on the clocks (but I think it'd be okay to allow referring to the clock value so you can do things like clock gating).  I think we could add an interface for declaring the clock signals and allow the user to refer to them in` __init__`/`__call__`. We can also add support for wiring inside `__init__`.
2) How do we support multiple clocks? I think we'll need some notion of separate functions for each clock event, then we'll need some support for handling clock crossing.

leonardt:
For implicit registers, we could add support for optionally wiring up a clock, e.g.

```
def __init__(self):
    self. x : m.Bits[2] = ...
    self.x @= self.CLK3
```

splhack:
> Clarification question:
> Is the issue here that'd you like CLK to be available inside the call and @combinational methods so that you can use it to wire up to an instance that's created inside the body of the combinational function? (e.g. a memory)

Yes. More specifically, sequential `__call__` method. (combinational itself doesn't have these issues. only on sequential `__call__`.)

1. We cant have `CLK` argument due to conflict with `__init__`'s `CLK`. It is not a design issue, just an implementation issue.

2. any `m.Clock` argument weren't connected. This is also implementation issue. m.Clock were ignored, and even passing m.Clock, still there were only implicit clock connection. so, using keyword argument to connect clock explicitly from outside circuit.

rdaly525:
@cdonovick There is nothing really magical about clock crossings. All that is necessary is the ability to treat clock signals as normal values and the ability to wire clocks to clock input ports of registers. So if we solve the problems of casting clocks to Bits, wiring clocks to a register's clock input, and instantiating memories, you can create any clock crossing circuit.

rdaly525:
@leonardt, your syntax in __init__ also seems reasonable. I could potentially imagine problems if you combine your syntax with the 'adding clk to __call__' option though.

splhack:
This PR fixes those two issues. It unblocks sequential to use ordinary cases(one clock for `__init__` register and zero-to-multiple clocks inside `__call__`) because we can now pass m.Clocks to sequential `__call__` and wire it up.

Once this PR is landed, for the time being, until new sequential design came up with, we can use combinational instead of sequential for the other use cases, such as using multiple clocks for registers.

This PR unblocks to use sequential for more cases than the current state. such as allowing Memory inside sequential `__call__`!

phanrahan:
@leonardt I like your proposal. My sense is should allow explicit declaration of clocks (in which case we don't add additional clock signals), we allow clock signals be wired up to registers as well as combinationally, and that we have good default clock wiring if the user has not already wired up the clocks. I think if there is not an obvious good default, we should force the user to wire up the clocks.

I think these comments are all addressed by your design.

splhack:
`__call__` can use `self.READ_CLK` and `self.WRITE_CLK`, derived from `ClockInterface`. am I right?

    @sequential
    class FIFO:
        ClockInterface = {
            "READ_CLK": m.In(m.Clock),
            "WRITE_CLK": m.In(m.Clock)
        }

        def __init__(self):
             self.memory = DefineReadWriteMemory(width, height)
             # wire clocks up
             self.memory.READ_CLK @= self.READ_CLK
             self.memory.WRITE_CLK @= self.WRITE_CLK

        def __call__(self, enq_val: m.Bit, enq_rdy: m.Bit, ...):
             # it can use self.READ_CLK, self.WRITE_CLK
             thing1(CLK=self.WRITE_CLK)
             thing2.CLK @= self.READ_CLK

If `__call__` wanted to use the 3rd clock, which is the way? I guess it would be B?

A. via ClockInterface

    @sequential
    class FIFO:
        ClockInterface = {
            "READ_CLK": m.In(m.Clock),
            "WRITE_CLK": m.In(m.Clock),
            "THIRD_CLK": m.In(m.Clock)  # It will be used only in __call__
        }

        def __init__(self):
             self.memory = DefineReadWriteMemory(width, height)
             # wire clocks up
             self.memory.READ_CLK @= self.READ_CLK
             self.memory.WRITE_CLK @= self.WRITE_CLK
             # self.THIRD_CLK is not used here

        def __call__(self, enq_val: m.Bit, enq_rdy: m.Bit, ...):
             # it can use self.READ_CLK, self.WRITE_CLK, and self.THIRD_CLK

B. via `__call__` argument

    @sequential
    class FIFO:
        ClockInterface = {
            "READ_CLK": m.In(m.Clock),
            "WRITE_CLK": m.In(m.Clock),
        }

        def __init__(self):
             self.memory = DefineReadWriteMemory(width, height)
             # wire clocks up
             self.memory.READ_CLK @= self.READ_CLK
             self.memory.WRITE_CLK @= self.WRITE_CLK

        def __call__(self, THIRD_CLK: m.Clock, enq_val: m.Bit, enq_rdy: m.Bit, ...):
             # it can use self.READ_CLK, self.WRITE_CLK, and self.THIRD_CLK

And the next question is, what would the call site look like? I guess, ClockInterface turns into the positional argument in front of `__call__` arguments like this, to keep the current calling convention

    fifo = FILE(read_clk, write_clk, third_clk, enq_val, enq_rdy, ...)

I like keyword argument more (and it is still possible in this case)

    fifo = FILE(
        READ_CLK=read_clk,
        WRITE_CLK=write_clk,
        THIRD_CLK=third_clk,
        enq_val=enq_val,
        enq_rdy=enq_rdy,
        ...
    )

Or should we connect separately?

    fifo = FILE(third_clk, enq_val, enq_rdy, ...)
    fifo.READ_CLK @= read_clk
    fifo.WRITE_CLK @= write_clk

Or Clock is not supposed to be in call args?

    fifo = FILE(enq_val, enq_rdy, ...)
    fifo.READ_CLK @= read_clk
    fifo.WRITE_CLK @= write_clk
    fifo.THIRD_CLK @= third_clk

splhack:
So now this PR unblocks `__call__` to use clock(s).

    @sequential(async_reset=True)
    class A:
        def __init__(self)
            self.reg1 : m.Bit = m.bit(1)
            # Implicit clock wiring with the default clock, CLK
            # self.reg1.CLK @= CLK

       def __call__(self, ..., read_clk: m.Clock, write_clk: m.Clock) -> ...
           mem = DualPorted2ClkRam(..., read_clk, write_clk)
           if ...:
               self.reg1 = m.Bit(0)
           else:
               self.reg1 = self.reg1

    # call site can wire up clocks explicitly
    a = A()(
        ...,
        read_clk=read_clk,
        write_clk=write_clk,
        CLK=third_clk
    )

    or

    a = A()
    ...
    a.I.read_clk @= read_clk
    a.I.write_clk @= write_clk
    a.I.CLK @= third_clk

And the 2nd part of this PR allows passing different clock to `__init__`.

    @sequential(async_reset=True)
    class A:
        def __init__(self, register_clk: m.Clock)
            self.reg1 : m.Bit = m.bit(1)
            # Implicit clock wiring for now
            # self.reg1.CLK @= register_clk

       def __call__(self, ..., read_clk: m.Clock, write_clk: m.Clock) -> ...
           mem = DualPorted2ClkRam(..., read_clk, write_clk)
           if ...:
               self.reg1 = m.Bit(0)
           else:
               self.reg1 = self.reg1

    a = A()(
        ...,
        read_clk=read_clk,
        write_clk=write_clk,
        register_clk=third_clk
    )

    or

    a = A()
    ...
    a.I.read_clk @= read_clk
    a.I.write_clk @= write_clk
    a.I.register_clk @= third_clk

the code would look like this if we allows explicit clock wiring in `__init__`

    @sequential(async_reset=True)
    class A:
        def __init__(self, register_clk: m.Clock, read_clk: m.Clock, write_clk: m.Clock)
            self.reg1 : m.Bit = m.bit(1)
            self.reg1.CLK @= register_clk

            self.mem = DualPorted2ClkRam(..., RCLK=read_clk, WCLK=write_clk)
            # or
            self.mem = DualPorted2ClkRam(...)
            self.mem.RCLK @= read_clk
            self.mem.WCLK @= write_clk

       def __call__(self, ..., fourth_clk: m.Clock) -> ...
           if ...:
               self.reg1 = m.Bit(0)
           else:
               self.reg1 = self.reg1

    a = A()(
        ...,
        register_clk=third_clk,
        read_clk=read_clk,
        write_clk=write_clk,
        fourth_clk=fourth_clk,
    )

    or

    a = A()
    ...
    a.I.register_clk @= third_clk
    a.I.read_clk @= read_clk
    a.I.write_clk @= write_clk
    a.I.fourth_clk @= fourth_clk

And probably we may want to show warning for the implicit clock wiring if multiple clocks are available, since it is not clear. 

    def __init__(self, read_clk: m.Clock, write_clk: m.Clock, CLK: m.Clock)
        self.reg1 : m.Bit = m.bit(1)
        # which clock was wired up implicitly?

splhack:
So I am now able to get rid of a lot of lines of code by using sequential with #538 (I can put Product(named tuple) to implicit register!), but the last and biggest blocker is this Clock interface in `__call__`.

Any thoughts on the idea https://github.com/phanrahan/magma/pull/524#issuecomment-573446310 ?

    @sequential(async_reset=True)
    class A:
        def __init__(self, read_clk: m.Clock, write_clk: m.Clock):
            self.mem = DualPorted2ClkRam(..., RCLK=read_clk, WCLK=write_clk)

        def __call__(self, third_clk: m.Clock):
            ...

leonardt:
Just had a chance to catch up to this, I'm liking the look of this so far.   Some questions:
* In https://github.com/phanrahan/magma/pull/524/files#diff-28b2716751b662bc7f7367419975acdaR473-R475 `CLK` is defined in the `__call__`, does this refer to the implicit clock generated for the registers in `__init__`?  I think it does judging by https://github.com/phanrahan/magma/pull/524/files#diff-d67611b383595d87691d67e0dfab03f4R280.  Would this mean that if i chose to name it `CLK2`, it would automatically use this name? Actually, does this just rely on the fact that the clock is automatically wired up by type? If so, then just one clock type with any name should be okay.
* https://github.com/phanrahan/magma/pull/524/files#diff-28b2716751b662bc7f7367419975acdaR486 In this case is the first `CLK` used to wire up the registers from __init__? If it's just use the automatic wiring pass, then I believe that's the logic it uses.

I think this looks like a reasonable start, we can add support for wiring multiple clocks in `__init__` as a later step (once we add support for generator code inside `__init__` which is a forthcoming feature)

leonardt:
Just to be complete, can you add an example/test of how you would now use this feature to pass through the clock to your desired internal unit? (e.g. a stubbed example of using a memory or fifo, maybe even with multiple clocks)

splhack:
Let me to clarify a bit first. For magma circuit, we have `ClockInterface`. It will add `CLK` implicitly.

    IO = ["I", m.In(m.Bit), "O", m.Out(m.Bit)] + m.ClockInterface()

Unfortunately we can't use that for combinational and sequential,

    @m.circuit.combinational
    def test(value: m.Bits[8]) + m.ClockInterface() -> m.Bits[8]:

    # sequential
    def __call__(self, value: m.Bits[8]) + m.ClockInterface() -> m.Bits[8]:

So, that's why my first idea was kinda mocking `ClockInterface` using decorator.

    @m.circuit.combinational(clock=True)
    def test(value: m.Bits[8]) -> m.Bits[8]:

Let's think about only `CLK` in sequential `__call__`. sequential is already using `CLK` in `__init__`. So, we can just propagate it to `__call__`.

    def __call__(self, value: m.Bits[8]) -> m.Bits[8]:
        ram = RAM(width=width, height=height)
        # ram.CLK is connected with CLK implicitly

But it doesn't allow to use `CLK` explicitly. `CLK` is not defined in the `__call__` context.
Also sequential now can exist without `__init__`. The outer circuit of `__call__` isn't guaranteed to have `CLK`. (related to #546)
On top of these, no options to use `CLK` in combinational. 

    @m.circuit.combinational
    def test(value: m.Bits[8]) -> m.Bits[8]:
        ram = RAM(width=width, height=height)
        # no CLK..

Thus, one of the solution is allowing `m.Clock` in the argument of sequential `__init__`, `__call__`, and combinational. This is phase one. Which is this pull request.

After `we add support for generator code inside __init__`, no longer need to declare RAM in `__call__`, we should do it in `__init__` like this. All required clocks are supposed to be in `__init__` argument.

    @sequential(async_reset=True)
    class A:
        def __init__(self, read_clk: m.Clock, write_clk: m.Clock):
            self.mem = DualPorted2ClkRam(..., RCLK=read_clk, WCLK=write_clk)

Then, most of the use case, we don't need to have any clock in `__call__` argument.

So these code are just for unit test. Not for the actual use case.
- https://github.com/phanrahan/magma/pull/524/files#diff-28b2716751b662bc7f7367419975acdaR473-R475
- https://github.com/phanrahan/magma/pull/524/files#diff-28b2716751b662bc7f7367419975acdaR486

However we still need to pass clock in combinational argument.

    @m.circuit.combinational
    def test(value: m.Bits[8], CLK: m.Clock) -> m.Bits[8]:
        ram = RAM(width=width, height=height)
        # ram.CLK is connected with CLK implicitly

splhack:
I'm happy to write test using RAM/memory/etc but it can be done in magma? I guess it need to be in mantle?

leonardt:
@splhack I would just make a simple stub of a RAM by declaring the circuit IO (ports).  Don't need to do a functional test (we can add that to mantle next), but just make sure the ability to wire up the clock ports properly.

Alternatively, you could add the tests to mantle and base them off this branch.

splhack:
rebased onto master and added memory examples in test. @leonardt let me know if anything I can help here.

splhack:
with combinational2/sequential2 interface #668, this is now really simple 🎉

    diff --git a/magma/syntax/sequential2.py b/magma/syntax/sequential2.py
    index dc42237b..60f3ffbe 100644
    --- a/magma/syntax/sequential2.py
    +++ b/magma/syntax/sequential2.py
    @@ -40,7 +40,7 @@ def sequential2(**clock_io_kwargs):
             if "self" in cls.__call__.__annotations__:
                 raise Exception("Assumed self did not have annotation")

    -        io_args = build_io_args(cls.__call__.__annotations__)
    +        io_args = build_io_args(cls.__call__.__annotations__, skip_clk=True)

             class SequentialCircuit(Circuit):
                 name = cls.__name__
    diff --git a/magma/syntax/util.py b/magma/syntax/util.py
    index 748c228e..75aa10a8 100644
    --- a/magma/syntax/util.py
    +++ b/magma/syntax/util.py
    @@ -4,9 +4,10 @@ import ast

     from ..t import In, Out
     from ..circuit import Circuit
    +from ..clock import Clock


    -def build_io_args(annotations):
    +def build_io_args(annotations, skip_clk=False):
         io_args = {}

         for param, annotation in annotations.items():
    @@ -18,6 +19,10 @@ def build_io_args(annotations):
                 else:
                     io_args["O"] = annotation
                 continue
    +        elif skip_clk and param == "CLK":
    +            assert annotation is Clock, "Expected CLK type is m.Clock"
    +            # Skip to add CLK since it will be added by ClockIO
    +            continue
             annotation = In(annotation)
             io_args[param] = annotation
         return io_args

sequential2 `__call__` can accept `CLK` argument to unblock code to use `CLK` explicitly

    class DualClockRAM(m.Circuit):
        IO = [
            "RADDR", m.In(m.Bits[8]),
            "WADDR", m.In(m.Bits[8]),
            "WDATA", m.In(m.Bits[8]),
            "RDATA", m.Out(m.Bits[8]),
            "WE", m.In(m.Bit),
            "RCLK", m.In(m.Clock),
            "WCLK", m.In(m.Clock),
        ]
        @classmethod
        def definition(io):
            m.wire(m.bits(0, 8), io.RDATA)

    @m.sequential2()
    class TestDualClockRAM:
        def __call__(
            self,
            ADDR: m.Bits[8],
            WDATA: m.Bits[8],
            WE: m.Bit,
            CLK: m.Clock, # explicit CLK argument
            WCLK: m.Clock,
        ) -> m.Bits[8]:
            rdata = DualClockRAM()(
                RADDR=ADDR, WADDR=ADDR, WDATA=WDATA, WE=WE,
                RCLK=CLK, # explicit CLK usage
                WCLK=WCLK
            )
            return rdata

PR #525, opened 2020-01-07
Fix typo and add unit tests

splhack:
The main purpose of this pull request is to ask a question 🙂
How to convert SInt to UInt and the opposite?
Actually, just commenting out `isinstance(value, SIntType/UIntType) if block` in `sint/uint` works.
Also the doc doesn't mention about the limitation.
https://github.com/phanrahan/magma/blob/master/docs/conversions.md

### DIFF
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -177,7 +177,7 @@ def uint(value, n=None):
 
 def sint(value, n=None):
     if isinstance(value, UIntType):
-        raise ValueError("uint cannot convert SInt")
+        raise ValueError("sint cannot convert UInt")
     return convertbits(value, n, SIntType, SInt, True)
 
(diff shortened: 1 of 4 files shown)


phanrahan:
@splhack Have you read this discussion https://github.com/leonardt/hwtypes/issues/98? Add your thoughts. 

We should add the rules we want for conversions.

leonardt:
Thanks for the tests, @phanrahan I believe we verbally discussed that:
* we will allow explicit conversions between sint <-> uint using the conversions operators
* we won't do the implicit conversions for binary operators (e.g. trying to do <uint> + <sint>)

I can update the code to reflect this.

phanrahan:
Sounds great!

Issue #495, opened 2019-11-15
[rfc] Using @ or @= for wiring instead of <=


leonardt:
There's an annoying issue with precedence for `<=`.

When writing `io.a <= b == c`, you have to write `io.a <= (b == c)`.

Furthermore, it's an overloaded operator (means two different things for different types of values), which can lead to confusion.

We're interested in adding support for "wiring" into hwtypes, but @cdonovick has requested that we use `@` or `@=` as the operator, as it makes some analyses easier (don't need to reason about the in/out qualifiers on types).  Furthermore, `@` and `@=` currently don't have any use in HDLs, so it's unlikely to be confusing.  Finally, it's new in Python, so it doesn't have much baggage in terms of user expectations.

Any comments? We can gracefully deprecate `<=` so we don't break existing code if we chose to do this.  I think the one major concern is `@` seems a bit less elegant/familiar than `<=` (non-blocking assignment).

phanrahan:
Using `@=` makes a lot of sense to me. We discussed that, but I don't remember what we concluded.

I think the precedence o `@=` is better too. Right?

cdonovick:
Related https://github.com/leonardt/hwtypes/issues/102

cdonovick:
If we make this change I would be happy to write a refactoring tool to automatically upgrade any legacy magma code to use `@=` instead of `<=` if we okay this change.  I think the other option would be to use `__setitem__` for wiring `x <= y` becomes `x[:] = y` .

PR #526, opened 2020-01-08
Add support for __imatmul__ wiring

cdonovick:


cdonovick:
implements #495

magma/t.py
```diff
@@ -73,6 +73,7 @@ def __le__(self, other):
         else:
             raise TypeError(f"Cannot use <= to assign to output: {self.debug_name} (trying to assign {other.debug_name})")
```
leonardt: Should we update this error message? (And maybe add a deprecation warning? Or maybe wait to do this)

cdonovick: Heh was going to add another commit to address the error message but then decided I didn't actually want to write new code.   I will fix.

Deprecation warning is up to you.

leonardt: I'll add it as part of the new release after hwtypes2 is merged

Issue #528, opened 2020-01-12
Performance issue


splhack:
I'm observing performance regression on master. master takes 196 seconds to compile, although a bit old hwtypes2 takes 3 seconds. will bisect later, or create a minimal repro code. probably related to namedtuple vs Product.

Versions
  - Python 3.7.5 (ubuntu 18.04, python3.7 package)
  - self build coreir (C++) from github master, should be same as binary wheel
  - `pip3 install -e` . with github master
    - coreir             2.0.29
    - fault              2.0.22
    - hwtypes            1.3.1
    - loam               0.1a0
    - mantle             1.0.12
    - pyverilog          1.2.0

compile.py

    set_mantle_target("coreir")
    main = getattr(__import__("top"), "main")
    compile("build/top", main, output="coreir-verilog", verilator_debug=True)

### python3 -m cProfile compile.py with a bit old hwtypes2
- the base commit is 6520b7c283944f6cb0d66390d976a9d0d31693a5
- with several fixes which I mentioned in https://github.com/phanrahan/magma/pull/515, these exact same fixes are now in master
- and https://github.com/phanrahan/magma/pull/524

result

             5286265 function calls (5046047 primitive calls) in 3.070 seconds

       Ordered by: call count

       ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    975501/945911    0.081    0.000    0.218    0.000 {built-in method builtins.isinstance}
    523332/523314    0.080    0.000    0.083    0.000 {built-in method builtins.hasattr}
       504223    0.065    0.000    0.090    0.000 inspect.py:63(ismodule)
       456967    0.027    0.000    0.027    0.000 {method 'get' of 'dict' objects}
       163085    0.010    0.000    0.010    0.000 {method 'append' of 'list' objects}
       129510    0.021    0.000    0.024    0.000 {built-in method builtins.getattr}
       126032    0.013    0.000    0.013    0.000 array.py:81(N)
    121490/119523    0.008    0.000    0.010    0.000 {built-in method builtins.len}
        90873    0.008    0.000    0.008    0.000 array.py:85(T)
        90422    0.007    0.000    0.007    0.000 digital.py:125(__eq__)
        85449    0.026    0.000    0.126    0.000 {built-in method _abc._abc_instancecheck}
        85449    0.012    0.000    0.138    0.000 abc.py:137(__instancecheck__)0

### python3 -m cProfile compile.py with master
[...]

PR #530, opened 2020-01-13
Simplify base class logic for tuple

leonardt:
Fixes https://github.com/phanrahan/magma/issues/528#issuecomment-573510435

Before, the base class logic for the tuple meta class was iterating over the product of the field bases.  This was used to make sure that `Tuple[Bit, Bit]` was a parent of `Tuple[In(Bit), In(Bit)`, however this general logic meant it was also checking the cross product of the possibilities, e.g. `Tuple[Bit, In(Bit)]` as candidates for a base class.

This simplifies the pipeline to simply check if any of the fields are directed.  If so, it sets the class with the undirected version of the fields as the parent.

It also removes the logic from the Product type meta class since it's able to use the Tuple meta class logic.

Finally, it updates the conversion code (tuple/namedtuple) to call the type constructors directly rather than using `type(...)` which was skipping a portion of the meta class pipeline (revealed when changing the subclass logic)

### DIFF
--- a/magma/tuple.py
+++ b/magma/tuple.py
@@ -70,15 +70,11 @@ class TupleKind(TupleMeta, Kind):
         if cls.is_bound:
             raise TypeError('Type is already bound')
 
-        bases = []
-        bases.extend(b[idx] for b in cls.__bases__ if isinstance(b, BoundMeta))
-        idx_bases = [[b for b in i.__bases__ if isinstance(b, type(i))] for i
-                     in idx]
-        for i_b in itertools.product(*idx_bases):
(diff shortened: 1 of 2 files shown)


leonardt:
@splhack if possible, can you confirm this branch fixes your performance issue on your original design? It seems to compile the test code you provided quickly.

splhack:
@leonardt thanks! verified this PR fixed the performance issue!

This is the cProfile result with this PR, for comparing with #528 

>          5795999 function calls (5487532 primitive calls) in 3.269 seconds

PR #532, opened 2020-01-13
Support 2d array to verilog

leonardt:
We support importing verilog with 2d arrays, but we can't compile them out using the verilog backend.

This simply moves up the logic to emit a module defined by a pure verilog string to emit the string before the type safety check of the interface.

The current use case is using imported verilog with fault (which may have n-d arrays) but compiling through the verilog backend.

Adding support for these modules to coreir is more involved since we'll need to handle the n-d arrays as non-flattened types which will touch a lot of code (right now everything assumes that types including n-d arrays are flattened out before the coreir verilog backend code generation)

### DIFF
--- a/magma/backend/verilog.py
+++ b/magma/backend/verilog.py
@@ -1,4 +1,3 @@
-import warnings
 import os
 import types
 import operator
--- a/magma/backend/verilog.py
+++ b/magma/backend/verilog.py
@@ -1,3 +1,4 @@
+import warnings
 import os
 import types
 import operator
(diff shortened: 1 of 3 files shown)


magma/backend/verilog.py
```diff
@@ -165,72 +168,69 @@ def compiledefinition(cls):
                 raise Exception(f'Argument {cls.__name__}.{name} of type {type(port)} is not supported, the verilog backend only supports simple 1-d array of bits of the form Array(N, Bit)')
 
 
-    if cls.verilogFile:
-       s = cls.verilogFile
```
leonardt: This is an awkward diff, but it basically removes the `if cls.verilogFile` condition, moves it up and uses a return, and removes the else case (causing the re-indentation and big diff)

PR #533, opened 2020-01-13
Support bit width promotion in Bits constructor

leonardt:
Fixes issue in lassen where the Bits constructor (or it's children) cannot be used to promote the width of the input value (e.g. `Bits[16](Bits[1](1))`).  This adds the corresponding code and tests

### DIFF
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -66,6 +66,17 @@ class BitsMeta(AbstractBitVectorMeta, ArrayMeta):
 class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
     __hash__ = Array.__hash__
 
+    def __init__(self, *args, **kwargs):
+        if args and len(args) == 1 and isinstance(args[0], m.Array) and \
+                len(self) > 1 and len(args[0]) <= len(self):
+            self.ts = args[0].ts[:]
+            # zext for promoting width
+            for i in range(len(self) - len(args[0])):
+                self.ts.append(m.VCC)
+            Type.__init__(self, **kwargs)
+        else:
+            Array.__init__(self, *args, **kwargs)
+
     def __repr__(self):
         if not isinstance(self.name, AnonRef):
             return repr(self.name)
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -72,7 +72,7 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
             self.ts = args[0].ts[:]
             # zext for promoting width
             for i in range(len(self) - len(args[0])):
-                self.ts.append(m.VCC)
+                self.ts.append(m.GND)
             Type.__init__(self, **kwargs)
         else:
             Array.__init__(self, *args, **kwargs)
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -498,6 +498,17 @@ class UInt(Bits):
 
 
 class SInt(Bits):
+    def __init__(self, *args, **kwargs):
+        if args and len(args) == 1 and isinstance(args[0], m.Array) and \
+                len(self) > 1 and len(args[0]) <= len(self):
+            self.ts = args[0].ts[:]
+            # zext for promoting width
+            for i in range(len(self) - len(args[0])):
+                self.ts.append(args[0].ts[-1])
+            Type.__init__(self, **kwargs)
+        else:
+            Array.__init__(self, *args, **kwargs)
+
     @bits_cast
     def bvslt(self, other) -> AbstractBit:
         return self.declare_compare_op("slt")()(self, other)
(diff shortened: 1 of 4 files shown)


magma/bits.py
```diff
@@ -66,6 +66,17 @@ def __str__(cls):
 class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
     __hash__ = Array.__hash__
 
+    def __init__(self, *args, **kwargs):
```
cdonovick: SMT / py support implicit extension and truncations.   I am fine with blocking this behavior and require people to extract or extend.  However, if we are going to replicate in magma then lets be consistent.

magma/bits.py
```diff
@@ -66,6 +66,17 @@ def __str__(cls):
...
+        if args and len(args) == 1 and isinstance(args[0], m.Array) and \
+                len(self) > 1 and len(args[0]) <= len(self):
+            self.ts = args[0].ts[:]
+            # zext for promoting width
```
cdonovick: py/smt also use zext but I think this might not be the most obvious behavior.  We should ext so that signed values are extended correctly. Or perhaps we should use the ext of the current type? not sure.

leonardt: That makes sense, I can update the SInt constructor sext for now.  We can always remove this code if we decide to not support the behavior, but it seems reasonable for now and the lassen code uses that behavior

magma/bits.py
leonardt: I think in magma we decided to support implicit extension but not truncation 

CC @phanrahan, what do you think the behavior of calling `Bits[16](a)` when a is of type `Bits[2]` should be? This PR will zext for Bits/UInt and sext for SInt.  Do we want to support truncation as well? E.g. calling `Bits[5](a)` when a is of type `Bits[16]` and just dropping the upper bits?

phanrahan:
Implicitly changing sizes seems dangerous to me. Especially, downsizing. I suppose upsizing is ok, but I think we should preserve values which means zext.

leonardt:
Okay, @cdonovick I can remove this logic but then we'll have to update the lassen code to extend where before it was doing this implicit sizing.

phanrahan:
If we are downsizing, I vote for changing the lassen code.

leonardt:
I don't think lassen is downsizing, just implicit extension (e.g. using `Bits[16](a)` to extend to 16 bits instead of `m.zext(a, 15)`

cdonovick:
I think if we extend we should use the `ext` method of the new type.  So that: `T[size](val) == T[size](S[size'](val))`.  In other words `zext` when `T` is `Bits` / `UInt` and `sext` is `SInt`.  If not that we should use  the `ext` of `S`.  

I think blindly using `zext` in all cases is wrong as it would lead to `SInt[8](-1) != SInt[8](SInt[7](-1))`

phanrahan:
As I said, the key is to preserve values. Preserving values means the new type can hold all the values of the old type. So that means if we are extending an sint, we should use sext.

cdonovick:
@phanrahan So in that case we should use the `ext` method of the original type (`S` in my example).

Specifically which of these statements should true:
```
SInt[8](UInt[4](-1)) == SInt[8](-1)
SInt[8](UInt[4](-1)) == SInt[8](15)
```
It sounds like you want the second option.

Which is not what is currently implemented, but not a complex change.

phanrahan:
As I mentioned before, I don't think saying uint(-1) is a good idea. I would rather right that as uint(sint(-1)),
where the conversion from sint to uint is a bit copy. That makes it clear that we are counting on the fact that we are using two's complement. But assuming uint(-1,4) = uint(15,4). Then sint(uint(15)) should be sint(15).

cdonovick:
@phanrahan I am confused as why `uint(-1)` is bad but `uint(sint(-1))` is fine. It's not hard to block the behavior and I will if it's something you feel strongly about.  However, it seems a bit unintuitive to allow casts from signed x < 0 to unsigned but not int x < 0 to unsigned. 

Personally I think `uint(-1)` is a common / useful idiom.

phanrahan:
It just bugs me to assume twos-complement. Old school I guess. There are many other representations of numbers.

This is ok with me!
```
SInt[8](Bits[4](0b1111)) == SInt[8](0xf)
```

cdonovick:
> It just bugs me to assume twos-complement.

Well, I can't argue with that.  I am usually the one preaching don't assume a specific encoding.

I guess the question is does hwtypes require twos-complement and if not is casting from sint to uint a well defined operation?

phanrahan:
I am happy with hwtypes being defined to use two's complement. I just think we shouldn't be encouraging people to make that assumption unless there is no way around it.

PR #534, opened 2020-01-14
Bugfixes for lassen

leonardt:
Depends on fix-tuple-perf

Issues:
* Implicit promotion (extension) in Bits/SInt/UInt constructor (see discussion here: https://github.com/phanrahan/magma/pull/533 we may drop this and change the lassen code)
* Bugfix in `mux` code generation (`Bit.ite` assumed the inputs were `Bit`, but `Bit.ite` can be used on BitVector types, so the generator dispatches on the type)

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -8,7 +8,6 @@ from .bit import VCC, GND, Bit
 from .bitutils import seq2int
 from .debug import debug_wire, get_callee_frame_info
 from .port import report_wiring_error
-import hwtypes
 
 
 class ArrayMeta(ABCMeta, Kind):
@@ -337,8 +336,6 @@ class Array(Type, metaclass=ArrayMeta):
         return sum([t.flatten() for t in self.ts], [])
 
     def concat(self, other) -> 'AbstractBitVector':
-        if isinstance(other, hwtypes.BitVector):
-            other = m.Bits[len(other)](other.as_uint())
         return type(self)[len(self) + len(other), self.T](self.ts + other.ts)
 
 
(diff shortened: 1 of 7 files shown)


cdonovick:
We might want unify on a set of typing rules for ite.  The typing I use can be found:
https://github.com/leonardt/hwtypes/blob/master/hwtypes/bit_vector.py#L81

I am going to add support for python tuples (to address https://github.com/leonardt/ast_tools/issues/30) where the type resolution will just be a recursive version of this. 

Any thoughts?

phanrahan:
This is related to data flow analysis. In data flow analysis the return type of a if-then-else is the join of the return types of each branch. That is, it should be a type that can store all the values of each of the sub-types.

Personally, I would be happier if we did no casting and requires that type(t_value) == type(f_value).

I don't like implicit type casts. I suggest we be very cautious.

cdonovick:
I am fine with requiring strict type equality, it would differ from the behavior of basically every other operation though as most allow implicit casts:
```
class TaggedSigned(Signed): pass
x = Signed[8]()
y = TaggedSigned[8]()
assert type(x + y) == type(y + x) == Signed[8]
assert type(y + y) == TaggedSigned[8]
```
Granted `ite` is a bit different as the type of the operands do not usually match the type of `self`, so I don't feel strongly.

@leonardt what do you think?

phanrahan:
I feel we are going down a very slippery slope with all these implicit conversions.

cdonovick:
@phanrahan I would thing of them more as implicit upcasts rather than convestions

PR #538, opened 2020-01-15
Support Product in sequential implicit register

splhack:
Unblock sequential to use Product as implicit register

    class TestChildProduct(m.Product):
        test = m.Bit

    class TestProduct(m.Product):
        test_child = TestChildProduct

    @m.circuit.sequential(async_reset=False)
    class Test:
        def __init__(self):
            self.test_product: TestProduct = m.namedtuple(
                test_child=m.namedtuple(
                    test = m.bit(0)
                )
            )

It generates

    // register definition
    test_product = {
        'test_child': {
            'test': DefineRegister(None, init=0, has_async_reset=False)()
        }
    }

    // call args
    comb_out = Test(
        self_test_product_O=m.namedtuple(
            test_child=m.namedtuple(
                test=test_product['test_child']['test'].O
            )
        )
    )

    // comb wiring
    test_product['test_child']['test'].I <= comb_out[0].test_child.test


splhack:
sure happy to do that🙂

PR #539, opened 2020-01-18
Support for SV bind statement

leonardt:
This is a temporary solution for enabling the definition of verification modules that are connected to a generated module using the system verilog bind statement.

It allows the user to us a method `bind` which binds a module as a monitor and allows passing through intermediate signals in the bind statement.  By default, the bind will pass through the entire IO, so it assumes that the monitor module's interface is a superset of the bound module's interface.  

See `tests/test_verilog/test_bind.py` for an example usage.

The current flow checks whether a circuit has bound modules, if it does, it emits the compiled module, followed by a bind statement in a separate file (named after the module).  Currently it puts this file in the same directory as the compiled main circuit.

One thing to consider: handling of bind_modules in coreir backend.  This basically has to register all the bind modules we've seen in compiling the current circuit, then emit them at the end in the same directory.  This was simpler than passing this through the coreir backend, but right now it uses a patched dictionary on the coreir backend, there's probably a cleaner/more robust way to handle this.


magma/backend/coreir_transformer.py
```diff
@@ -155,6 +156,18 @@ def run_self(self):
...
+        for bind_module, bind_stmt in self.defn.bind_modules:
+            if not os.path.isdir(".magma"):
+                os.mkdir(".magma")
+            curr_compile_dir = m.config.get_compile_dir()
+            m.config.set_compile_dir("normal")
+            m.compile(f".magma/{bind_module.name}", bind_module,
+                      output="verilog")
+            m.config.set_compile_dir(curr_compile_dir)
+            with open(f".magma/{bind_module.name}.v", "r") as f:
+                self.backend.sv_bind_files[bind_module.name] = f.read() + \
```
rsetaluri: nit: prefer

```
content = "\n".join((f.read(), bind_stmt))
self.backend.sv_bind_files[bind_module.name] = content
```

magma/backend/coreir_transformer.py
```diff
@@ -3,6 +3,7 @@
 import json
 import logging
 import os
+import magma as m
```
rsetaluri: We should avoid this in internal code. Looks like you need:

```
from ..compile import compile
from ..config import get_compile_dir, set_compile_dir
```

If the problem is circular imports, then we do something like

```python
from .verilog import VerilogCompiler

...

VerilogCompiler(bind_module, f".magma/{bind_module.name}").compile()
```

In fact I think this is preferred anyway.

tests/test_verilog/test_bind.py
```diff
@@ -0,0 +1,63 @@
+import magma as m
+import magma.testing
+
```
rsetaluri: nit: extra whitespace

PR #549, opened 2020-01-28
Bug545

### LINKED ISSUE #545: circuit.sequential expects 'm' to be defined as magma in local environment
```
import magma
import fault
def test_magma():

    @magma.circuit.sequential
    class PE:
        def __call__(self, in0 : magma.Bit, in1: magma.Bit) -> magma.Bit:
            return in0 & in1

    tester = fault.Tester(PE)
    vals = [0, 1]
    for i0, i1 in itertools.product(vals, vals):
        tester.circuit.in0 = i0
        tester.circuit.in1 = i1
        tester.eval()
        tester.circuit.O.expect(i0 & i1)
    tester.compile_and_run("verilator")
```

Errors with:
```
>       circuit_def = circuit_def_constructor(combinational_decorator)
E       NameError: name 'm' is not defined

../magma/magma/syntax/sequential.py:438: NameError
```

cdonovick:
Should resolve #545 

Test in its own file because its needs a clean env.

### DIFF
--- a/magma/syntax/sequential.py
+++ b/magma/syntax/sequential.py
@@ -13,6 +13,7 @@ from collections import Counter
 import itertools
 
 from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT, SymbolTable
+from ast_tools import gen_free_name
 
 class RewriteSelfAttributes(ast.NodeTransformer):
     def __init__(self, initial_value_map):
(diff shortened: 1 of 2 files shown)


cdonovick:
Failing because of lack of mantle.  @leonardt how do you test sequential without mantle?

leonardt:
test_sequential defines a custom register generator (stub) so it's already populated into the environment (the sequential logic checks to see if DefineRegister exists, else it imports it from mantle).

I would recommend just importing the stub function from the `test_sequential.py` file (https://github.com/phanrahan/magma/blob/91771151df045b6b096120179b89eafadd3a4a3d/tests/test_syntax/test_sequential.py#L98).

You'll probably also need the `phi` stub from `test_combinational.py` (https://github.com/phanrahan/magma/blob/91771151df045b6b096120179b89eafadd3a4a3d/tests/test_syntax/test_sequential.py#L3) but you can also just import that from test_sequential.

Issue #548, opened 2020-01-28
Enum classes are not subclass of magma.Enum


rdaly525:
```
import magma as m
class A(m.Enum):
  a = 1
  b= 2
assert issubclass(A, m.Enum) #This fails
```

Is there any way to determine if a type is a magma enum?

leonardt:
I'll fix this, but for now, there's an attribute `_is_magma_enum` that you could check https://github.com/phanrahan/magma/blob/e69d0d89a9bf80644f501a7bdb1a5cd440756550/magma/enum.py#L20

Issue #500, opened 2019-11-28
Issue with magma.Enum having a field 'N'


rdaly525:
The following code throws an error

```
class A(m.Enum):
  a=0
  N=1

```

I would prefer that the letter 'N' works, but if that is not possible, I would prefer an error message indicating reserved fields.

PR #551, opened 2020-01-29
Make user defined Enums subclass a m.Enum

leonardt:
Fixes https://github.com/phanrahan/magma/issues/548 and https://github.com/phanrahan/magma/issues/500 (for now we just raise an error for `N`, in the future we can make this unreserved, but right now a lot of internal code unrelated to enums (arrays) use this attribute).

### DIFF
--- a/magma/enum.py
+++ b/magma/enum.py
@@ -1,28 +1,35 @@
 import magma as m
+from .bits import BitsMeta, Bits
 from enum import Enum as pyEnum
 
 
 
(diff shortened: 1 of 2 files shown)


leonardt:
I'll open an issue, there is a massive list of reserved names used internally in the type system, so it'll be pretty hard to track them reasonably.

I think a better approach would be to come up with a way to use a separate namespace for enum fields so that they don't clobber those reserved names.

PR #552, opened 2020-01-29
Bug547

### LINKED ISSUE #547: New wire syntax and arrays
Was updating magma code to use the new wire syntax when I discovered a problem, wiring arrays elements doesn't work.
Consider:

```Python
arr = Array[N, T]()
arr[0] @= elem
```
Which is equivalent to:
```Python
temp = arr[0] # arr.__getitem__(0)
temp @= elem # temp.__imatmul__(elem)
arr[0] = temp # arr.__setitem__(0, temp) <--- THE PROBLEM
```

Now the old syntax worked because:
```Python
arr[0] <= elem
```
is equivalent to:
```Python
temp = arr[0] # arr.__getitem__(0)
temp <= elem # temp.__le__(elem)
```

I think I have pretty simple solution.

Currently `t.__imatmul__` returns `self`, which is probably just a stray piece of code it should logicall return `None`.  However, this is perfect and allows the following solution:

```Python
class Array(...):
    ...
    def __setitem__(self, key, val):
        old = self[key]
        if old is not val:
            raise ValueError()
```

This wouldn't change much about allowable syntax beyond permitting tautological statements like:
```Python
arr[0] = arr[0]
```

cdonovick:
fixes (partial) #547

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -222,6 +222,21 @@ class Array(Type, metaclass=ArrayMeta):
 
             return self.ts[key]
 
+    def __setitem__(self, key, val):
+        error = False
+        old = self[key]
+        if isinstance(old, Array):
+            if len(old) != len(val):
+                error = True
+            elif any(old[i] is not val[i] for i in range(len(old))):
+                error = True
+        elif old is not val:
+            error = True
+
+        if error:
+            raise ValueError('May not mutate array')
+
+
     def __add__(self, other):
         other_len = other.N
         total = self.N + other_len
--- a/magma/array.py
+++ b/magma/array.py
@@ -234,7 +234,8 @@ class Array(Type, metaclass=ArrayMeta):
             error = True
 
         if error:
-            raise ValueError('May not mutate array')
+            _logger.error(f'May not mutate array, trying to replace '
+                          f'{self}[{key}] ({old}) with {val}')
 
 
     def __add__(self, other):
--- a/tests/test_issues/test_547.py
+++ b/tests/test_issues/test_547.py
@@ -1,3 +1,5 @@
+import pytest
+
 import magma as m
 
 def test_basic():
@@ -25,4 +27,11 @@ def test_compound():
             io.O[0] @= io.I0
             io.O[1] @= io.I1
 
+def test_errors():
+    with pytest.raises(ValueError):
+        class Test(m.Circuit):
+            IO = ["I", m.In(m.Bits[1]), "O", m.Out(m.Bits[1])]
 
+            @classmethod
+            def definition(io):
+                io.O[0] = io.I[0]
--- a/tests/test_issues/test_547.py
+++ b/tests/test_issues/test_547.py
@@ -1,0 +1,28 @@
+import magma as m
+
+def test_basic():
+    class Test(m.Circuit):
+        IO = ["I", m.In(m.Bits[1]), "O", m.Out(m.Bits[1])]
+
+        @classmethod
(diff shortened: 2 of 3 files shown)


magma/array.py
```diff
@@ -222,6 +222,21 @@ def __getitem__(self, key):
...
+        if isinstance(old, Array):
+            if len(old) != len(val):
+                error = True
+            elif any(old[i] is not val[i] for i in range(len(old))):
+                error = True
+        elif old is not val:
+            error = True
+
+        if error:
+            raise ValueError('May not mutate array')
```
rsetaluri: Since this would likely arise in user facing code, let's make this a logged error: `_logger.Error("Directly calling setitem on Array disallowed")`

tests/test_issues/test_547.py
```diff
@@ -0,0 +1,49 @@
...
+
+        @classmethod
+        def definition(io):
+            io.O[0:1][:][0] @= io.I[0:1][:][0] # [0]
+            io.O[1:2][0][0:1] @= io.I[1:2][0][0:1] # [1][0]
+            io.O[1][1:2] @= io.I[1][1:2] # [1][1]
+            io.O[2][:] @= io.I[2][:]
+
+def test_errors():
+    with pytest.raises(ValueError):
```
rsetaluri: If the above code is changed to use log rather than exception, than this would change. See https://github.com/phanrahan/magma/blob/7bcc06cd5d520798ff5d34b4c499687db3d38592/tests/test_wire/test_errors.py#L4 for usage of getting logger output using `caplog`.

magma/array.py
cdonovick: /I am fine with this, however it would change current behavior. `Array[4, Bit]()[0] = Bit(0)` is a `TypeError` currently.

I also don't know how to use the logger can you point me at example code?

tests/test_issues/test_547.py
cdonovick: Probably not the right place to raise this issue but...
The way you are using caplog is testing the implementation not the interface.  Which is decidedly not agile. Unless we are saying the exact error message is part of the interface (seems dubious).

leonardt: You could just regex or match on the error string prefix if you want to avoid testing the exact message.  E.g just check for the presence of ERROR

magma/array.py
rsetaluri: Why would this depend on whether we `raise` or log?

Usage of logger: https://github.com/phanrahan/magma/blob/77689a30e595e343c8ae4030d6cb20fd0206b834/magma/port.py#L39-L40. We define a `_logger` at the top of this file, and it has methods `error`/`log`/`warning`

tests/test_issues/test_547.py
rsetaluri: Agreed. We can do `assert caplog.records[0].levelname == "ERROR"`. But either way, hard to test a "wrong" program errors as expected unless exceptions are standard in your (rather than logging; open to this too).

cdonovick: Personally I have don't really think it matters if it logs or errors, if the standard is logs I am fine with that.

I do prefer checking the `levelname == "ERROR"` to checking exact error messages.  It seems like the property we actually care about.

magma/array.py
cdonovick: @rsetaluri I was just saying it changes code that previously raised into code that logs.  Which doesn't matter to me, but it does change behavior.

PR #556, opened 2020-01-30
Support tuple/array in bind

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/833


magma/circuit.py
```diff
@@ -524,28 +530,48 @@ def place(cls, inst):
...
+            ports += cls.gen_bind_port(mon_arg, bind_arg)
         ports_str = ", ".join(ports)
         bind_str = f"bind {cls.name} {monitor.name} {monitor.name}_inst ({ports_str});"  # noqa
-        cls.bind_modules.append((monitor, bind_str))
+        if not os.path.isdir(".magma"):
+            os.mkdir(".magma")
+        curr_compile_dir = get_compile_dir()
+        set_compile_dir("normal")
+        # Circular dependency, need coreir backend to compile, backend imports
+        # circuit (for wrap casts logic, we might be able to factor that out)
```
leonardt: One way around this circular dependency would be to run a pass to compile bind modules during the `compile` phase, but I think that's more work than is desired for a temporary solution that works with this circular import logic.

tests/test_verilog/test_from_file.py
```diff
@@ -24,7 +24,7 @@ def check_rxmod(RXMOD):
     check_port(RXMOD, "data", m.Array, "output")
     check_port(RXMOD, "valid", m.Bit, "output")
 
-    m.compile("build/test_rxmod", RXMOD, output="verilog")
+    m.compile("build/test_rxmod", RXMOD)
```
leonardt: This tests compiling a top level wrapped verilog module using the coreir backend

magma/circuit.py
```diff
@@ -524,28 +530,50 @@ def place(cls, inst):
...
         cls.instances.append(inst)
 
+    def gen_bind_port(cls, mon_arg, bind_arg):
+        if isinstance(mon_arg, Tuple) or isinstance(mon_arg, Array) and \
+                not issubclass(mon_arg.T, m.Digital):
+            result = []
+            for child1, child2 in zip(mon_arg, bind_arg):
+                result += cls.gen_bind_port(child1, child2)
+            return result
+        else:
```
rsetaluri: nit: move `else` body out of `else` block and remove `else`.

magma/circuit.py
rsetaluri: You read my mind :). Can you add a note on the `import magma as m` that this is undesired and should be removed asap?

PR #558, opened 2020-01-30
Add unused/undriven primitives

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/834

Allows declaring undriven values, e.g.
```
            io.O0 @= m.Undriven
```
and unused values, e.g.
```
            m.Unused @= io.I0
```

Currently this works by hijacking the wiring logic/operators to dispatch on the wiring types, and if it finds the special class/object for Unused and Undriven, it inserts a reference to a generated coreir primitive module that is used to represent them (term/undriven).

### DIFF
--- a/magma/t.py
+++ b/magma/t.py
@@ -2,6 +2,7 @@ import functools
 import warnings
 import enum
 from abc import abstractmethod
+import magma as m
 from .ref import Ref, AnonRef, DefnRef, InstRef
 from .compatibility import IntegerTypes, StringTypes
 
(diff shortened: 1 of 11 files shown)


rsetaluri:
Understand the need for this functionality, but not sold on the design. What if instead of wiring, it was a function on types? i.e. rather than `x @= m.Undriven` -> `x.undriven()` and we can delegate that responsibility to the types. I think both (a) the syntax is better for the users, and (b) the design on the backend is cleaner (don't need specialized functionality in `def wire()`). We could make these functions part of the spec, ala if we want to add a `fill()` function.

leonardt:
Makes sense, I'll try that out to see how it looks

leonardt:
Changed unused/undriven to be methods on the type, also added them as abstractmethods on Type.  There's a wonky circular dependency in that the implementation of unused/undriven requires defining/instantiating a Circuit (term/unused primitive), but Circuit depends on Bit/Bits/etc... For now, we just use Python's dynamic import staging to work around it.  One possible way to avoid it would be to simply "mark" a value as undriven/unused and run a pass before the backend to insert the primitive modules.

magma/t.py
```diff
@@ -98,17 +99,35 @@ def debug_name(self):
 
     def __le__(self, other):
         if not self.is_output():
-            self.wire(other)
```
rsetaluri: why does this need to change? (and below)

leonardt: hmm, that was left over from the old implementation (so it would dispatch to the logic in m.wire), will remove.

PR #560, opened 2020-01-31
Product access

splhack:
source

    a = m.namedtuple(a0=value, a1=self.a.a1)

generated code and error

    a_0 = m.namedtuple(a0=io.value, a1=self.a.a1)
    AttributeError: type object 'TestProductAccess' has no attribute 'a'

fix

    a_0 = m.namedtuple(a0=io.value, a1=io.self_a_O.a1)

### DIFF
--- a/magma/syntax/sequential.py
+++ b/magma/syntax/sequential.py
@@ -26,7 +26,7 @@ class RewriteSelfAttributes(ast.NodeTransformer):
                 return ast.Name(f"self_{node.attr}_I", ast.Store())
             else:
                 return ast.Name(f"self_{node.attr}_O", ast.Load())
-        return node
+        return self.generic_visit(node)
 
     def visit_Call(self, node):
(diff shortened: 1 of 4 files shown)


magma/syntax/sequential.py
```diff
@@ -26,6 +26,8 @@ def visit_Attribute(self, node):
                 return ast.Name(f"self_{node.attr}_I", ast.Store())
             else:
                 return ast.Name(f"self_{node.attr}_O", ast.Load())
+        else:
+            self.generic_visit(node)
         return node
```
leonardt: I think this relies on the `generic_visit` modifying the node in place which isn't always guaranteed.  Can you update this to remove the `else` and say
```
return self.generic_visit(node)
```

that we always recurse if we're not doing the rewrite behavior and return the result.

splhack: sure, updated

PR #564, opened 2020-01-31
Fix issue with BitVector constructor

### LINKED ISSUE #562: Issue with casting Bit to Bits[1] within ite.
Here is a minimal failing test that I would expect to work.
```
    BV1 = m.Bits[1]
    @m.circuit.sequential
    class A:
        def __call__(self, a: m.Bit, b: m.Bits[2]) -> m.Bits[1]:
            return a.ite(BV1(a), b[0:1])
```
It fails with:
`TypeError: ite expects same type for both branches: Bits[1] != Out(Bits[1])`

I have a branch ite-bug that contains this test in test_sequential.py

leonardt:
Fixes #562 

The issue was that when the Bits constructor was used with an output (converting between a Bit output to a Bits[1] output), the result doesn't preserve the directionality (so `Out(Bit)` becomes `Bits[1]`.  Now, it checks if all the children are outputs, if so, it returns `Out(Bits[n])` instead of `Bits[n]`.

We should ensure that this is consistent with all the types, (so if you convert a directed value with an undirected type, the result should preserve the direction).  The other choice is to force the user to say `Out(Bits[1])` for their type conversion.  But this would at least unblock your issue in sequential.

CC @rsetaluri and @phanrahan any thoughts on this? The issue is, if I have an output type such as
```
a = m.Out(Bit)
```
and I convert it to a `Bits[1]` using
```
b = m.Bits[1](a)
```
I'll get an instance of `m.Bits[1]` instead of `m.Out(m.Bits[1])`.  This changes the behavior to give you `m.Out(m.Bits[1])` instead.  The alternative is to use `m.bits(a, 1)` which will give you an output, but I think that syntax is not compatible with peak.  The general issue is: when using an undirected constructor with a value, should the result return a directed value?

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -342,7 +342,7 @@ class Array(Type, metaclass=ArrayMeta):
         if self.iswhole(ts):
             return ts[0].name.array
 
-        return type(self)(*ts)
+        return type(self).flip()(*ts)
 
     def const(self):
         for t in self.ts:
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -15,7 +15,7 @@ from .ref import AnonRef
 from .bit import Bit, VCC, GND
 from .array import Array, ArrayMeta
 from .debug import debug_wire
-from .t import Type
+from .t import Type, Direction
 
 
 def _coerce(T: tp.Type['Bits'], val: tp.Any) -> 'Bits':
@@ -39,6 +39,15 @@ def bits_cast(fn: tp.Callable[['Bits', 'Bits'], tp.Any]) -> \
 class BitsMeta(AbstractBitVectorMeta, ArrayMeta):
     def __new__(mcs, name, bases, namespace, info=(None, None, None), **kwargs):
         return ArrayMeta.__new__(mcs, name, bases, namespace, info, **kwargs)
+
+    def __call__(cls, *args, **kwargs):
+        result = super().__call__(*args, **kwargs)
+        if all(x.is_output() for x in result.ts) and not cls.is_output():
+            if cls.is_input() or cls.is_inout():
+                raise TypeError("Can't construct output with input/inout")
+            # Make it an output
+            return cls.qualify(Direction.Out)(*args, **kwargs)
+        return result
 
     def __getitem__(cls, index):
         if isinstance(index, int):
--- a/tests/test_issues/test_562.py
+++ b/tests/test_issues/test_562.py
@@ -1,0 +1,19 @@
+import magma as m
+
+import sys
+import os
+
+TEST_SYNTAX_PATH = os.path.join(os.path.dirname(__file__), '../test_syntax')
+
+sys.path.append(TEST_SYNTAX_PATH)
+
+from test_sequential import DefineRegister, phi, compile_and_check
+
+
+def test_562():
+    BV1 = m.Bits[1]
+    @m.circuit.sequential
+    class A:
+        def __call__(self, a: m.Bit, b: m.Bits[2]) -> m.Bits[1]:
+            return a.ite(BV1(a), b[0:1])
+    compile_and_check("test_562", A, "coreir-verilog")
--- a/tests/test_syntax/gold/test_562.v
+++ b/tests/test_syntax/gold/test_562.v
@@ -1,0 +1,16 @@
+module coreir_mux #(parameter width = 1) (input [width-1:0] in0, input [width-1:0] in1, input sel, output [width-1:0] out);
+  assign out = sel ? in1 : in0;
+endmodule
+
+module A_comb (output [0:0] O, input a, input [1:0] b);
+wire [0:0] magma_Bit_ite_Out_Bits_1_inst0_out;
+coreir_mux #(.width(1)) magma_Bit_ite_Out_Bits_1_inst0(.in0(b[0]), .in1(a), .out(magma_Bit_ite_Out_Bits_1_inst0_out), .sel(a));
+assign O = magma_Bit_ite_Out_Bits_1_inst0_out;
+endmodule
+
+module A (input ASYNCRESET, input CLK, output [0:0] O, input a, input [1:0] b);
+wire [0:0] A_comb_inst0_O;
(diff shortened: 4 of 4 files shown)


rdaly525:
From a user perspective, I would definitely prefer not needing to write `m.Out(m.Bits[1])`. Is there any use case (or would it make even make sense) to coerce an instance of a directed type to be undirected? If not, maintaining direction seems fine.

magma/array.py
```diff
@@ -342,7 +342,7 @@ def value(self):
         if self.iswhole(ts):
             return ts[0].name.array
 
-        return type(self)(*ts)
+        return type(self).flip()(*ts)
```
rsetaluri: is this a separate bug?

magma/bits.py
```diff
@@ -40,6 +40,15 @@ class BitsMeta(AbstractBitVectorMeta, ArrayMeta):
     def __new__(mcs, name, bases, namespace, info=(None, None, None), **kwargs):
         return ArrayMeta.__new__(mcs, name, bases, namespace, info, **kwargs)
 
+    def __call__(cls, *args, **kwargs):
+        result = super().__call__(*args, **kwargs)
+        if all(x.is_output() for x in result.ts) and not cls.is_output():
+            if cls.is_input() or cls.is_inout():
```
rsetaluri: how is this possible?

rdaly525:
@rsetaluri, if we are that strict, then circuit.combinational and circuit.sequential will always have to reference directionality when casting to different types. This breaks the abstract interpretation goal of PEak since directionality does not exist/make sense for BitVector and SMTBitVector.

rsetaluri:
I understand -- it seems like then it should be the job of the `combinational/sequential` functions to figure that stuff out and do the right thing for directionality (rather than change the base behavior of magma).

Again, if it's for utility, I don't need to be a stickler about this, but think that's more "correct".

magma/array.py
leonardt: This bug was revealed when enforcing this directionality changed.

Basically, when you have an input, you can retrieve what was driving it by calling `.value`.  However, the driver is an output, so you should be creating the flipped type.  It turns out the code using these values was not affected by the directionality (it's used by the backend to find out what is driving a value), but this is technically what should be happening.

magma/bits.py
leonardt: This would happen if the user said `m.In(Bits[5])(value)` where value contained outputs.

leonardt:
I think this might have to do with the conflation between `BitVector(<value>)` as a way to coerce a value to a type, whereas in magma we use `m.bits(<value>)` as a way to convert value to a type.  This conversion is where it normally handles preserving the direction of the converted type.  

But really, I think the only time we see type conversions are on outputs, e.g. it wouldn't really make sense to say `m.bits(io.O, 5)`  where `io.O` is an input, so that logic really just assumes everything is an output.

There's also the semantics right now with `Bit(0)` and `Bit(1)` constants which will return `VCC` and `GND` which are actually `m.Out(m.Bit)`, so in terms of consistency there's already behavior that returns a directed type from an undirected type.

I do think it is a bit confusing and I'm not sure what the best way to handle this is.  I think enabling this behavior might be okay for now since, as I said, it's consistent with Bit constructor semantics.  However, we may want to find a way to specify a concrete rule for this type of behavior.  

E.g. if you construct a value of an undirected type with an output value (constant or not), it will return an output value.

rdaly525:
I do not know if this is exactly the same issue, but I could imagine wanting to use an "input" as an output. Something like this verilog code:

```
module A(
  input [15:0] a,
  input [15:0] b,
  output [15:0] sum,
  output [15:0] sum_plus1
);
  assign sum = a+b;
  assign sum_plus1 = sum + 1;
endmodule
```

Or in magma:
```
    Data = m.SInt[16]
    class A(m.Circuit):
        IO = ['a', m.In(Data), 'b', m.In(Data), 's', m.Out(Data), 'sp1', m.Out(Data)]

        @classmethod
        def definition(io):
            io.s <= (io.a + io.b)
            io.sp1 <= (io.s+1)
```
^this does complain that io.s is an input and being used as an output.

leonardt:
I don't think that's a pattern we support at the moment.  I think the idiom would be use a temporary instead.  One reason to enforce this strictness would be to catch cases where a user accidentally uses an input unintentionally.  

One weird idea would be to turn an input into an output once it's been driven, so it can be reused elsewhere (or some variant of this concept).  This could allow the reuse in Ross's example without sacrificing the strictness of input/output rules.

rdaly525:
> One weird idea would be to turn an input into an output once it's been driven, so it can be reused elsewhere (or some variant of this concept). This could allow the reuse in Ross's example without sacrificing the strictness of input/output rules.

It would allow some cases but not all, since using a signal is always allowed "before" you assign to it.
```
module A(
  input [15:0] a,
  input [15:0] b,
  output [15:0] sum,
  output [15:0] sum_plus1
);
  assign sum_plus1 = sum + 1; //This can come first
  assign sum = a+b;
endmodule

```

leonardt:
Yes, but I don't know if we want to support all cases (i.e. what are our desired semantics for a good design, rather than recreating what verilog does)

leonardt:
Magma's wiring rules are stricter than verilog:
* an output can be wired to an inputs
* an output can be wired to multiple inputs
* an input must be wired to a single output

There is a small convenience in adding a verilog style rule for wiring an input to an input, but I worry that it introduces a class of errors that are avoided without the rule, and the convenience is not worth that possibility of error

rdaly525:
I think it is a philosophical question of whether or not you should be allowed to use any internal signal of a module (including IO) for use in the definition. I personally like that mentality since for a physical circuit you are always allowed to peek any signal at any location. From a typing/safety perspective I understand that it could prevent some bugs and might also be annoying to support.

leonardt:
I'm going to merge this to unblock the current issue, I'll open a separate issue to track this discussion of type constructors and qualifiers.

PR #565, opened 2020-01-31
Refactor from verilog

rsetaluri:


magma/frontend/verilog.py
```diff
@@ -1,89 +1,138 @@
...
 __all__ += ['DeclareFromTemplatedVerilog']
 __all__ += ['DeclareFromTemplatedVerilogFile']
 __all__ += ['DefineFromVerilog']
 __all__ += ['DefineFromVerilogFile']
 __all__ += ['DefineFromTemplatedVerilog']
 __all__ += ['DefineFromTemplatedVerilogFile']
+__all__ += ['declare_from_verilog']
+__all__ += ['declare_from_verilog_file']
+__all__ += ['define_from_verilog']
+__all__ += ['define_from_verilog_file']
```
leonardt: While we're at it, should we transition to having `__init__` import the desired functions, rather than using `__all__` and `import *`?

tests/test_verilog/test_from_file.py
```diff
@@ -118,99 +121,98 @@ def definition(io):
...
+# module foo(input I, output O);
+#     assign O = I;
+# endmodule
+
+# module bar(input I, output O);
+#     foo foo_inst(I, O);
+# endmodule""", external_modules={"foo": foo})
+#         assert False
+#     assert pytest_e.type is MultipleModuleDeclarationError
+#     assert pytest_e.value.args == ("foo",)
```
leonardt: Should this be added back in?

rsetaluri: Yes, planning to refactor the tests a bit too.

magma/frontend/verilog.py
rsetaluri: Agreed, will do.

tests/test_verilog/test_from_file.py
rsetaluri: Actually, I misspoke. This change gets rid of the functionality these tests were originally written for. This changes gets rid of anything to do with `shallow` or `external_modules`.

I think we want to move towards importing verilog only for the sake of getting the interface; we shouldn't be passing the verilog down to CoreIR. It should be linked in otherwise.

leonardt:
@rsetaluri it seems like shallow has been removed and is by default now, just wanted to check if now we're expecting the user manage this issue? (e.g. importing a module that depends on other modules in the same file)

rsetaluri:
@leonardt yes, I wanted to get rid of these smarts and just push the issue up to maybe the CoreIR linking level, or force the user to manually do it. Having this be opaque didn't seem like a good design. I actually think we should completely move away from "defining" from verilog, i.e. only use the parser to get an interface so we can know how to instance and wire to a module, rather than pass it's source to the compiler.

Issue #553, opened 2020-01-29
_magma_ protocol


cdonovick:
Its is not possible to encapsulate types and still allow wiring.
Consider:
```Python
class Foo:
    def __init__(self, val: Optional[Bits[8]] = None):
        if val is None: val = Bits[8]()
        self._val = val

    def non_standard_operation(self):
        return Foo((self._val << 2) | (self._val[0] << 1) | self._val[0])
```
While a `Foo` holds a `Bits[8]`  it is not possible to wire it to a `Bits[8]`

In a previous discussion we decided that we would add some sort of protocol to allow types to cast themselves to a magma base type before wire would try to dispatch on them.

leonardt:
We've added basic support this, we can open new issues to track more specific things.

PR #566, opened 2020-01-31
Prototype MagmaProtocol

leonardt:
Prototype for https://github.com/phanrahan/magma/issues/553

I don't think this is complete, but it's a minimum working example we can use to comment on.

The prototype I came up with involves two constructs: `MagmaProtocol` and `MagmaProtocolMeta`, implementing the protocol requires four concepts (which reveals why we need the meta class protocol).
1) We need to be able to convert a type to the corresponding magma type, done using the `_to_magma_()` method
2) We need to be able to create a type from a corresponding magma type, done using the `_from_magma_()` method.  The ability to go to and from magma is required for type qualification, since this involves creating new type (i.e. give me your magma type, I'll qualify it, now create a new version of your type with the qualified version of the magma type).
3) We need to be able to access the underlying magma value from an instance of a type, done using `_get_magma_value_()`.  This is used by the various `MagmaProtocol` methods that dispatch to the underlying value (e.g. for wiring).
4) ~~We need to be able to instantiate a "named version" of the type.  This is the part I'm a little unsure of how to best handle.  Basically, when you have a magma port (e.g. `m.Bits[5]`), it will create an instance of the type with a name referring to the port (e.g. `m.Bits[5](name="I0")`).  To support using these types in the interface of a `Circuit`, you'll need to support this constructor pattern (just providing a name keyword arg).  One option would be to hide this logic inside the meta class logic to dispatch to a special method like `T._magma_make_port(name)` or something.~~ Can use 1/2 to achieve this.

The current implementation of MagmaProtocol was sufficient to compile the `Foo` example through coreir, so I believe this should would for MagmaAADT and will likely simplify the implementation.


cdonovick:
Is 4 not accomplished by 1?

leonardt:
Good point, I'll rework it to use that logic

leonardt:
@cdonovick to better enforce the interface of requiring a constructor with a magma value (necessary for implementing 4 with 1/2), I moved the logic into the base class (MagmaProptocol), so any user defined class needs to call `MagmaProtocol.__init__(val)` where `val: m.Type`.  With this, it's then possible to define `_get_magma_value()` in the base class since the base class stores the value internally.

Are you okay with the management of the underlying magma value being handled by the base class? Or do you need custom behavior for this.

cdonovick:
Hmmm,  I think that should be fine. It is not really clear to me why it's necessary but I have no problem with it. I would advocate for using `_val_`  rather than `_val` so it is clear that `_val_`  is a special name not not a implementation detail (e.g. `__dict` vs `__dict__`).

leonardt:
In the interface code, we need to get the underlying magma type, then instantiate a version of they type containing the magma value with a reference to the port, so the code looks like
```
                args[name] = port(port._to_magma_()(name=ref))
```

Where `port` is an instance of MagmaProtocol.  So, it's required that the `port` constructor, accept a magma value as the first argument.  To enforce this, we require the custom type to dispatch to the superclass `__init__` method with this value, and since we're doing this, we now have control over how the value is stored.

The other option is to do something like, `port._from_magma_value_(port._to_magma_()(name=ref))` which would construct the port from the magma value.  So, the custom type can have any implementation of `__init__` that they'd like, so long as they provide a way to construct an instance from a provided magma value using the `_from_magma_value` function.

I think either would work, do you have a preference for which interface to use?

magma/backend/coreir_transformer.py
```diff
@@ -17,6 +17,7 @@
 from ..is_definition import isdefinition
 from ..logging import root_logger
 from ..passes import InstanceGraphPass
+from ..t import MagmaProtocol
```
rsetaluri: this looks stray?

leonardt:
Yea, a related thought is that the built-in types should implement the Protocol, so the backend compiler should just assume that everything is an instance of Protocol, which i think would simplify things (in doing this, I think we can add the feature's you'd like, which would essentially allow users to extend the language with new base types, assuming they added the required methods to convert to a coreir type).

cdonovick:
@leonardt I am fine with either `_from_magma_value_` or the expectation that init can handle a magma value.  I think I have a slight preference for `_from_magma_value_` just seems the most consistent. 

@rsetaluri @leonardt If I was implementing this you better believe I would make all the base types implement the protocol.   Maybe `circuit.interface` could be a sort of alias for `_to_magma_`?  I don't know all the details of what's going on there.

PR #568, opened 2020-02-02
[Proposal] AnonProduct

splhack:
I really miss old `Tuple` for declaring return type.

    def test(I: m.Bits[2]) -> m.Tuple(x=m.Bit, y=m.Bit):

Current options are, 1. use `from_fields` (too long to write)

    def test(I: m.Bits[2]) -> m.Product.from_fields("anon", dict(x=m.Bit, y=m.Bit)):

Or, 2. declare `Product` type beforehand (not convenient)

    class Out(m.Product)
        x = m.Bit
        y = m.Bit

    def test(I: m.Bits[2]) -> Out:

Can we introduce `AnonProduct`? just alias for `from_fields("anon", dict(`
since magma already has `AnonRef`

    def test(I: m.Bits[2]) -> m.AnonProduct(x=m.Bit, y=m.Bit):

### DIFF
--- a/magma/tuple.py
+++ b/magma/tuple.py
@@ -336,6 +336,27 @@ class Tuple(Type, Tuple_, metaclass=TupleKind):
             mixed |= field.is_mixed()
         return mixed or (input + output + inout) > 1
 
+
+def _add_properties(ns, fields):
+    def _make_prop(field_type, idx):
+        @TypedProperty(field_type)
+        def prop(self):
+            return self[idx]
+
+        @prop.setter
+        def prop(self, value):
+            self[idx] = value
+
+        return prop
+
+    for idx, (field_name, field_type) in enumerate(fields.items()):
+        if field_name == "N":
+            # TODO: Make N unreserved
+            raise ValueError("N is a reserved name in Product")
+        assert field_name not in ns
+        ns[field_name] = _make_prop(field_type, idx)
+
+
 class ProductKind(ProductMeta, TupleKind, Kind):
     __hash__ = type.__hash__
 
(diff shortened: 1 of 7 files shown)


leonardt:
This seems reasonable to me, the downside is we lose compatibility with Peak, but this should be okay if it's only meant to be used within the context with magma (@cdonovick any thoughts?)

cdonovick:
I have no problem with adding an anonymous product to hwtypes. I would just argue for somewhat inelegant but consistent type constructor: `AnonProduct[dict(x=Bit, y=Bit)]`

leonardt:
That seems reasonable, @splhack would you be okay with using the bracket syntax `[...]` for the type constructor so it's consistent with the other types? (e.g. Array[N, T] and Bits[N])

splhack:
yeah, it'd be still much better than `from_fields`🙂
It'd be nice if python allows to remove `dict()` for the square brackets, but it is impossible, right?

    def test(I: m.Bits[2]) -> m.Tuple(x=m.Bit, y=m.Bit):
    def test(I: m.Bits[2]) -> m.AnonProduct(x=m.Bit, y=m.Bit):
    def test(I: m.Bits[2]) -> m.AnonProduct[dict(x=m.Bit, y=m.Bit)]:
    def test(I: m.Bits[2]) -> m.Product.from_fields("anon", dict(x=m.Bit, y=m.Bit)):

leonardt:
Unfortunately yes, this is a limitation of the Python syntax (keywords args are not supported in __getitem__)

splhack:
updated for square bracket constructor

    def test(I: m.Bits[2]) -> m.AnonProduct[dict(x=m.Bit, y=m.Bit)]:

magma/tuple.py
```diff
@@ -430,6 +430,19 @@ def value(self):
         return namedtuple(**dict(zip(self.keys(),ts)))
 
 
+class AnonProductMeta(ProductMeta):
```
cdonovick: This type relationship is backward.

cdonovick:
related https://github.com/leonardt/hwtypes/pull/123

magma/tuple.py
leonardt: @splhack can you update this to match the new `hwtypes` hierarchy? See https://github.com/leonardt/hwtypes/pull/123 (basically, anon product is a parent of product)

rsetaluri:
@leonardt @cdonovick @splhack what's the state of this?

leonardt:
The type hierarchy needs to be fixed to match hwtypes (Product should be a subclass of AnonProduct, not the other way around (see my comment above))

splhack:
(finally) synced with leonardt/hwtypes#123 🙂

Issue #563, opened 2020-01-31
Enum namespace


leonardt:
We should either:
(a) have a list of reserved names and error when a user provides an enum field that uses one (see https://github.com/phanrahan/magma/issues/500)
(b) use a separate namespace for enum fields that avoids clobbering the internally used names by the type system, so the user can use arbitrary names without breaking internal code.

splhack:
Product needs same thing #567

cdonovick:
hwtypes uses a list of reserved `names` which may not be fields.
There is also list of reserved `_names_` which if set will have strange side effects but allow for extension.

https://github.com/leonardt/hwtypes/blob/master/hwtypes/adt_meta.py#L40
https://github.com/leonardt/hwtypes/blob/master/tests/test_adt.py#L359

It would make sense for magma to do something

PR #572, opened 2020-02-05
replace method for Array/Tuple/Product

### LINKED ISSUE #559: Can sub-portion of Product be set?
Probably I should use `m.namedtuple` to set `Product` but let me give it a shot.

Isn't `a.a0 = value` the same thing as `a = m.namedtuple(a0=value, a1=a.a1)`?

    import magma as m
    import fault
    
    
    class A(m.Product):
        a0 = m.Bits[8]
        a1 = m.Bits[8]
    
    
    @m.circuit.sequential(async_reset=False)
    class TestTupleUpdate:
        def __init__(self):
            self.a: A = m.namedtuple(a0=0, a1=0)
    
        def __call__(self, sel: m.Bit, value: m.Bits[8]) -> A:
            a = self.a

            # TypeError: 'odict_values' object is not subscriptable    
            # if sel:
            #    a.a0 = value
            # else:
            #    a.a1 = value
    
            if sel:
                a = m.namedtuple(a0=value, a1=a.a1)
            else:
                a = m.namedtuple(a0=a.a0, a1=value)
    
            self.a = a
            return a
    
    
    tester = fault.Tester(TestTupleUpdate, clock=TestTupleUpdate.CLK)
    
    tester.circuit.sel = 0
    tester.circuit.value = 11
    tester.step(2)
    
    tester.circuit.O.a0.expect(0)
    tester.circuit.O.a1.expect(11)
    
    tester.circuit.sel = 1
    tester.circuit.value = 12
    tester.step(2)
    
    tester.circuit.O.a0.expect(12)
    tester.circuit.O.a1.expect(11)
    
    tester.compile_and_run(
        "verilator",
        flags=["-Wno-fatal"],
        magma_opts={"verilator_debug": True},
        directory="build",
    )

splhack:
close #559 

    val0 = m.namedtuple(x=m.bit(0), y=m.bit(1))
    val1 = val0._replace(x=m.bit(1))


rsetaluri:
@splhack thanks for the change! I would prefer the name `update` rather than `replace` (similar to python `dict` naming).

@cdonovick @leonardt

splhack:
@rsetaluri the thing is, dict `update` is in-place update, namedtuple `_replace` is creating a new instance, which is the same behavior of this. take a look at #559

cdonovick:
I am not in loved with the name `_replace` we do not use this name style for other methods in product / tuple.  The reasoning for python `namedtuple` to do this is avoid reserving field names.  Which is reasonable, however we should be consistent in style.

@rsetaluri I agree with @splhack regarding `update`.

splhack:
maybe depends on #563 ? (a) `replace` (b) `_replace`

leonardt:
I don't feel strongly either way (consistency with namedtuple is nice, but consistency with existing interfaces is also nice).  If starting from scratch, I might suggest that all methods/attributes of products/enums be hidden using the `_` prefix to avoid the namespace issue, but that will involve a lot of retroactive code changes at this point.

cdonovick:
Agree with the point about starting over but I feel like that ship has mostly sailed.  I personally don't care about consistency with python namedtuple though (or at least it is significantly less important to me than internal consistency)

splhack:
ok, will update it for `replace`.
we need to update https://github.com/phanrahan/magma/blob/master/magma/tuple.py#L342-L344 to check field_names against methods and properties (I got some complex error when I used `value` field in Product.)

cdonovick:
Another option would be to make this a function `replace(product, **kwargs)` so we wouldn't have to reserve a name.  I don't feel strongly though.

splhack:
val0 = m.namedtuple(x=m.bit(0), y=m.bit(1))
    val1 = val0.replace(x=m.bit(1))

leonardt:
Hmm, using a functional syntax might be good because it makes it more clear that a new copy is returned (we could also augment the function later to support replacing array members)

splhack:
so, the function would look like this.

    val0 = m.namedtuple(x=m.bit(0), y=m.bit(1))
    val1 = m.replace(val0, x=m.bit(1))

how it works for array? maybe need to use tuple?

    array1 = m.replace(array0, (1, m.bit(1))

If that's the case, we should use dict for Product?

    val1 = m.replace(val0, dict(x=m.bit(1)))

cdonovick:
the `dict` syntax might be nice as it could be uniform between array / tuple / product
```
replace(array, {idx: new_val})
replace(tuple, {idx: new_val})
replace(product, {field_name: new_val})
```

phanrahan:
I like using a single function `replace`!

splhack:
Array/Tuple

    val1 = m.replace(val0, dict([(0, m.bit(1))]))
    val1 = m.replace(val0, {"0": m.bit(1)})

Product

    val1 = m.replace(val0, dict(x=m.bit(1)))
    val1 = m.replace(val0, {"x": m.bit(1)})

PR #576, opened 2020-02-07
Improve circuit

rsetaluri:


magma/interface.py
```diff
@@ -234,3 +262,34 @@ def DeclareInterface(*decl, **kwargs):
     name = f"Interface({', '.join([str(a) for a in decl])})"
     dct = dict(Decl=decl, **kwargs)
     return InterfaceKind(name, (_DeclareInterface,), dct)
+
+
+def DeclareLazyInterface(io, **kwargs):
+    """LazyInterface factory function"""
+    decl = io.decl()
+    name = f"Interface({', '.join([str(a) for a in decl])})"
```
leonardt: Repeated from line 262, should we make a helper function `make_name`?

magma/interface.py
```diff
@@ -234,3 +262,34 @@ def DeclareInterface(*decl, **kwargs):
...
+    """
+    # Note that because we require kwargs to be ordered, we have a strong
+    # requirement here for >= python version 3.6. See
+    # https://www.python.org/dev/peps/pep-0468/.
+    def __init__(self, **kwargs):
+        self.ports = {}
+        self.__decl = []
+        for name, typ in kwargs.items():
+            ref = LazyDefnRef(name=name)
+            port = typ.flip()(name=ref)
```
leonardt: We could insert a type check here on `name, type` to be `str, subclass(Type)` to catch errors early

magma/interface.py
```diff
@@ -174,23 +196,29 @@ class _DeclareInterface(_Interface):
...
+        # (non-lazy) interface construction logic.
+        if not defn:
+            self.ports = _make_interface_args(self.Decl, renamed_ports, inst,
+                                              defn)
+            return
+        args = OrderedDict()
+        for name, port in self.io.ports.items():
+            ref = port.name
+            if inst:
+                ref.set_inst(inst)
```
leonardt: I don't see `set_inst` defined in LazyDefnRef, is this a case we need to consider?

magma/interface.py
rsetaluri: Agreed, though after several conversations w/ @cdonovick, he's convinced me that kind of error checking is not pythonic 😄but might be misinterpreting his point

rsetaluri:
@leonardt yes definitely need to add documentation and tests. Still WIP.

magma/interface.py
cdonovick: @rsetaluri its not pythonic to type check because duck typing.

cdonovick: Now sometimes it's type checking will give you significantly better error messages.  But in those cases I would promote checking an abstract base class not the existing implementation.

rsetaluri:
Thought about some issues relevant to this change:

# Testing
How can we test that this change is good? For one, the fact that tests pass presumably means that at least we didn't break any existing functionality. I've also included a test (`tests/test_circuit/test_new_style_syntax.py`) that tests basic functionality and some pathological cases.

I also have a small script that converts the old-style syntax to the new-style syntax. I've run this on *all* existing tests, and now have a full suite of the original tests with the new syntax. We should definitely keep the existing tests around to make sure things work, but we want to make sure the new syntax is robust on all the expected cases. Not sure what's the best thing to do here.

# Reporting wiring errors with new-style syntax
If a wiring error was made, previously, this would be caught and reported like `Foo.O (an output) is being used as an input`. However, now we don't have a handle to the circuit name `"Foo"` at wiring time, because the circuit definition (class) itself doesn't exist yet! (See commit msg of 274a767b3afbc00fad16c8245fa027cdbc9d6582.) The solution to this right now is to report it as `.O (an output) is being used as an input`, which isn't too terrible (especially if line numbers are emitted in debug mode). There are some clever things we can do to try to figure out the name; otherwise we'd have to stage the wiring errors and emit them once the class is created. I'm not convinced this complexity is warranted right now though.

# Semantics of `isdefinition()`
The semantics of `isdefinition()` is kind of muddy/opaque as is, and the introduction of the new syntax makes it more so. (See commit msg of 1ab4e8fb0b278c1f48c542e385e3feb99617c135 for a detailed explanation.) Since this *mostly* an internal convenience function, and nothing is clearly broken as of yet, I'm not convinced it's worth making this 100% robust right now.

@leonardt let me know if you have particular feedback on the above. I think this is good to merge modulo those concerns.

rsetaluri:
@leonardt: merging; planning to look at these issues once this is in.

PR #578, opened 2020-02-07
Add ability to do a dynamic slice with a constant width

leonardt:
(CC @priyanka-raina and @kong0329)

Adds the ability to do a constant width slice, but using a dynamic offset value.

We currently don't support a truly dynamic slice such as `x[y:z]` where `y` and `z` are magma values, since this would require the notion of a value with a variable width (depending on the dynamic value of `y` and `z`).  However, this adds support for the simpler case of `x[y:y+OFFSET]` where `OFFSET` is a constant value.  This essential selects a constant slice of values, with a dynamic start offset (`x`).

The initial implementation uses a functional style operator `m.slice(value: Bits, start: Bits, width: int` since this would be easier to implement then the general slice syntax (e.g. `value[start:start+width]` we'd have to determine that the top of the slice was a constant offset, but that could be tricky in general, for example if you had `(1 << N)` instead of `width`.  Using a functional style operator requires the user to perform any computation and realize it as a constant value before passing it to the operator.

### DIFF
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -84,3 +84,5 @@ def get_family():
 
 from .generator import Generator
 from .monitor import MonitorIO, MonitorGenerator
+import magma.operators
+from .operators import slice
--- a/magma/array.py
+++ b/magma/array.py
@@ -210,7 +210,13 @@ class Array(Type, metaclass=ArrayMeta):
     def flat_length(cls):
         return cls.N * cls.T.flat_length()
 
+    def dynamic_mux_select(self, key):
+        return m.operators.Mux(len(self), self.T)(*self.ts, key)
+
     def __getitem__(self, key):
+        if isinstance(key, Type):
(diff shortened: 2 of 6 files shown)


leonardt:
Here's the some code:
```python
    class TestSlice(m.Circuit):
         # Slice a 6 bit window of I using x as the start index
         IO = [
             "I", m.In(m.Bits[10]), 
             "x", m.In(m.Bits[2]), 
             "O", m.Out(m.Bits[6])
         ]

         @classmethod
         def definition(io):
             io.O @= m.slice(io.I, start=io.x, width=6)
```

cdonovick:
While this may not be the way you want to implement this in magma this is doable in smt compliant way with:
```Python
def extract_offset(bv: BitVector[n], offset: BitVector[n], width: int):
    return (bv >> offset)[:width]
```

PR #581, opened 2020-02-10
Use _qualify_magma_

leonardt:
As requested by @cdonovick change the protocol to use `_qualify_magma_` to specifically handle type qualifiers, rather than supporting a general `_from_magma_` function.

### DIFF
--- a/magma/t.py
+++ b/magma/t.py
@@ -1,7 +1,7 @@
 import functools
 import warnings
 import enum
-from abc import abstractmethod, ABCMeta
+from abc import abstractmethod
 from .common import deprecated
 from .ref import Ref, AnonRef, DefnRef, InstRef
 from .compatibility import IntegerTypes, StringTypes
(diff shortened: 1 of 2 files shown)


magma/t.py
```diff
@@ -142,54 +142,50 @@ def flip(cls):
...
-    def wrapper(T):
-        if issubclass(T, MagmaProtocol):
-            return T._from_magma_(func(T._to_magma_()))
-        return func(T)
-    return wrapper
-
-
-@dispatch_magma_protocol
 def In(T):
     return T.qualify(Direction.In)
```
cdonovick: Don't we still need to dispatch to the protocol? E.g.  `In(MagmaProtocol_t)` needs to dispatch to  `MagmaProtocol_t._qualify_magma_(In)`

cdonovick: Oh wait that's in `MagmaProtocol.qualify`

PR #582, opened 2020-02-10
Add support for qualifying abstract Array/Bits (e.g. In(Bits))

leonardt:


tests/test_type/test_array.py
```diff
@@ -41,6 +41,24 @@ def test_array():
...
 
 
+def test_qualify_array():
+    assert str(m.In(Array)) == "In(Array)"
+    assert str(m.Out(m.In(Array))) == "Out(Array)"
+    assert str(m.In(Array)[5, m.Bit]) == "Array[5, In(Bit)]"
+    assert str(m.Out(m.In(Array))[5, m.Bit]) == "Array[5, Out(Bit)]"
+
+    # Array qualifer overrides child qualifer
+    assert str(m.In(Array)[5, m.Out(m.Bit)]) == "Array[5, In(Bit)]"
```
rsetaluri: if we have these weird nested direction qualifiers, I feel like we should log/warn what rules we're applying. (Don't disagree with the rules, just want them to be transparent to the user who maybe isn't aware of this complexity.)

magma/array.py
```diff
@@ -50,15 +50,49 @@ def __new__(cls, name, bases, namespace, info=(None, None, None), **kwargs):
...
+            # the default direction logic
+            if not isinstance(index[1], Direction):
+                # Otherwise, we expect that we're qualifying a Type with the
+                # direction (e.g. In(Array)[5, Bit])
+                if not issubclass(index[1], Type):
+                    raise TypeError("Expected Type as second index to Array")
+                if not index[1].is_oriented(cls.T):
+                    _logger.warning(
+                        f"Parametrizing qualifed Array {cls} with inner type "
+                        f" {index[1]} which does match, will use array "
```
rsetaluri: nit: doesn't match?

Issue #575, opened 2020-02-07
[RFC] Naming Temporary Values


leonardt:
There's been requests for support for passing named temporary values down through the toolchain.  For now, we could add support for named values such as `x = Bit(name="x")` and find a way to pass that name through the toolchain so that coreir generates a `wire x`.

However, there's also been requests for automatically inserting this name (so supporting `x = Bit()` and automatically capturing the name `x`).  One simple option is to use `inspect` inside the constructor to find the source file/line and retrieve the LHS of the assignment statement.  However, this is problematic from a performance perspective, seeing as it will require a stack frame traversal for every intermediate value that's constructed.  There's further complications when you consider python code such as `x, y = Bit(), Bit()` (which variable should we use?).

Perhaps a more performant and robust approach would be to traverse the AST and annotate names a preprocessing step.  That would cover the tuple assignment case (because we can match on each side and rewrite it to be `x, y = Bit(name="x"), Bit(name="y")`.

But when I think about this more generally, there's challenges handling the arbitrary python code that magma users can write.  For example, consider `x = f(...)` where `f` returns a `Bit()`.  Ideally, we'd be able to capture the name of the Bit as `x` (this also goes for the result of an expression such as `x = y + z`).

One way to perhaps capture this pattern is to add a wrapper function for assignments, so `x = f(...)` would become `x = m._set_name("x", f(...))`.  Then, magma could inspect the result of `f`, if it's a magma value, set the name to be `x`.  

Any thoughts? I'm particularly interested in whether there's any gotchas here, or more clever/robust ways to handle this.  (I don't think calling `inspect` will be a scalable solution).

We could also use the pre-processing the AST step to capture the filename/lineno (avoiding the need for inspect in debug_mode).  Normally rewriting the AST is tricky because it can obscure error messages, but in this case, we can do this by just wrapping assignment values with a function call, so ideally we are not changing any filename/linenumbers (there's some trickiness here with deserializing the transformed AST, because we could lose the original linebreaks/comments which actually could mess up the mapping to the original source, but maybe this is not an issue if we can capture the original file/line which we'll need for the debug info anyways).

cdonovick:
;I'd be surprised it there exists some solution that doesn't incur some massive performance penalty.
I'm also not entirely clear on desired semantics.  For example how should the following respond:
```Python
def returns_3_bits():
  return Bit(), Bit(), Bit()

x, *others = returns_3_bits()

lst = [None]
lst[0] = Bit()

[...]

PR #583, opened 2020-02-11
Refactor Port logic to use Wire representation

leonardt:
In preparation for https://github.com/phanrahan/magma/issues/575

Before, magma represented circuits as direct source -> sink connections
with no intermediates.  When a temporary value was wired to another
value, their connections were merged and temporary effectively removed.

In preparation for the logic that will preserve wire names in the
generated code, these changes refactor the internals to preserve
the intermediate connections.  This is done by changing the concept
of a Port into a Wire that maintains a notion of a `driver` and a
list of Wires that it is `driving`.  Wires are qualified by the
bit that they correspond to and use the same logic.

The function of `trace` and `value` were adjust slightly for this, which
required updating the tests but still works as expected for the
internals.  Before, `trace` and `value` operated in certain ways for
inputs and outputs, however the notion of tracing an output is not
used internally anywhere except in the test.  To make the logic more
consistent/precise, `trace` will return the source input/inout driver
of a wire, while `value` will return the directly connected driver.

There's a slight confusion with the name `Wire`, since to me that could
correspond to a bidirectional entity, but here they have a notion
of being an input, output, inout, or neither.  I'm not sure if there's
a better name.  Port worked for this because we never realized the 
intermediate values, so we always connected one port to another 
(and these were qualified).  Now we have intermediate Wire objects
that have no direction, which end in directed Wires.  Perhaps
the best thing to resolve this would be to introduce a Port object
that is only used by interfaces, and wire objects for temporary values.


leonardt:
In doing this, I also think I better understand the issue with intermediate values being undirected.

This code relies on the fact that wire is a directed function `m.wire(driver, drivee)`.  This reveals a subtle issue with temporary values, for example
```
a = m.Bit()
b = m.Bit()
m.wire(a, b)
```
we have two connected, undirected values, where `a` is driving `b`.  If we do something like
```
m.wire(m.GND, a)
```
this should work as expected.

However, because `b` is undirected, one might think you can write
```
c = m.Bit()
m.wire(c, b)
```
In this case, `c` will become the new driver for `b`, replacing `a`, even though another interpretation would treat `a`, `b`, and `c` now all connected. One way to fix this might be to treat `b` as an output now that it is driven, and wire it to be the driver of `c`.

magma/backend/coreir_transformer.py
```diff
@@ -192,7 +192,9 @@ def connect_non_outputs(self, module_defn, port, non_input_ports):
             for elem in port:
                 self.connect_non_outputs(module_defn, elem, non_input_ports)
         elif not port.is_output():
-            self.connect(module_defn, port, port.value(), non_input_ports)
+            if port.trace() is None:
```
rsetaluri: stray `print`? if we want to have this information we should use `logger.info()`

leonardt:
This should be entirely internal (the only logic using the `Port` construct was the backend compilers, which use it to trace the driver of inputs to emit the connections).

Performance wise, tracing will take longer now since the old logic effectively computed the source at wiring time, but wiring should be faster now (since wires don't need to be merged).  We could speed this up by updating the trace whenever a connection occurs, but let's test if there's any performance regressions introduced before adding the complexity.  I think it boils down to spending the time during wiring versus tracing (memoization could help, although I'm not sure we ever trace more than once).  I think optimizing other aspects of the wiring would be more beneficial than these kinds of optimizations.  For example, we can attach `Wire` objects to Arrays and Tuples and connect those, rather than the leaf Bits, and only realize the children when necessary.

PR #584, opened 2020-02-11
Insert CoreIR Wire Module Pass

leonardt:
This adds a pass that replaces all wires with an instance of CoreIR's Wire primitive.  This will be used by the CoreIR backend to insert wire primitives that are then used to code generate temporary values.

This updates the `repr` logic to properly handle the introduction of the notion of temporary values, and to better reconstruct the original circuit (used in the testing of this logic).

Depends on https://github.com/phanrahan/magma/pull/583


magma/array.py
```diff
@@ -199,10 +204,10 @@ def __eq__(self, rhs):
     __hash__ = Type.__hash__
 
     def __repr__(self):
-        if not isinstance(self.name, AnonRef):
-            return repr(self.name)
-        ts = [repr(t) for t in self.ts]
-        return 'array([{}])'.format(', '.join(ts))
+        if self.name.anon():
+            ts = [repr(t) for t in self.ts]
+            return 'array([{}])'.format(', '.join(ts))
```
rsetaluri: f-string instead of `.format()`

magma/array.py
```diff
@@ -374,5 +379,32 @@ def unused(self):
         for elem in self:
             elem.unused()
 
+    def as_bits(self):
+        if isinstance(self.T, Digital):
+            return Bits[len(self)](self.ts)
+        else:
```
rsetaluri: don't need else

magma/bits.py
```diff
@@ -456,6 +456,13 @@ def undriven(self):
             raise TypeError("undriven can only be used on input")
         m.wire(DefineUndriven(len(self))().O, self)
 
+    def as_bits(self):
```
rsetaluri: should all these functions (`as_bits`, `from_bits`, etc.) be added as methods to the metaclasses?

magma/circuit.py
```diff
@@ -134,6 +136,37 @@ def __str__(cls):
             interface = f"({interface})"
         return f"{cls.__name__}{interface}"
 
+    def add_intermediate_value(cls, value, values):
```
rsetaluri: why are these cls functions? I don't see them modifying or introspecting `cls` at all

magma/circuit.py
```diff
@@ -587,6 +633,12 @@ def bind(cls, monitor, *args):
             content = "\n".join((f.read(), bind_str))
         cls.bind_modules[monitor.name] = content
 
+    def add_instance(cls, circuit, *args, **kwargs):
```
rsetaluri: This seems a little wonky -- if we want to just place the instance, we should do that directly in the pass rather than call a function on the circuit which is exposed to the world? Or this function `add_instance` should be something that can be expected to be used by outside users

magma/interface.py
```diff
@@ -52,22 +52,46 @@ class _Interface(Type):
     def __str__(self):
         return str(type(self))
 
+    def make_wire_str(self, driver, value):
```
rsetaluri: let's make these `_make_wire_str()` and `_make_wires()` unless we expect the outside world to use them

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def simulate_wire(self, value_store, state_store):
+    value_store.set_value(self.O, value_store.get_value(self.I))
+
+
+class Wire(Generator):
+    @staticmethod
+    def generate(name, T):
+        if issubclass(T, Digital):
```
rsetaluri: probably can be a ternary: `width = 1 if issubclass(T, Digital) else T.flat_length()`

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,89 @@
...
+                Wire, name, type(value), name=name
+            )
+            wire_inst.I @= driver
+            value @= wire_inst.O
+            if not driver.is_output():
+                self.insert_wire(driver, definition)
+
+    def __call__(self, definition):
+        # Copy instances because inserting wire will append to instances
+        instances_copy = definition.instances[:]
```
rsetaluri: use `.copy()`

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,89 @@
...
+                circuit.O @= T.from_bits(wire_inst.O)
+        return WrappedWire
+
+
+class InsertCoreIRWires(DefinitionPass):
+    def __init__(self, main):
+        super().__init__(main)
+        self.seen = set()
+
+    def sanitize_name(self, name):
```
rsetaluri: make these 2 functions private with `_` prepended

magma/ref.py
```diff
@@ -13,17 +13,34 @@ def __repr__(self):
...
 
     def qualifiedname(self, sep='.'):
-        return str(self.name) if self.name else f"AnonymousValue_{id(self)}"
+        return f"AnonymousValue_{id(self)}"
+
+    def anon(self):
+        return True
+
+
+class NamedRef(Ref):
```
rsetaluri: I think we should call this `AnonNamedRef` or something...or inherit `DefnRef` and `InstRef` from `NamedRef`. It sounds like a baseclass (which it could be useful for).

leonardt: AnonNamedRef doesn't really make sense to me as a name (it's a contradiction), I'll see if I can weave it into a base class

magma/circuit.py
leonardt: The issue is that then the user needs to do
```
        _DefinitionBlock.push(CurrentDefinition)
        inst = circuit(*args, **kwargs)
        _DefinitionBlock.pop()
```
Adding at as a method to the class allows us to easily abstract this logic (since the class has a reference to itself).  However, this also means we can't create the instance until the DefinitionBlock is populated.

I'd rather avoid the user having to push/pop the definition stack every time an instance is added.  I'm not sure if there's another way around this other than maybe adding a notion of a floating instance and placing it afterwards? The problem is that right now there is an implicit `place` into the top of the definition stack, which I'm not sure we can guarantee is empty when running a pass (in theory a user could use a pass while inside a definition).

magma/circuit.py
leonardt: Just for organization, I could make them helper functions outside the class, or make them staticmethods

magma/bits.py
leonardt: `as_bits` needs to be a a function on a value (given a value, return me the bits representation).  We could add an interface to convert a meta class (type) to a Bits, but we would maybe want to use a different name to clearly distinguish the behavior. (e.g. as_Bits or to_Bits).

`from_bits` is a classmethod on the type, so I can say something like `T.from_bits`, so create me a T from an instance of Bits.  I think this is a reasonable organization.

magma/circuit.py
rsetaluri: That makes sense...what about something like this:

```python
def open_definition(defn):
    return _PlacerBlock(defn._placer)
```

Then later, instead of 

```python
            wire_inst = definition.add_instance(
                Wire, T, name=name
            )
            self.wire_map[driver] = wire_inst
            if issubclass(T, Digital):
                wire_inst.I @= driver
            else:
                wire_inst.I @= driver.as_bits()
```

we could write

```python
            with open_definition(definition):
                wire_inst = Wire(T, name=name)
                self.wire_map[driver] = wire_inst
                if issubclass(T, Digital):
                    wire_inst.I @= driver
                else:
                    wire_inst.I @= driver.as_bits()
```

I like this pattern better because it makes the places where we edit circuits very explicit. I've thought about it a bit, and having circuit editing methods on the classes directly seems dangerous.

magma/interface.py
```diff
@@ -52,22 +52,55 @@ class _Interface(Type):
     def __str__(self):
         return str(type(self))
 
+    def _make_wire_str(self, driver, value):
```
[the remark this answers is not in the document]
rsetaluri: or even better

```python
    def _make_wire_str(self, driver, value):
        if isinstance(driver, (Array, Tuple)) and \
                not driver.iswhole(driver.ts):
            return "".join(_make_wire_str(d, v) for d, v in zip(driver, value))
        iname = value.name.qualifiedname()
        oname = driver.name.qualifiedname()
        return f"wire({oname}, {iname})\n"
```

magma/circuit.py
```diff
@@ -145,6 +184,19 @@ def __repr__(cls):
             return f"{name} = DeclareCircuit(\"{name}\", {args})"
         s = f"{name} = DefineCircuit(\"{name}\", {args})\n"
         sorted_instances = sorted(cls.instances, key=lambda x: x.name)
+
+        values = []
```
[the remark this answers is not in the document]
leonardt: This will treat circuits with different intermediate values as unique, whereas before if they had the same structure ignoring intermediates (since ports would be directly connected), they would be treated as the same.  This shouldn't break anything, since it will be more conservative, but it may uniquify in cases where the structure is identical but intermediates used were different.  This is desired in the case of debug mode (when we are code generating the intermediates).  I would say that this would be a better default, and to "optimize" we could set a flag (or check nodebugmode) and ignore intermediates in the uniqueness check.

leonardt:
Comments addressed, I'm going to merge this into https://github.com/phanrahan/magma/pull/584 before going into master as some of these changes are useful for updating the fault/mantle repos (address some issues in the previous branch).

magma/circuit.py
```diff
@@ -124,6 +126,57 @@ def _get_interface_decl(cls):
     return None
 
 
+def add_intermediate_value(value, values):
```
rsetaluri: make these 2 functions private with `_`

magma/interface.py
```diff
@@ -1,7 +1,7 @@
 from itertools import chain
 from collections import OrderedDict
 from .conversions import array
-from .ref import AnonRef, InstRef, DefnRef, LazyDefnRef
+from .ref import AnonRef, InstRef, DefnRef, LazyDefnRef, ArrayRef, TupleRef, NamedRef
```
rsetaluri: looks like `AnonRef`, `ArrayRef`, and `TupleRef` imports are not used in this file

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,103 @@
...
+from ..circuit import Circuit, DeclareCoreirCircuit, _PlacerBlock
+from ..t import In, Out
+from ..digital import Digital
+from ..array import Array
+from ..tuple import Tuple
+from ..bits import Bits
+from ..ref import NamedRef
+
+
+def simulate_wire(self, value_store, state_store):
```
rsetaluri: make private with `_`

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,103 @@
...
+            driver_name = driver.name.qualifiedname("_")
+            value_name = value.name.qualifiedname("_")
+            driver_name = _sanitize_name(driver_name)
+            value_name = _sanitize_name(value_name)
+
+            # Rename driver so it doesn't conflict with new wire instance name
+            if isinstance(driver.name, NamedRef):
+                driver.name.name = "_" + driver.name.name
+            name = f"{driver_name}"
+            with _PlacerBlock(definition._placer):
```
rsetaluri: this is better but I think having an explicit open() function in circuit is good as an external API, and keeping _PlacerBlock as an internal to the circuit implementation. It will pretty much dispatch to this but hides those details. Also, I think all the wiring should happen in that with context. Right now it doesn't make a difference but I could forsee checking that wiring always happens inside in the right context, and either way it is good practice.

magma/t.py
```diff
@@ -3,7 +3,7 @@
 import enum
 from abc import abstractmethod
 from .common import deprecated
-from .ref import Ref, AnonRef, DefnRef, InstRef
+from .ref import AnonRef, NamedRef, DefnRef, InstRef, ArrayRef, TupleRef
```
rsetaluri: looks like `ArrayRef`, `TupleRef` not used here

magma/passes/insert_coreir_wires.py
leonardt: How about
```
    @classmethod
    def open(cls):
        return _PlacerBlock(cls._placer)
```
and
```
            with definition.open():
                wire_inst = Wire(T, name=name)
                self.wire_map[driver] = wire_inst
                if issubclass(T, Digital):
                    wire_inst.I @= driver
                else:
                    wire_inst.I @= driver.as_bits()
```

we don't enforce the wiring happens inside the context, but we can have this be the documented pattern (all edits need to go inside the open() context)

PR #585, opened 2020-02-12
Refactor Port -> Wire

leonardt:
Repull of https://github.com/phanrahan/magma/pull/583 (this broke mantle/fault, so I'm going to wait until those repos are ready)


rsetaluri:
what's the deal here? waiting on mantle/fault repo's?

leonardt:
yes

magma/circuit.py
```diff
@@ -124,6 +126,57 @@ def _get_interface_decl(cls):
     return None
 
 
+def add_intermediate_value(value, values):
```
rsetaluri: make these 2 functions private with `_`

magma/interface.py
```diff
@@ -1,9 +1,8 @@
 from itertools import chain
 from collections import OrderedDict
 from .conversions import array
-from .ref import AnonRef, InstRef, DefnRef, LazyDefnRef
-from .t import Type, Kind, MagmaProtocolMeta
-from .port import INPUT, OUTPUT, INOUT
+from .ref import AnonRef, InstRef, DefnRef, LazyDefnRef, ArrayRef, TupleRef, NamedRef
```
rsetaluri: looks like `AnonRef`, `ArrayRef`, and `TupleRef` imports are not used in this file

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,103 @@
...
+from ..circuit import Circuit, DeclareCoreirCircuit, _PlacerBlock
+from ..t import In, Out
+from ..digital import Digital
+from ..array import Array
+from ..tuple import Tuple
+from ..bits import Bits
+from ..ref import NamedRef
+
+
+def simulate_wire(self, value_store, state_store):
```
rsetaluri: make private with `_`

magma/passes/insert_coreir_wires.py
```diff
@@ -0,0 +1,103 @@
...
+            driver_name = driver.name.qualifiedname("_")
+            value_name = value.name.qualifiedname("_")
+            driver_name = _sanitize_name(driver_name)
+            value_name = _sanitize_name(value_name)
+
+            # Rename driver so it doesn't conflict with new wire instance name
+            if isinstance(driver.name, NamedRef):
+                driver.name.name = "_" + driver.name.name
+            name = f"{driver_name}"
+            with _PlacerBlock(definition._placer):
```
rsetaluri: this is better but I think having an explicit open() function in circuit is good as an external API, and keeping _PlacerBlock as an internal to the circuit implementation. It will pretty much dispatch to this but hides those details. Also, I think all the wiring should happen in that with context. Right now it doesn't make a difference but I could forsee checking that wiring always happens inside in the right context, and either way it is good practice.

magma/t.py
```diff
@@ -3,7 +3,7 @@
 import enum
 from abc import abstractmethod
 from .common import deprecated
-from .ref import Ref, AnonRef, DefnRef, InstRef
+from .ref import AnonRef, NamedRef, DefnRef, InstRef, ArrayRef, TupleRef
```
rsetaluri: looks like `ArrayRef`, `TupleRef` not used here

PR #588, opened 2020-02-18
Add primitive python simulation methods and tuple hotfix

leonardt:
Adds support for Bit primitives in python simulator, it would be nice if we didn't have to cast to/from the `hwtypes` Bit (i.e. the Python simulator used that internally).  That would vastly simplify this code.  Also the use of `value_store` and `state_store` is a bit cumbersome, ideally it would give us our IO and state values directly (perhaps we could register these up front then just refer to them as simulation method arguments).

### DIFF
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -3,8 +3,11 @@ Defines a subtype of m.Array called m.Bits
 
 m.Bits[N] is roughly equivalent ot m.Array[N, T]
 """
+import operator
+import keyword
 from functools import lru_cache, wraps
 import typing as tp
+import hwtypes as ht
 from hwtypes import BitVector
 from hwtypes import AbstractBitVector, AbstractBitVectorMeta, AbstractBit, \
     InconsistentSizeError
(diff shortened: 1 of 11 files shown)


leonardt:
Also adds a hot fix for simulating a tuple.

leonardt:
Still need to add uint/sint

leonardt:
primitive simulation methods finished!

tests/test_type/test_bit.py
```diff
@@ -262,6 +263,12 @@ def definition(io):
     assert check_files_equal(__file__, f"build/TestBitInvert.v",
                              f"gold/TestBitInvert.v")
 
+    sim = PythonSimulator(TestInvert)
```
rsetaluri: Looks like this (and below stuff) implies that the python simulator is part of the main magma spec; meaning if the simulator is broken, magma is broken. That is ok - just making sure that is a conscious decision.

magma/bits.py
```diff
@@ -157,10 +190,27 @@ def declare_binary_op(cls, op):
     @lru_cache(maxsize=None)
     def declare_compare_op(cls, op):
         N = len(cls)
+        python_op_name = {
```
rsetaluri: looks like there are a bunch of different mappings for these op names. Can we factor them out to somewhere common?

magma/bits.py
```diff
@@ -78,6 +80,7 @@ def __str__(cls):
 
 class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
     __hash__ = Array.__hash__
+    hwtypes_T = ht.BitVector
```
rsetaluri: are these used anywhere besides locally? if outside users don't need them, would suggest not including them in the class namespace.

tests/test_type/test_bit.py
leonardt: Yea, that's the reason I added the test (so in future changes we would have caught errors related to these code paths)

magma/bits.py
leonardt: I think leaving it in the class namespace could be useful for inheritance.  For example, if a subclass a UInt, but I want to inherit the simulate logic (e.g. just a named type that behaves exactly as a Uint) it would be convenient to just inherit this property.

leonardt: Similarly, if the user wanted to override the simulation implementation of the primitive, they could override this name and then reuse the primitive simulation methods (they just need to define the operators on a new type with the desired behavior)

leonardt: (that's what UInt and SInt are doing by simply overriding this name, while inheriting the Bits logic for the primitive simulation methods)

PR #589, opened 2020-02-20
Add ability to ignore undriven/unused errors

leonardt:
Useful for formal verification flows where undriven inputs are treated in a special fashion.

This adds flags for magma to ignore undriven/unused value errors.  Specifying a flag to be true in the top level circuit will cause magma to run a pass that automatically marks any undriven or unused value as desired, which avoids any downstream errors.


magma/digital.py
```diff
@@ -206,19 +206,20 @@ def undriven(self):
...
+            renamed_ports = m.circuit.coreir_port_mapping
+            name = f"corebit_{_name}"
+            coreir_name = _name
+            coreir_lib = "corebit"
+
+            def simulate(self, value_store, state_store):
+                pass
+
+            io = m.IO(**{port: direction(Digital)})
+        return _Primitive
```
leonardt: Refactor to avoid deprecation warning from old IO syntax (which is used implicitly by declare circuit.

magma/circuit.py
```diff
@@ -518,10 +519,23 @@ def __new__(metacls, name, bases, dct):
         self.default_kwargs = dct.get('default_kwargs', {})
         self.firrtl = None
 
+        # If enabled, we first set interface ports undriven so it is treats a
```
rsetaluri: I feel like these passes should happen at compile time (or some later time), not at define time (i.e. in the metaclass). I know putting it here means you avoid getting a bunch of `ERROR`'s but those ERROR's aren't blocking anyway, and it may be a good idea to still have those. I feel like it would be better to have another kwarg in m.compile that will add all of these, ala `drive_undriven` and `terminate_unused`. On that note I would change the names of the passes to this. "ignore" seems like passive suppression of warnings rather than actively changing things to avoid them. Also, we could have a single function (with those names) that would apply that pass on a definition, so the user could run that pass without dealing with the "pass" itself.

If we want to configure whether or not we `ERROR` on undriven things, I think we should add a separate flag/config for that -- that would just configure how we report those things, not modify anything.

magma/digital.py
rsetaluri: Great!

magma/passes/ignore_errors.py
```diff
@@ -0,0 +1,32 @@
+from abc import abstractmethod
+from .passes import DefinitionPass
+
+
+class IgnoreErrorsPass(DefinitionPass):
```
rsetaluri: see note above about names

magma/circuit.py
leonardt: updated to use those kwargs and passes at the end and added a flag to ignore undriven errors

magma/passes/passes.py
```diff
@@ -39,6 +39,30 @@ def run(self):
         return self
 
 
+class CircuitPass(Pass):
```
rsetaluri: Since we're making serious use of the passes, let's take this opportunity to clean up these designs.

`CircuitPass` seems like a good idea, but `DefinitionPass` can probably inherit from it:

```python
class DefinitionPass(CircuitPass):
    def _run(self, circuit):
        if not isdefinition(circuit):
            return
        super()._run(circuit)
```
(actually currently in `DefinitionPass` there is a redundant `isdefinition()` check on instance types).

magma/passes/passes.py
```diff
@@ -39,6 +39,30 @@ def run(self):
...
+    """
+    Run on all circuits (not just definitions)
```
rsetaluri: `"""Run on all circuits (definitions and declarations)"""`

magma/passes/passes.py
```diff
@@ -39,6 +39,30 @@ def run(self):
...
+                self(circuit)
+
+    def run(self):
+        self._run(self.main)
+        self.done()
+        return self
+
+
 class DefinitionPass(Pass):
     def __init__(self, main):
```
rsetaluri: docstring: `"""Run only on definitions"""`

magma/passes/unused_undriven.py
```diff
@@ -0,0 +1,32 @@
+from abc import abstractmethod
+from .passes import CircuitPass
+
+
+class UndrivenUnusedPass(CircuitPass):
```
rsetaluri: I like moving this to `passes.py` as follows:

```python
class EditCircuitPass(CircuitPass):
    @abstractmethod
    def edit(self, circuit):
        raise NotImplementedError()

    def __call__(self, circuit):
        with circuit.open():
            self.edit(circuit)
```

and then having the following:

```python
def _drive_undriven(interface):
    undriven = False
    for port in interface.ports.values():
        if port.is_input() and port.trace() is None:
            undriven = True
            port.undriven()


def _terminate_unused(interface):
    terminated = False
    for port in interface.ports.values():
        if port.is_output() and port.wired() is False:
            terminated = True
            port.unused()


class DriveUndrivenPass(EditCircuitPass):
    def edit(self, circuit):
        if _drive_undriven(circuit.interface):
            circuit._is_definition = True
        for inst in circuit.instances:
            _drive_undriven(inst.interface)


class TerminateUnusedPass(EditCircuitPass):
    def edit(self, circuit):
        if _terminate_unused(circuit.interface):
            circuit._is_definition = True
        for inst in circuit.instances:
            _terminate_unused(inst.interface)
```

If possible I would find a better name than `unused_undriven.py`. I also think it's ok to have 2 files: `drive_undrive.py` and `terminate_unused.py`.

magma/passes/unused_undriven.py
```diff
@@ -0,0 +1,32 @@
...
+            for instance in circuit.instances:
+                self.process(circuit, instance.interface.ports.values())
+
+
+class DriveUndriven(UndrivenUnusedPass):
+    def process(self, circuit, port_values):
+        for value in port_values:
+            if value.is_input() and value.trace() is None:
+                # Turn decls into definition
+                circuit._is_definition = True
```
rsetaluri: as an aside, this is why we need to revisit the circuit machinery when editing becomes possible. There's potentially a lot of things which could get out of sync, and `_is_definition` is one of them.

magma/circuit.py
```diff
@@ -607,7 +607,8 @@ def __new__(metacls, name, bases, dct):
                                 "definition syntax is not supported, ignoring "
                                 "'definition'")
         run_unconnected_check = run_unconnected_check or (
-            has_definition and self._syntax_style_ is _SyntaxStyle.NEW)
+            has_definition and self._syntax_style_ is _SyntaxStyle.NEW) and \
+            not dct.get("IGNORE_UNDRIVEN", False)
```
rsetaluri: still not a fan of having this be a class-level flag, but not the worst thing if we really want to enable/disable this at a class granularity. If we do keep it it should be `_ignore_undriven_` per Caleb's standard.

And would have this be on a separate line: `run_unconnected_check  = run_unconnected_check and not dct.get("_ignore_undriven_", False)`

leonardt: Would you prefer it to be a a global flag? I think the rational is that the verification module wants to ignore these errors, but we should keep it on for the RTL so we want to guard it per-circuit.

rsetaluri: ok, seems like there is a need for it per-circuit. fine with this design

magma/circuit.py
```diff
@@ -608,6 +608,8 @@ def __new__(metacls, name, bases, dct):
                                 "'definition'")
         run_unconnected_check = run_unconnected_check or (
             has_definition and self._syntax_style_ is _SyntaxStyle.NEW)
+        run_unconnected_check = run_unconnected_check and not \
```
rsetaluri: double `not`

magma/passes/ignore_undriven.py
```diff
@@ -0,0 +1,18 @@
+from .passes import DefinitionPass
```
rsetaluri: looks like a stale file

magma/passes/ignore_unused.py
```diff
@@ -0,0 +1,18 @@
+from .passes import DefinitionPass
```
rsetaluri: looks like a stale file

PR #597, opened 2020-02-25
Update old style circuit definitions

leonardt:
WIP removal of internal deprecation warnings

### DIFF
--- a/magma/bit.py
+++ b/magma/bit.py
@@ -42,7 +42,7 @@ class Bit(Digital, AbstractBit, metaclass=DigitalMeta):
     def declare_unary_op(cls, op):
         assert op == "not", f"simulate not implemented for {op}"
 
-        class MagmaBitOp(m.Circuit):
+        class _MagmaBitOp(m.Circuit):
             name = f"magma_Bit_{op}"
             coreir_name = op
             coreir_lib = "corebit"
@@ -57,7 +57,7 @@ class Bit(Digital, AbstractBit, metaclass=DigitalMeta):
                 O = int(~I)
                 value_store.set_value(self.O, O)
 
-        return MagmaBitOp
+        return _MagmaBitOp
 
     @classmethod
     @lru_cache(maxsize=None)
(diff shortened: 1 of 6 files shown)


magma/bit.py
```diff
@@ -42,73 +42,81 @@ def get_family() -> TypeFamily:
...
-            O = int(~I)
-            value_store.set_value(self.O, O)
-
-        return m.circuit.DeclareCoreirCircuit(f"magma_Bit_{op}",
-                                              "I", m.In(m.Bit),
-                                              "O", m.Out(m.Bit),
-                                              simulate=simulate,
-                                              coreir_name=op,
-                                              coreir_lib="corebit")
+        class MagmaBitOp(m.Circuit):
```
rsetaluri: let's make all these classes private with leading `_` (e.g. `_MagmaBitOp`)

magma/interface.py
```diff
@@ -379,7 +379,11 @@ def __init__(self, **kwargs):
         self.__decl = []
         for name, typ in kwargs.items():
             ref = LazyDefnRef(name=name)
-            port = typ.flip()(name=ref)
+            if isinstance(typ, MagmaProtocolMeta):
+                port = typ.flip()
+                port = port._from_magma_value_(port._to_magma_()(name=ref))
```
rsetaluri: why do we need to lower and lift here? think I'm missing something. We could make this part of the metacls `__call__` so we don't have to do this everytime we want to instance a value of it.

magma/transforms.py
```diff
@@ -24,8 +24,14 @@ class TransformedCircuit:
     def __init__(self, orig_circuit, transform_name):
         # Maps from original bits to bits in transformed circuit
         self.orig_to_new = {}
-        self.circuit = DefineCircuit(orig_circuit.name + '_' + transform_name,
-                                     *orig_circuit.interface.decl())
+
+        class _TransformedCircuit(m.Circuit):
+            name = orig_circuit.name + '_' + transform_name
+            io_dict = {}
```
rsetaluri: would move this logic outside of the class def and do:

```python
decl = orig_circuit.interface.decl()
ports = dict(zip(decl[::2], decl[1::2]))
...
class _TransformedCircuit(m.Circuit):
    ...
    io = IO(**ports)
    ...
````

magma/interface.py
leonardt: Because the containing type doesn't necessary need to have an interface where it can be constructed with a magma Ref (name).  I discussed with caleb, the options were to have either an enforced interface where a type can be constructed with a ref or to have an interface that constructs an instance from a value constructed with a ref (magma types must satisfy the interface to construct with a ref).  This was more explicit so that's what we chose.  If we want to change that, it's really an interface level change we'd want to discuss with @cdonovick

leonardt: More discussion here: https://github.com/phanrahan/magma/pull/566

Issue #432, opened 2019-08-22
sequential bug when state is not given a new value


leonardt:
```
@m.circuit.sequential(async_reset=True)
class Test:
    def __init__(self):
        self.x: m.Bits[8] = m.bits(0xFE, 8)

    def __call__(self, index: m.Bits[3]) -> m.Bit:
        return self.x[index]
```

gives the error

```
Traceback (most recent call last):
  File "/Users/lenny/repos/magma/magma/ast_utils.py", line 54, in compile_f
unction_to_file
    exec(code, defn_env)
  File ".magma/Test_comb.py", line 5, in <module>
    class Test_comb(m.Circuit):
  File "/Users/lenny/repos/magma/magma/circuit.py", line 471, in __new__
    self.definition()
  File ".magma/Test_comb.py", line 11, in definition
    __magma_ssa_return_value_0 = self_x_I, io.self_x_O[io.index]
NameError: name 'self_x_I' is not defined
```
it should set the default value of register input to be its output.

leonardt:
Similarly, sub-sequential-circuits require a new value (call), both these would be resolved by implicitly generating "enable/valid" logic.

leonardt:
This issue has been noted in the documentation: https://github.com/phanrahan/magma/blob/master/docs/circuit_definitions.md#sequential-circuit-definition

For optional updates of state elements, the algorithm seems straightforward (generate an enable signal that is set high when a write occurs (can default to be 0, with the input being the output value, rewrite the tree such that any write to a state element is followed by setting the enable signal high, optimize out the enable signal in the case that it is always high in every branch).

Optional invocations of sub-circuits is trickier. We can model is as an "implicit valid" on the input values, so a valid low would set all the sub-circuit's state elements to be low. This slightly changes the semantics that `__call__` is invoked every cycle to be `__call__` is invoked only with valid data (which is at most every cycle, and in this case, if valid is always true, then the valid logic can be removed). This extension could imply that we have an option to generate a sequential circuit with a valid input interface (or we could call it an enable).

THofstee:
1. Why are you creating a register in this use-case? You're effectively making a LUT or a ROM, just make it combinational logic and let a synthesis tool create the circuit for it.

[...]

PR #599, opened 2020-02-26
Fix sequential logic for handling writes to registers

leonardt:
Before, sequential was always using the register output value, so it wasn't receiving the latest written value (blocking semantics).  This changes the logic to always use the latest written value by first assigning the register's next value (input) to be the output, then treating all subsequent reads/writes to be of the next value (input).

Also fixes https://github.com/phanrahan/magma/issues/432


magma/bits.py
```diff
@@ -503,6 +503,14 @@ def as_bits(self):
     def from_bits(cls, value):
         return value
 
+    def __getitem__(self, index):
```
rsetaluri: is this an extraneous change?

magma/passes/clock.py
```diff
@@ -1,5 +1,5 @@
 from ..is_definition import isdefinition
-from ..clock import wiredefaultclock
+from ..clock import wiredefaultclock, wireclock
```
rsetaluri: also here? related?

magma/bits.py
leonardt: This was necessary for the test

magma/passes/clock.py
leonardt: This was necessary for a test

leonardt: The backends automatically wire up clocks and clock related signals (reset, enable, ...).  For the simulator, we need to run these passes independently since they don't use a backend

magma/bits.py
leonardt: The issue was with this code
```
    class TestDefault:
         def __init__(self):
             self.x: m.Bits[8] = m.bits(0xFE, 8)

         def __call__(self, index: m.UInt[3]) -> m.Bit:
             return self.x[index]
```
this revealed that get item wasn't implemented.

rsetaluri: ok, just so I understand, the dynamic select as part of `__getitem__`  was part of the spec, just dropped when we moved to the new types? (or is it something we're adding to the spec?)

magma/passes/clock.py
rsetaluri: got it. in that case, should this be called directly by the simulator codepath? rather than adding it to the general pass. fine by me if we think it's generally useful to have this, and should be in the general pass

magma/bits.py
leonardt: yes, it was previously implemented inside mantle (which I think it worked originally, but magma doesn't import mantle to patch on the implementation)

magma/passes/clock.py
leonardt: I think it's generally useful, we can have the simulator call it by default too though

magma/simulator/coreir_simulator.py
```diff
@@ -11,9 +11,10 @@
 from ..bit import Bit
 from ..bitutils import int2seq
 from ..clock import Clock
-from ..transforms import setup_clocks, flatten
+from ..transforms import flatten
 from ..circuit import CircuitType
 from ..uniquification import uniquification_pass, UniquificationMode
+from magma.passes.clock import WireClockPass
```
rsetaluri: from `..passes.clock import WireClockPass`

magma/simulator/python_simulator.py
```diff
@@ -19,6 +19,7 @@
 import hwtypes
 from ..bitutils import seq2int
 from ..clock import Clock
+from magma.passes.clock import WireClockPass
```
rsetaluri: `from ..passes.clock import WireClockPass`

PR #600, opened 2020-02-27
update magma IO syntax

hofstee:


hofstee:
@leonardt this seems like a test bug to me. Why does `tests/test_types/test_clock.py:test_asyncreset_cast` define `Bit_Arr_out` as an output twice?

leonardt:
That does look like a "bug".  Judging by the code, I think the first occurrence is ignored ("overriden" by the second declaration), so it should be okay to delete the first one.

leonardt:
CC @rsetaluri I added `__add__` to IO here and implemented `ClockIO` since it was useful for updating these circuits, it just creates and returns a new IO object

rsetaluri:
whenever we do merge this, let's squash the commits :)

leonardt:
Done! next stop, mantle.

It seems like mostly it's removing
```
    @classmethod
    def definition(io):
```
and de-indenting the contained block,
then it's updating `ClockInterface` to `ClockIO` where necessary, but there were some minor nuanced changes needed which pretty much makes it a manual effort to close it out.

leonardt:
There's also some declare/define circuit deprecations in the test suite that we might just want to remove now as well

rsetaluri:
@hofstee @leonardt does this change get rid of all `IO = []` syntax in the tests? We're promising it's backwards compatible so we definitely some tests with the old syntax. perhaps keeping the old syntax in `tests/test_{circuit,interface}`??

leonardt:
Hmm, it does most of, but not all, i did merge a squashed commit, so we
could do a revert of that and copy some old tests over.

hofstee:
@leonardt: @rsetaluri put some tests here: [tests/test_deprecated/test_old_io_syntax](https://github.com/phanrahan/magma/tree/master/tests/test_deprecated/test_old_io_syntax)

rsetaluri:
think we're good here with the new tests. thanks.

PR #601, opened 2020-02-27
Improve bind flow to handle temporary values

leonardt:
This allows users to create an intermediate/temporary value (e.g. pack value into 2d array), then pass it to the bind string.

The key required feature: We wire up the value to an "Unused" instance, so it gets code generated (otherwise it's ignored as a dangling wire).  This won't matter if it's actually used elsewhere, since then it will just drive multiple places, but we can optimize this out later if needed (but we'll probably go with a different long term solution so i don't think it's too important).

### DIFF
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -140,17 +140,23 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
         N = len(cls)
         python_op_name = primitive_to_python_operator_name_map.get(op, op)
 
-        def simulate(self, value_store, state_store):
-            I = cls.hwtypes_T[N](value_store.get_value(self.I))
-            O = int(getattr(operator, python_op_name)(I))
-            value_store.set_value(self.O, O)
-        return m.circuit.DeclareCoreirCircuit(f"magma_Bits_{N}_{op}",
-                                              "I", m.In(cls),
-                                              "O", m.Out(cls),
-                                              simulate=simulate,
-                                              coreir_name=op,
-                                              coreir_genargs={"width": N},
-                                              coreir_lib="coreir")
+        class _MagmaBitOp(m.Circuit):
+            name = f"magma_Bits_{N}_{op}"
+            coreir_name = op
+            coreir_lib = "coreir"
+            coreir_genargs={"width": N}
+            renamed_ports = m.circuit.coreir_port_mapping
+            primitive = True
+            stateful = False
+
+            io = m.IO(I=m.In(cls), O=m.Out(cls))
+
+            def simulate(self, value_store, state_store):
+                I = cls.hwtypes_T[N](value_store.get_value(self.I))
+                O = int(getattr(operator, python_op_name)(I))
+                value_store.set_value(self.O, O)
+
+        return _MagmaBitOp
 
     @classmethod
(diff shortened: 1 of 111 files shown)


magma/bits.py
```diff
@@ -487,13 +487,13 @@ def get_family(cls):
...
-            raise TypeError("unused can only be used on output")
+        if self.is_input() or self.is_inout():
+            raise TypeError("unused cannot be used with input/inout")
         m.wire(self, DefineUnused(len(self))().I)
 
     def undriven(self):
-        if not self.is_input():
-            raise TypeError("undriven can only be used on input")
+        if self.is_output() or self.is_inout():
+            raise TypeError("undriven cannot be used with output/inout")
```
leonardt: These type checks needed to be updated because they don't properly handle temporary values, which are neither inputs/outputs/inouts

tests/test_verilog/gold/RTLMonitor.sv
```diff
@@ -16,15 +30,37 @@ module RTLMonitor (
     input mon_temp2,
     input out
 );
+wire [3:0] arr_2d_0;
+wire [3:0] arr_2d_1;
+assign arr_2d_0 = in1;
+assign arr_2d_1 = in2;
```
leonardt: Here's the arr_2d temporary named value

tests/test_verilog/gold/RTLMonitor.sv
```diff
@@ -16,15 +30,37 @@ module RTLMonitor (
...
 logic temp1, temp2;
-logic [3:0] temp3;
+logic [4-1:0] temp3;
 assign temp1 = |(in1);
 assign temp2 = &(in1) & intermediate_tuple__0;
-assign temp3 = in0 ^ in1;
+assign temp3 = in1 ^ in2;
 assert property (@(posedge CLK) handshake_valid -> out === temp1 && temp2);
+logic [4-1:0] temp4 [1:0];
+assign temp4 = '{arr_2d_1, arr_2d_0};
```
leonardt: Here's it's use

tests/test_verilog/gold/bind_test.v
```diff
@@ -41,6 +40,5 @@ assign handshake_arr_0_valid = handshake_arr_2_ready;
 assign handshake_arr_1_valid = handshake_arr_1_ready;
 assign handshake_arr_2_valid = handshake_arr_0_ready;
 assign handshake_valid = handshake_ready;
-assign out = logical_and_inst0_O;
```
leonardt: Minor code generation update since it uses `inline=True` (going to move to using this for all new tests to make sure it's working with new features)

tests/test_verilog/test_bind.py
```diff
@@ -20,3 +20,5 @@ def test_bind():
     assert m.testing.check_files_equal(__file__,
                                        f"build/RTLMonitor.sv",
                                        f"gold/RTLMonitor.sv")
+    assert not os.system('cd tests/test_verilog/build && '
+                         'verilator --lint-only bind_test.v RTLMonitor.sv')
```
leonardt: Actually lint the output to make sure it looks correct

PR #604, opened 2020-02-28
Prepend deprecated test file names

rsetaluri:
Master is currently broken because colliding test file names introduced in a15ef781cdd557c3bb80580d45d149c6be5f0c8a. Fixes that


hofstee:
Ok so it looks like in theory you could so the thing I did and just put `__init__.py` inside `test_deprecated/test_io_syntax/` which also gave this same python fatal error.

That said, I have no clue what's going on to get a fatal python error.

Issue #180, opened 2018-04-19
Move assignments from CS448H offering into examples


leonardt:
We should take labs 1 and 2 from https://github.com/rdaly525/cs448h and add them to the examples directory, perhaps using a prompt/solutions style similar to the chisel tutorial (i.e. an implemented version like the labs in the class that a person can use to start learning magma, accompanied by a reference solution so they can check their work)

PR #606, opened 2020-02-29
Regfile primitive

leonardt:
CC @priyanka-raina

This adds prototype support for a staged register file primitive.

The proposed semantics are:
* Instantiate the generator with a height and width
* Write to the register file with the syntax `reg_file[addr] = data`
* read to the register file with the syntax `reg_file[addr]`
* writing and reading append ports to the interface, wire up address and data
* "last connect" semantics, writing multiple times, the last write will have priority (implemented in mux logic)
* read forwarding semantics (this anticipates blocking sequential semantics support), writes in a a cycle will automatically be forwards to reads combinationally (implemented using a mux on the read port address)

Possible enhancements:
* make write forwarding a parameter (so writes can be delayed by a cycle from the view of a read port).  This would be fine in core magma (since there's no "sequential" logic in the code).  Sequential would not use this unless we change the blocking write semantics.
* generate a verilog body instead of structural magma (we can easily implement `finalize` such that it dumps an always block with if statements for the register file)
* pass the parameters to coreir to have it generate the verilog body (treat it as a black box primitive and have corier deal with it)

I'm open to feedback on all aspects, particularly the generator staging logic.  I just hacked it in and patched on things where needed to get the result I wanted, we should probably abstract this into an API.  The key features I needed:
* unwiring ports (for adding in new muxes for new write ports)
* adding instances (covered by the `circuit.open()` API)
* adding ports to the interface (hacked this in by editing the IO and re-running the meta class logic)
* adding ports to the instance (hacked this in by editing the instance attributes and interface object)

Another issue is dealing with uniqueness of generators.  Right now, it turns off cacheing, so every instance has its own unique type to edit.  This is the safest, but may be a performance issue, so we could look into "cacheing" or "memoizing" these edits and sharing created definitions, but perhaps this isn't such a big issue (again I think optimizing elsewhere such as the array logic would be more fruitful).


leonardt:
CC @rsetaluri any thoughts on how we should structure the circuit editing API for this pattern? The main logic is inside magma/primitives/register_file.py with some modificaitons to the circuit class to expose the required functionality.  For example, we edit the circuit ports to append new ports.  Then we re-invoke the original interface setup logic with the new IO.  Perhaps we can have an "add port" action that will abstract this away.  This logic defers any wiring until a `finalize` stage that runs at compile time, so we don't need to worry about existing wiring when recreating the interface.  It also effectively uses "copy on write" since each instance has a unique defn (no cacheing), but we may consider how we want to handle this more formally.  For now, I think using a unique definition for each instance will probably be the safest and easiest to reason about it.

leonardt:
Closing in favor of #180

Issue #609, opened 2020-03-02
test_tuple failure


phanrahan:
When I run the tests on my machine I get the following two failures in `test_tuple.py`.
```
=================================== FAILURES ===================================
_________________________________ test_nested __________________________________

    def test_nested():
        # Test for https://github.com/phanrahan/magma/issues/445
        def hierIO():
            class dictIO(Product):
                baseIO = make_baseIO()
                ctr = m.In(m.Bit)
            return dictIO
    
        def DefineCtrModule():
            class ctrModule(m.Circuit):
                name = "ctr_module"
                io = m.IO("ctr", m.In(m.Bit))
            return ctrModule
    
        def make_baseIO():
            class dictIO(Product):
                in0 = m.In(m.Bit),
                out0 = m.Out(m.Bit)
            return dictIO
    
        def DefineBaseModule():
            class baseModule(m.Circuit):
                name = "base_module"
                io = m.IO("baseIO", make_baseIO())
            return baseModule
    
        def DefineHier():
            class HierModule(m.Circuit):
                name = "hier_module"
                io = m.IO(hier=hierIO())
                @classmethod
                def definition(io):
                    baseM = DefineBaseModule()()
                    ctrM = DefineCtrModule()()
                    m.wire(baseM.baseIO, io.hier.baseIO)
                    m.wire(ctrM.ctr, io.hier.ctr)
            return HierModule
    
        baseMH = DefineHier()
>       m.compile("build/baseMH", baseMH, output="coreir-verilog")

tests/test_type/test_tuple.py:231: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
magma/compile.py:52: in compile
    compiler.compile()
magma/backend/coreir_compiler.py:39: in compile
    self.__compile_verilog()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <magma.backend.coreir_compiler.CoreIRCompiler object at 0xb2389de90>

    def __compile_verilog(self):
        # TODO(rsetaluri): Expose compilation to verilog in pycoreir rather than
        # calling the binary from the command line.
        cmd = _make_verilog_cmd(self.deps, self.basename, self.opts)
[...]

PR #610, opened 2020-03-02
[Issue #609] Fix new IO syntax

rsetaluri:


tests/test_type/test_tuple.py
```diff
@@ -258,9 +256,7 @@ class MyCircuit(m.Circuit):
     def DefineTop(params):
         class Top(m.Circuit):
             io = m.IO(IFC1=IFC1(params))
-            @classmethod
-            def definition(io):
-                m.wire(io.IFC1.port4, DefineMyCircuit(params)().IFC0.port4)
+            m.wire(io.IFC1.port4, DefineMyCircuit(params)().IFC0.port4)
         return Top
 
     m.compile("top", DefineTop({'param0': 5}))
```
rsetaluri: It looks like these tests don't really check that things work, which is why #609 was probably not caught by travis (still not sure why the CoreIR exit code is not set properly on travis vs. in #609...)

leonardt: Hmm yea that's strange, we should probably check the exit code inside compile and raise an exception (if we don't already)

rsetaluri: We do have that check in the code: https://github.com/phanrahan/magma/blob/5bdb58bda377b363824629b719ddb0a4fb433fa8/magma/backend/coreir_compiler.py#L55-L56

but for some reason we're not hitting it on travis...

PR #616, opened 2020-03-05
Catch exceptions in debug mode parsing

leonardt:
Sometimes we cannot parse the line, e.g.
```
x = [Circuit()
     for _ in range(8)]
```
Will give use the second line for the list comprehension, which
cannot be parsed.  It's probably generally difficult to catch all
these cases
For now, we just silently ignore exceptions and just assume we can't
find a good debug name

### DIFF
--- a/magma/placer.py
+++ b/magma/placer.py
@@ -16,14 +16,24 @@ def _parse_name(inst):
 
     otherwise, returns None.
     """
-    with open(inst.debug_info.filename, "r") as f:
-        line = f.read().splitlines()[inst.debug_info.lineno - 1]
-    tree = ast.parse(textwrap.dedent(line)).body[0]
-    # Simple case when <Name> = <Instance>().
-    if isinstance(tree, ast.Assign) and len(tree.targets) == 1 \
-       and isinstance(tree.targets[0], ast.Name):
-        return tree.targets[0].id
-    return None
+    try:
+        with open(inst.debug_info.filename, "r") as f:
+            line = f.read().splitlines()[inst.debug_info.lineno - 1]
+        tree = ast.parse(textwrap.dedent(line)).body[0]
+        # Simple case when <Name> = <Instance>().
+        if isinstance(tree, ast.Assign) and len(tree.targets) == 1 \
+           and isinstance(tree.targets[0], ast.Name):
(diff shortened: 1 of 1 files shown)


magma/placer.py
```diff
@@ -16,14 +16,24 @@ def _parse_name(inst):
...
     """
-    with open(inst.debug_info.filename, "r") as f:
-        line = f.read().splitlines()[inst.debug_info.lineno - 1]
-    tree = ast.parse(textwrap.dedent(line)).body[0]
-    # Simple case when <Name> = <Instance>().
-    if isinstance(tree, ast.Assign) and len(tree.targets) == 1 \
-       and isinstance(tree.targets[0], ast.Name):
-        return tree.targets[0].id
-    return None
+    try:
```
rsetaluri: Can we localize the source of the cases we can't handle? Should we just wrap L22 in try/catch? Or do we forsee that other parts could error and this is the safe option?

leonardt: We can wrap L22

leonardt: I think it has to do when a statement is broken onto multiple lines

leonardt: it seems that it takes the last linenumber of the statement, so we would have to trace back a possibly arbitrary number of lines

leonardt: One option might be to parse the whole file, then walk the AST to find the Assign node on that lineno (the AST nodes contain the source line as a field), but this might be expensive compared to a string index lookup

leonardt: Here's a small case:
```python
# test.py
import inspect


def foo():
    callee_frame = inspect.getframeinfo(inspect.stack()[1][0])
    print(f"file_name={callee_frame.filename}, lineno={callee_frame.lineno}")
    return range(0, 8)

a = tuple(i
          for
          i
          in
          foo())
```

```shell
❯ python test.py
file_name=test.py, lineno=13
```

So when we get the function call (e.g. when an instance is created), it's on a floating line.  The only reasonable way to map this back to the original assign would be inspecting the AST i think (find the corresponding node, walk to the parent statement)

rsetaluri: Makes sense, thanks for clarifying

PR #618, opened 2020-03-06
Disallow instancing outside of defn

rsetaluri:
** There's a chance this affects downstream tools! But that's wrong usage anyway and should be fixed **

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -74,14 +74,10 @@ class _PlacerBlock:
 
     @classmethod
     def pop(cls):
-        if not cls.__stack:
-            return None
         return cls.__stack.pop()
 
     @classmethod
     def peek(cls):
-        if not cls.__stack:
-            return None
         return cls.__stack[-1]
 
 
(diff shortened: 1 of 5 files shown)


magma/circuit.py
```diff
@@ -516,8 +516,9 @@ def __init__(self, *largs, **kwargs):
         super(CircuitType, self).__init__(*largs, **kwargs)
         # Circuit instances are placed if within a definition.
         placer = _PlacerBlock.peek()
-        if placer:
-            placer.place(self)
+        if placer is None:
+            raise Exception("Can not instance a circuit outside a definition")
+        placer.place(self)
```
leonardt: Maybe we can do a try/except here since this should be the uncommon case? (Avoid the branch overhead in the usual case)

tests/test_type/top.json
```diff
@@ -0,0 +1,26 @@
...
+          }
+        },
+        "connections":[
+          ["self.IFC1.port4","MyCircuit_inst0.IFC0.port4"]
+        ]
+      }
+    }
+  }
+}
+}
```
leonardt: Extra?

rsetaluri: yes, removed

magma/circuit.py
leonardt: Looks like PlacerBlock.peek() handles the None return logic, so the try except could be inside there looking for an `IndexError: list index out of range` (so we don't accidentally catch unrelated exceptions)

rsetaluri: Good points. See the lastest...think it's more pythonic this way

PR #620, opened 2020-03-07
Fix CoreIR wrap pass

rsetaluri:
Using `open()`!

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -84,22 +84,24 @@ class InsertWrapCasts(DefinitionPass):
         if isinstance(port, (Array, Tuple)):
             for t in port:
                 self.wrap_if_named_type(t, definition)
-        elif port.is_input():
-            if isinstance(port, (AsyncReset, AsyncResetN, Clock)) or \
-                    isinstance(port.trace(), (AsyncReset, AsyncResetN, Clock)):
-                value = port.trace()
-                if value is not None and not issubclass(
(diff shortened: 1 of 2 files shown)


magma/backend/coreir_.py
```diff
@@ -84,22 +84,25 @@ def wrap_if_named_type(self, port, definition):
...
-                    wire(inst.out, port)
+            return
+        if not port.is_input():
+            return
+        if not (isinstance(port, (AsyncReset, AsyncResetN, Clock)) or
+                isinstance(port.trace(), (AsyncReset, AsyncResetN, Clock))):
+            return
+        value = port.trace()
+        if not (value is not None and
+                not issubclass(type(value), type(port).flip())):
```
leonardt: I think we can remove the double negative and make it an or?

magma/backend/coreir_.py
```diff
@@ -84,22 +84,25 @@ def wrap_if_named_type(self, port, definition):
...
+            return
+        if isinstance(port, (AsyncReset, AsyncResetN, Clock)):
+            T = type(port).flip()
+        else:
+            T = type(value).flip()
+        with definition.open():
+            port.unwire(value)
+            inst = self.define_wrap(T, type(port), type(value))()
+            getattr(inst, "in") <= value
```
leonardt: Let's update this to `@=` if we can (but I think maybe there's a problem with "assigning to getattr" not working, in which case maybe just swap it with `wire`?

PR #621, opened 2020-03-09
Init coroutine syntax

leonardt:
This is an initial PR adding Silica's coroutine capabilities to magma by building on top of sequential.

A magma coroutine is a sequential circuit that can use `yield` inside the `__call__` method.

The implementation requires a fork of the `staticfg` package to support `break` statements.  I will be submitting a PR to the mainline shortly, but since this is an experimental feature I think the `requirements.txt` pattern should suffice.  Anyone interested in using this can manually install the fork easily.

The implementation uses the basic "tracing" algorithm that generates the logic for each yield.  The SSA algorithm will come in a follow up pull request.  As will more complicated circuits/tests for the JTAG and SSA.

This also adds support for the `self.<register_name>.prev()` syntax proposed by @phanrahan .  This will return the previous value of the register, even if it has already been updated with a blocking assignment.  This is useful for circuits that register their outputs.  It is syntax sugar for storing the current value in a temporary, then referring to the temporary (avoiding the syntactic overhead for the common case of registered outputs). 

The gold files for some tests were updated because the SSA logic was changed to use the `~` operator primitive instead of the `Not` mantle function.  Also needed to fix a bug in the SSA logic for nested returns.

There's some updates to old declare circuit syntax to avoid deprecation warnings that came up in the tests.

### DIFF
--- a/tests/test_syntax/test_sequential.py
+++ b/tests/test_syntax/test_sequential.py
@@ -94,8 +94,8 @@ def DefineCoreirReg(width, init=0, has_async_reset=False,
         coreir_name = "reg_arst" if (has_async_reset or has_async_resetn) else "reg"
         verilog_name = "coreir_" + coreir_name
         coreir_lib = "coreir"
-        simulate=gen_sim_register(width, init, False, has_async_reset,
-                                  has_async_resetn)
+        simulate = gen_sim_register(width, init, False, has_async_reset,
+                                    has_async_resetn)
     return CoreIRReg
 
 
--- a/tests/test_syntax/test_sequential.py
+++ b/tests/test_syntax/test_sequential.py
@@ -86,6 +86,7 @@ def DefineCoreirReg(width, init=0, has_async_reset=False,
         io = m.IO(**io_dict)
         renamed_ports = m.circuit.coreir_port_mapping
         stateful = True
+        primitive = True
         circuit_type_methods = methods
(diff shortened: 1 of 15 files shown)


rsetaluri:
Cool! Just wondering in terms of what code lives where - what is the overlap of this work and silica? Is the plan to migrate silica *into* magma, or keep both separate, or import from silica?

leonardt:
migrate silica into magma.  silica uses it's own verilog code generation path that's pretty brittle.  the idea is to remove that complexity by just leveraging magma (e.g. using sequential).  That simplified the implementation quite a bit

leonardt:
(It also provides the feature to sequential users)

tests/test_syntax/test_sequential.py
```diff
@@ -68,43 +67,36 @@ def reset(self, condition):
...
+        stateful = True
+        primitive = True
+        circuit_type_methods = methods
+        default_kwargs = {"init": coreir.type.BitVector[width](init)}
+        coreir_genargs = gen_args
+        coreir_configargs = config_args
+        coreir_name = "reg_arst" if (has_async_reset or has_async_resetn) else "reg"
+        verilog_name = "coreir_" + coreir_name
+        coreir_lib = "coreir"
         simulate=gen_sim_register(width, init, False, has_async_reset,
```
rsetaluri: nit: space between `=`

PR #623, opened 2020-03-09
Update monitor/inline syntax for new style

leonardt:
This adds support for defining inline verilog using the new `m.IO` style syntax.

The old inline verilog syntax required using `cls.inline_verilog`, it also required that the port names be not anonymous.

The new flow stages the realization of the inline verilog names until later on (once the anonymous ports have been resolved to their definition ports).  It also uses a magic `_inline_verilog_` class variable for constructing these strings.  The interface allows users to incrementally construct the string using `+=` similar to the old interface.


rsetaluri:
I have a proposal that I think will enable a lot of features we've been thinking about (see end) and also make this feature cleaner. Though this change is kind of necessitated by the new syntax, I actually think it's a bit more ergonomic and more importantly separates out a lot of these features from the main Circuit code path (ala separation of concerns).

The main idea is to make these features like the placer pattern, where we can collect a bunch of stuff during the course of circuit definition, and then finalize them and process them once the circuit has been created. I think it makes the various stages more clear. 

#625 is what would be needed (pretty simple, placer -> more general context). This context can then be used to e.g. capture and stage logging information, allow nice syntaxes like `m.display()` or `m.assert()`, and even `m.inline_verilog()`. If we go this direction the user syntax would look like

```python
class Foo(m.Circuit):
    io = m.IO(...)
    m.inline_verilog(...)
    m.print(...)
```
(note how we don't have to do `_inline_verilog_ = ...` and `_inline_verilog_ += ...`)

I would suggest that we add well defined functions to `DefinitionContext` class in #625, such as `inline_verilog()`, and then remove the class method `inilne_verilog()`.

leonardt:
@rsetaluri updated to use new pattern, I'd like to leave `inline_verilog` class method for now since it's already being used, but we can deprecate it.

magma/circuit.py
```diff
@@ -58,6 +59,18 @@ class _SyntaxStyle(enum.Enum):
 class DefinitionContext:
     def __init__(self, placer):
         self.placer = placer
+        self._inline_verilog = []
```
rsetaluri:
```suggestion
        self._inline_verilog = []
        self._finalized = False
```

magma/circuit.py
```diff
@@ -58,6 +59,18 @@ class _SyntaxStyle(enum.Enum):
...
+    def add_inline_verilog(self, format_str, format_args):
+        self._inline_verilog.append((format_str, format_args))
+
+    def finalize(self, defn):
+        final_placer = self.placer.finalize(defn)
+        final_context = DefinitionContext(final_placer)
+        with _DefinitionContextManager(final_context):
+            for format_str, format_args in self._inline_verilog:
+                defn.inline_verilog(format_str, **format_args)
+        return final_context
```
rsetaluri:
```suggestion
    def finalize(self, defn):
        if self._finalized:
            raise Exception("Can only call finalize on a definition context once")
        with _DefinitionContextManager(self):
            for format_str, format_args in self._inline_verilog:
                defn.inline_verilog(format_str, **format_args)
        self.placer = self.placer.finalize(defn)
        self._finalized = True
```

magma/circuit.py
```diff
@@ -210,8 +223,7 @@ def __new__(metacls, name, bases, dct):
         try:
             context = _definition_context_stack.pop()
             assert context.placer.name == cls_name
-            placer = context.placer.finalize(cls)
-            cls._context_ = DefinitionContext(placer)
+            cls._context_ = context.finalize(cls)
```
rsetaluri:
```suggestion
            cls._context_ = context
            cls._context_.finalize(cls)
```

magma/inline_verilog.py
```diff
@@ -0,0 +1,9 @@
+from .circuit import _definition_context_stack
+from .verilog_utils import value_to_verilog_name
+from .t import Type
+from .view import PortView
+
+
+def inline_verilog(format_str, **kwargs):
+    context = _definition_context_stack.peek()
```
rsetaluri:
```suggestion
    try:
        context = _definition_context_stack.peek()
    except IndexError:
        raise Exception("Can not inline verilog outside a definition")
```

magma/circuit.py
```diff
@@ -326,8 +338,12 @@ def inline_verilog(cls, inline_str, **kwargs):
             elif isinstance(arg, PortView):
                 arg = value_to_verilog_name(arg)
             format_args[key] = arg
-        cls.inline_verilog_strs.append(
-            inline_str.format(**format_args))
+        cls.inline_verilog_strs.append(inline_str.format(**format_args))
+
+    @deprecated(msg="cls.inline_verilog is deprecated, use m.inline_verilog"
```
rsetaluri:
```suggestion
    @deprecated(msg="cls.inline_verilog is deprecated, use m.inline_verilog "
```

magma/circuit.py
```diff
@@ -58,6 +59,18 @@ class _SyntaxStyle(enum.Enum):
...
+
+    def add_inline_verilog(self, format_str, format_args):
+        self._inline_verilog.append((format_str, format_args))
+
+    def finalize(self, defn):
+        final_placer = self.placer.finalize(defn)
+        final_context = DefinitionContext(final_placer)
+        with _DefinitionContextManager(final_context):
+            for format_str, format_args in self._inline_verilog:
+                defn.inline_verilog(format_str, **format_args)
```
rsetaluri: if we're deprecating `cls.inline_verilog()` then this codepath should ideally not use it. I would ideally move all the logic in `_inline_verilog` to an external function in `inline_verilog.py` that does the same thing but then all the logic is one file. And the deprecated `inline_verilog()`calls to that same function. Avoids the deprecation warning on the new API.

cdonovick:
I haven't done a thorough review of how this code works. But this sort of stack introspection always scares me.  I know its not ergonomic but I would agvocate having people manually pass the environment.  

I don't think how you are formatting would work with non local names but maybe I am missing something.  It looks like its trying to format in the calling scope and if that fails try the next scope up etc... Is it partially formatting each time?  If so than I am wrong.  If it isn't I don't think this would work:

```python
def f():
  x = 5
  def g():
     y = 6
    inline("{x} {y}") 
```

I would just walk all the way up the call stack building a chainmap like in ast tools. I guess now that I think about I would just use asttools if you really want to save people the effort of manually passing the environment. 
https://github.com/leonardt/ast_tools/blob/master/ast_tools/stack.py#L45 

Also wrapping inline in anyway would be very unsafe as it would leak names (why I don't like the stack introspection).  

Alternative idea: have magma types define `__format__` and just use f strings
https://docs.python.org/3/reference/datamodel.html#object.__format__
`__format__` could call `value_to_verilog_name`.  Note `__format__` allows you to define your own format_spec so you could have `{m_val:debug}`, `{m_val:inline}`, and `{m_val}` produce different strings.

magma/circuit.py
leonardt: Most of this logic is moved out of the file in the inline-env PR so I'll finish the cleanup there.

leonardt:
Hmm using the __format__ method might be the best option here, I will investigate it.

leonardt:
Actually it doesn't quite work given the staged nature of the execution (we can't format until later on, once the staged placer has been resolved)

leonardt:
This is a temporary feature that doesn't need to work in every single possible case.  It's just to be used as a stop gap solution for producing inline verilog properties, debug statements, etc... until we add native support to magma, so I wouldn't worry too much about covering every possibility

leonardt:
Updated it to use ast_tools symbol table, figured at least we can share the code path to maintain there.  For now, we'll enforce a rule that `inline_verilog` must be invoked in the class level definition (so it can't be wrapped in a function).  I think this should cover most of the basic use cases this is intended for (interpolating multi-dimensional arrays into a verilog module instantation)

PR #624, opened 2020-03-09
Add support for f-string like behavior in inline

leonardt:
Depends on https://github.com/phanrahan/magma/pull/623

This is a feature that implements f-string like behavior for inline verilog to make it easier to inline simple expressions.

It tracks the calling frame of `m.inline_verilog` and uses it to evaluate keys in the format string that are not present in the kwargs.  It has to use stack tracing logic to handle the expected environment behavior, which can be expensive, but is probably okay for most use cases and will hopefully be avoided once we add better syntax for python properties etc...


rsetaluri:
Following the comment on #623, I think we want to have more separation of concerns. I think the principle is that we can have temporary solutions but they should be as outside of the main code path as possible. So I think if we have a single entry point `m.inline_verilog()` (which is *almost* what we have) then functions like `_convert_values_to_verilog_str` and `_process_inline_verilog` can be outside of the main circuit code path (maybe more just a code organization thing than anything else).

Separately, I think it's better to have some specialized syntax for interpolating, rather than exec'ing code and grabbing variables and all that (seems prone to error and hard to maintain -- basically trying to capture the entirety of f-string semantics). I think interpolating generic python variables like `width` should be explicitly different than doing so for magma ports. For the former, normal f-string syntax would be used, and for the latter we would have some special escape, so

```python
                arr_2d[i] @= getattr(io, f"in{i + 1}")
            _inline_verilog_ = m.inline_verilog("""
logic temp1, temp2;
logic [{width-1}:0] temp3;
assign temp1 = |(in1);
assign temp2 = &(in1) & {io.intermediate_tuple[0]};
assign temp3 = in1 ^ in2;
assert property (@(posedge CLK) {valid} -> out === temp1 && temp2);
logic [{width-1}:0] temp4 [1:0];
assign temp4 = {arr_2d};
                                   """,
                                   valid=io.handshake.valid)
```

would become something like

```python
                arr_2d[i] @= getattr(io, f"in{i + 1}")
            m.inline_verilog(f"""
logic temp1, temp2;
logic [{width-1}:0] temp3;
assign temp1 = |(in1);
assign temp2 = &(in1) & %%p;
assign temp3 = in1 ^ in2;
assert property (@(posedge CLK) {valid} -> out === temp1 && temp2);
logic [{width-1}:0] temp4 [1:0];
assign temp4 = {arr_2d};
                                   """,
                                   io.handshake.valid)
```

where `%%p` implies a magma port (just picked something random). We could do `%%<port name>%%` to get named args. Probably can do better for the escape sequences :D -- but the point is making these two things explicitly different. Then downstream, we could use a simple regex/templating to resolve those. Think the small overhead on users is worth the simplicity in implementation and clarity.

leonardt:
FYI, that exact overhead is what this PR was attempting to address.

doing something like `{io.handshake.valid}` is the desired feature, because they don't want to have to name every port they pass through

leonardt:
So while it adds complexity, this is something that's been externally requested.  Without adding f-string support for ports, I suggest we scrap this feature all together, since their issue isn't really related to inlining parameters as much as inlining value references (for writing assertions) so the convenience provided by this syntax is mainly targeting that use case.  Without adding support for that in the f-string style, then using the existing inline_verilog syntax with explicit kwargs is fine

leonardt:
I moved most of the logic outside of the main codepath, but I left `_inline_verilog` as a private method since I think it makes the most sense for handling accessing the class members.

leonardt:
I don't think we need to worry about supporting/maintaining this longterm, I suspect we will only need it for a few months, so I don't think investing a large amount of time/effort into this is necessary.  Again, they will only be using the most basic cases, and we can just suggest alternative workarounds if they run into weird corner cases.  The more time we invest in this code path, the longer we delay the long term solution, which I'd like to minimize.

PR #625, opened 2020-03-09
Generalize placer to definition context

rsetaluri:


rsetaluri:
See discussion on #623 for motivation

PR #628, opened 2020-03-10
Propose new generator class

rsetaluri:
Proposing new (non-staged) generator syntax. Most notable is the type relationships (Generator -> Circuit -> Instance), and the ease of writing generators (free form in init). There are some TODO's but would be great to get feedback.

TODO:
-- ~~Caching~~
-- Make dataclass-like with parameters

### DIFF
--- a/magma/generator.py
+++ b/magma/generator.py
@@ -1,6 +1,7 @@
-from abc import abstractmethod
-import functools
+from abc import abstractmethod, ABCMeta
 import collections
+import functools
+from .circuit import DefineCircuitKind, Circuit
 
 
 class ParamDict(dict):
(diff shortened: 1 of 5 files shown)


tests/test_generator2.py
```diff
@@ -0,0 +1,82 @@
...
+    MyMux4x8_other = _MyMux(4, 8)
+    assert MyMux4x8 is MyMux4x8_other
+    MyMux4x16 = _MyMux(4, 16)
+    assert MyMux4x16 is not MyMux4x8
+
+
+def test_no_cache():
+
+    class _MyGen(m.Generator2):
+        _no_cache_ = True
```
rsetaluri: not tied to this name, but I like the negative default being false (rather than `cache = True` being default)

magma/generator.py
```diff
@@ -55,3 +57,58 @@ def generate(*args, **kwargs):
...
+
+
+class _Generator2Meta(type):
+    _cache = weakref.WeakValueDictionary()
+
+    def __new__(metacls, name, bases, dct):
+        bases = bases + (DefineCircuitKind,)
+        return type.__new__(metacls, name, bases, dct)
+
+    def __call__(*args, **kwargs):
```
cdonovick: `def __call__(cls, *args, **kwargs)`

tests/test_generator2.py
leonardt: maybe `_disable_cache_`?

magma/generator.py
```diff
@@ -55,3 +57,58 @@ def generate(*args, **kwargs):
...
+        cls = args[0]
+        args = args[1:]
+        if cls is Generator2:
+            try:
+                name, bases, dct = args
+            except ValueError:
+                raise Exception("Can not initialize base Generator class")
+            assert not kwargs
+            return type.__new__(cls, name, bases, dct)
+        cache = not getattr(cls, "_no_cache_", False)
```
cdonovick: seems a bit strange why not do `getattr(cls, "_cache_", True)`

tests/test_generator2.py
cdonovick: Why?

cdonovick: My comment was to Raj's preference.

rsetaluri: ehh no strong preference...will go with that

magma/generator.py
rsetaluri: agreed

PR #629, opened 2020-03-11
Add coroutine jtag test

leonardt:
Preliminary JTAG implementation.

There's still some improvements to be made but this adds what's needed for a functional compiler output.

improvements:
1) Merge JTAG yield output and yield state register (the FSM yields it's state as the output, the choice of each yield's encoding is important, allow the user to override the default encoding for yield IDs, need a good syntax for this).  This is important for optimization (shouldn't need two separate registers), but functionality this implementation works.
2) The compiler does not prune infeasible paths from the control flow graph, so if you write this the wrong way, it will break the compiler (e.g. infinite loop in the CFG even though it's infeasible).  CC @cdonovick given a list of conditions in AST form, e.g. `(tmp == 0) & (tms == 1)` is there a simple way to check satisfiability using pysmt and/or smtbitvector? I can collect the variable names in the expression pretty easily, so I could create a lambda's and pass in a symbolic variable for each value and return a constructed expression.  Then I just need to pass this to a solver.

### DIFF
--- a/magma/syntax/__init__.py
+++ b/magma/syntax/__init__.py
@@ -1,3 +1,3 @@
 from magma.syntax.combinational import combinational
 from magma.syntax.verilog import build_kratos_debug_info
-from .coroutine import coroutine, generator
+from .coroutine import coroutine
--- a/magma/syntax/__init__.py
+++ b/magma/syntax/__init__.py
@@ -1,3 +1,3 @@
 from magma.syntax.combinational import combinational
 from magma.syntax.verilog import build_kratos_debug_info
-from .coroutine import coroutine
+from .coroutine import coroutine, generator
--- a/magma/syntax/coroutine.py
+++ b/magma/syntax/coroutine.py
@@ -168,7 +168,7 @@ def _coroutine(defn_env, fn):
     if not manual_encoding:
         method_name_map["__init__"].body.append(ast.parse(
             f"self.yield_state: m.Bits[{yield_state_width}] = 0"
-    ).body[0])
+        ).body[0])
 
     # create new call body
     call_method.body = []
--- a/magma/syntax/coroutine.py
+++ b/magma/syntax/coroutine.py
@@ -127,24 +127,6 @@ def get_options(tree):
     return class_vars
 
 
-def get_manual_encoding(paths):
-    encoding = None
-    for path in paths:
-        for stmt in path:
-            print(ast.dump(stmt))
-            if isinstance(stmt, ast.Assign) and len(stmt.targets) == 0 \
-                    and isinstance(stmt.targets[0], ast.Attribute) and \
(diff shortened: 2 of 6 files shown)


leonardt:
Note on infeasible paths: it is worth checking if the downstream synthesis tool can optimize these cases out.  However, there's still the user facing issue of being able to send the compiler into an infinite loop.  We can early exit this by detecting possible cycles in the CFG, but we can't reason about whether they're legal or not.  So we can at least guard against programs that have these possible cycles (although maybe in the case of hardware checking satisfiability of conditions along paths isn't too expensive)

cdonovick:
If you know the types and values than it should not be hard to pass to an smt to solver.
If you don't know the types and values then its impossible.

for instance how do you plan to tell the difference between the following:
```Python
x = SMTBitVector[8](1)
if x > 1: # infeasible
    ...

x = SMTBitVector[8]()
if x > 1: #feasible
    ...
```
Now I don't know anyway to even get the types of variables let alone the values from an ast.
I guess you could try to exec it partially or something and extract information?

But assuming you could somehow do that. It would be pretty easy

Suppose I am trying to show `path = [cond_0, cond_1, ...]` is infeasible.  
Where `cond_i` is some `SMTBitVector` expression
I would do the following:
```Python
import pysmt.shortcuts as sc
import operator as op
constraints = reduce(op.And, path) # get the & reduction of  path
with sc.Solver(solver_name) as solver:
    solver.add_assertion(constraint.value) # .value gets the underlying pysmt expr
    if solver.solve():
         # path is feasible, the model will give you a trace
    else:
         # path is infeasible
```

leonardt:
Thanks, I'll try it out.  What is the type of path? Is it something like the return value of `SMTBitVector[8]() > 1`? (So just pass an SMTBitVector into a lambda). 

The silica compiler has a type inference algorithm it uses to get the types of all the variables in the program so we'll know that.  

While I don't think catching every possible feasible/infeasible path will be possible/scalable, there are plenty of simple cases, e.g. when a `for i in range(0, 8)` loop is desugared into `i = 0; while i <8; ...` we know we'll always enter the loop for the first iteration (so the short-circuit path around the loop is false).  Similarly with `while True` loops.  We can use some basic heuristics to test for these cases, but in general we probably won't be able to determine all infeasible paths and will have to be conservative.

cdonovick:
`path :: Iterable[Bit]`
Each element is a condition required to reach a code point.  So exactly like `SMTBitVector[8]() > 1`

```Python
if x < y:
    if x & z == 0:
        # path = [x < y, x & z == 0]
    else:
        # path = [x < y, ~(x & z == 0)]
else:
    # path =  [~(x < y)]
```
As long as you over approximate the reachable paths it shouldn't be a problem.

rsetaluri:
Design looks good to me...will defer on the details of coroutine stuff - but good to merge.

PR #633, opened 2020-03-12
Make sequential use __annotations__ to generate IO

cdonovick:


magma/syntax/sequential.py
```diff
@@ -348,15 +312,33 @@ def _sequential(
...
+                    outputs.append((f'O{idx}', t_name))
+                    defn_env[t_name] = t
+            else:
+                t_name = gen_free_name(call_def, defn_env, 'T')
+                outputs.append(('O', t_name))
+                defn_env[t_name] = v
+        else:
+            t_name = gen_free_name(call_def, defn_env, 'T')
+            defn_env[t_name] = v
+            inputs.append((k, t_name))
```
leonardt: Can we move this into a helper function just for organization?

cdonovick: Will do

PR #634, opened 2020-03-12
Init m.display

leonardt:
Adds basic support for display using inline_verilog logic.  Allows registering events that trigger the display and guarding with a single condition.


leonardt:
See the comments in https://github.com/phanrahan/magma/pull/638 and https://github.com/phanrahan/magma/pull/639 for some details on logging and file display/logging.  Merged both branches into here for a full PR

docs/display_and_logging.md
```diff
@@ -0,0 +1,197 @@
...
+m.info
+m.warning
+m.error
+```
+They the provide the same interface as `m.display`.  They will prepend an uppercase
+string of the form `[<LEVEL_NAME>] ` to the display statement, e.g. `[INFO]
+<display str>.`
+
+To control verbosity in the downstream tool, simply set the preprocessor
+variable `MAGMA_LOG_LEVEL` to the desired value.  Magma will only display log
```
rsetaluri: there's a little bit of overloading with internal logging and external logging (not conceptually, just in terms of names). not critical, but we should resolve that at some point

magma/circuit.py
```diff
@@ -62,13 +62,55 @@ class DefinitionContext:
...
         self.placer = placer
         self._inline_verilog = []
+        self._displays = []
+        self._insert_default_log_level = False
+        self._files = []
+
+    def add_file(self, file):
+        self._files.append(file)
+
+    def finalize_file_opens(self):
```
rsetaluri: let's make these `finalize` helpers private with prepended `_`

magma/circuit.py
```diff
@@ -62,13 +62,55 @@ class DefinitionContext:
...
+        for file in self._files:
+            self.add_inline_verilog(
+                f"""
+integer \_file_{file.filename} ;
```
rsetaluri: I've found that these long inline multiline strings make the code harder to read and scan, can we make the private globals at the top of the file? So like

```python
_logger = root_logger()

_VERILOG_FILE_CLOSE = """
integer \_file_{filename} ;
initial \_file_{filename} = $fopen(\"{filename}\", \"{mode}\");
"""

...

            self.add_inline_verilog(_VERILOG_FILE_CLOSE.format(filename=file.filename,
                                                               mode=file.mode)
                                   {}, {})
```

magma/circuit.py
```diff
@@ -62,13 +62,55 @@ class DefinitionContext:
...
     def add_inline_verilog(self, format_str, format_args, symbol_table):
         self._inline_verilog.append((format_str, format_args, symbol_table))
 
+    def insert_default_log_level(self):
+        self._insert_default_log_level = True
+
+    def add_display(self, display):
+        self._displays.append(display)
+
+    def finalize_displays(self):
```
rsetaluri: also private

magma/circuit.py
```diff
@@ -62,13 +62,55 @@ class DefinitionContext:
...
+        for display in self._displays:
+            self.add_inline_verilog(*display.get_inline_verilog())
+
     def finalize(self, defn):
         final_placer = self.placer.finalize(defn)
         final_context = DefinitionContext(final_placer)
+        if self._insert_default_log_level:
+            self.add_inline_verilog(f"""
```
rsetaluri: same re: multiline strings

magma/log.py
```diff
@@ -0,0 +1,52 @@
...
+
+class Warning(Log):
+    level = 2
+
+
+class Error(Log):
+    level = 3
+
+
+def make_log_func(T):
```
rsetaluri: private, as well as the above classes

magma/display.py
```diff
@@ -0,0 +1,142 @@
+from .circuit import _definition_context_stack
+from .t import Type
+
+
+class Event:
```
rsetaluri: we can make all these classes (Event, Posedge, Negedge, Display, File) private, and just have the functions be public

docs/display_and_logging.md
```diff
@@ -0,0 +1,197 @@
+# m.display
```
rsetaluri: great!

rsetaluri:
Looks good, some minor comments

docs/display_and_logging.md
leonardt: yea, for the user's convenience, perhaps we can rename our internal logger to be explicit (e.g. "internal_log" or something)

PR #640, opened 2020-03-16
Circuit builder

rsetaluri:


magma/circuit.py
```diff
@@ -112,22 +115,20 @@ def _finalize_displays(self):
...
+        self._finalize_file_close()  # close after displays
+        # Inline logic may introduce instances.
+        with _DefinitionContextManager(self):
             for format_str, format_args, symbol_table in self._inline_verilog:
                 process_inline_verilog(defn, format_str, format_args,
                                        symbol_table)
-        return final_context
+        for builder in self._builders:
+            inst = builder.finalize()
+            self.placer.place(inst)
```
leonardt: I wonder if this should go before the inline/display logic since they might introduce these constructs during the finalize?

magma/circuit.py
```diff
@@ -271,16 +266,20 @@ def __new__(metacls, name, bases, dct):
...
+            setattrs(cls, cls.interface.ports, lambda k, v: isinstance(k, str))
         try:
             context = _definition_context_stack.pop()
             assert context.placer.name == cls_name
-            cls._context_ = context.finalize(cls)
+            try:
+                context = dct["_context_"]
+            except KeyError:
+                pass
+            cls._context_ = context
```
leonardt: could simplify to `cls._context_ = dct.get("_context_", context)`

magma/circuit.py
```diff
@@ -830,3 +830,52 @@ def wrapped(*args, **kwargs):
 def DeclareCoreirCircuit(*args, **kwargs):
     return DeclareCircuit(*args, **kwargs,
                           renamed_ports=coreir_port_mapping)
+
+
+class _CircuitBuilderMeta(type):
+    pass
+
+
+def builder_method(func):
```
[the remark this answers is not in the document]
leonardt: seems reasonable

magma/circuit.py
```diff
@@ -830,3 +830,52 @@ def wrapped(*args, **kwargs):
...
+
+
+class CircuitBuilder(metaclass=_CircuitBuilderMeta):
+    def __init__(self, name):
+        try:
+            context = _definition_context_stack.peek()
+            context.add_builder(self)
+        except IndexError:  # builders must be inside a definition context
+            raise Exception("Can not instance a circuit builder outside a "
+                            "definition")
```
leonardt: Comment is redundant with exception string

magma/circuit.py
```diff
@@ -830,3 +830,52 @@ def wrapped(*args, **kwargs):
...
+
+    def finalize(self):
+        if self._finalized:
+            raise Exception("Can only call finalize on a CircuitBuilder once")
+        self._finalize()
+        bases = (AnonymousCircuitType,)
+        dct = {"io": self._io, "_context_": self._context, "name": self._name}
+        DefineCircuitKind.__prepare__(self._name, bases)
+        t = DefineCircuitKind(self._name, bases, dct)
+        return t(name=self._name)
```
leonardt: Should this set _finalized = True? (Or should self._finalize do it?)

magma/circuit.py
```diff
@@ -830,3 +830,52 @@ def wrapped(*args, **kwargs):
...
+
+    def _finalize(self):
+        pass
+
+    def finalize(self):
+        if self._finalized:
+            raise Exception("Can only call finalize on a CircuitBuilder once")
+        self._finalize()
+        bases = (AnonymousCircuitType,)
+        dct = {"io": self._io, "_context_": self._context, "name": self._name}
```
leonardt: I wonder if we want to allow the builder to passthrough other names into the `dct` context?  Magma relies on a a lot of special names in the class namespace for certain features.  Alternatively, we should consider a different API for these features.

magma/interface.py
```diff
@@ -53,24 +49,28 @@ def _parse(decl):
...
 
-def _make_interface_args(decl, renamed_ports, inst, defn):
-    names, ports = _parse(decl)  # parse the class Interface declaration
-    args = OrderedDict()
-    for name, port in zip(names, ports):
-        if   inst: ref = InstRef(inst, name)
-        elif defn: ref = DefnRef(defn, name)
-        else:      ref = NamedRef(name)
+def _make_ref(name, inst, defn):
+    assert not (inst is not None and defn is not None)
```
leonardt: could simplify this to be `inst is not None or defn is not None`

leonardt: never mind, misread the condition

regfile.py
```diff
@@ -0,0 +1,181 @@
+from magma import Generator, Circuit, CircuitBuilder, In, Out, IO, Bits, Clock, AsyncResetIn, ClockIO, wire, Product, InstRef
```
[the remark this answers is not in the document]
leonardt: Can we move this to a test file?

tests/test_interface/test_interface.py
```diff
@@ -23,7 +23,7 @@ def _io_similar(lhs, rhs):
 
 
 def test_1():
-    I0 = m.DeclareInterface("a", m.In(m.Bit), "b", m.Out(m.Bits[2]))
```
leonardt: Should we keep the old function and deprecate it? Just in case there's external code using it

tests/test_circuit_builder.py
```diff
@@ -0,0 +1,50 @@
...
+        io = m.IO(x=m.In(m.Bit), y=m.Out(m.Bit))
+
+        builder = _PassThroughBuilder(name="my_pt")
+        builder.add()
+
+        builder.I @= io.x
+        io.y @= builder.O
+
+    print (repr(_Top))
+    m.compile("build/builder_top", _Top, output="coreir")
```
leonardt: I'm assuming you've visually inspected the output, but maybe we should add a regression check here to make sure it doesn't break in the future

regfile.py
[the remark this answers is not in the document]
rsetaluri: I think the regfile should be part of mantle. I will implement it in mantle and add a test there once this is merged. (just included it here to show how we could implement it)

magma/circuit.py
rsetaluri: Done.

magma/circuit.py
rsetaluri: Done.

magma/circuit.py
rsetaluri: Forgot to include that. Done.

magma/circuit.py
rsetaluri: Good idea. Added a hook to do that called `set_namespace_key`

tests/test_interface/test_interface.py
rsetaluri: Don't think anyone is using it but added it back.

tests/test_circuit_builder.py
rsetaluri: Done.

magma/circuit.py
rsetaluri: Done.

PR #642, opened 2020-03-17
Enable backwards compatability for undriven/unused logic

leonardt:
The old-style `DeclareCircuit` does not populate the `instances` attribute.  This updates the passes to check `isdefintion` before referencing the attribute for backwards compatibility.

### DIFF
--- a/magma/passes/drive_undriven.py
+++ b/magma/passes/drive_undriven.py
@@ -1,4 +1,5 @@
 from .passes import EditCircuitPass
+from ..is_definition import isdefinition
 
 
 def _drive_undriven(interface):
@@ -14,5 +15,6 @@ class DriveUndrivenPass(EditCircuitPass):
     def edit(self, circuit):
         if _drive_undriven(circuit.interface):
             circuit._is_definition = True
-        for inst in circuit.instances:
-            _drive_undriven(inst.interface)
+        if isdefinition(circuit):
+            for inst in circuit.instances:
+                _drive_undriven(inst.interface)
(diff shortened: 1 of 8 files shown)


magma/passes/terminate_unused.py
```diff
@@ -1,4 +1,5 @@
 from .passes import EditCircuitPass
+from ..is_definition import isdefinition
```
rsetaluri: unrelated, but realized line 8 should be

```python
        if port.is_output() and not port.wired():
```

magma/passes/terminate_unused.py
```diff
@@ -14,5 +15,6 @@ class TerminateUnusedPass(EditCircuitPass):
     def edit(self, circuit):
         if _terminate_unused(circuit.interface):
             circuit._is_definition = True
-        for inst in circuit.instances:
-            _terminate_unused(inst.interface)
+        if isdefinition(circuit):
+            for inst in circuit.instances:
+                _terminate_unused(inst.interface)
```
rsetaluri:
```suggestion
        if not isdefinition(circuit):
            return
        for inst in circuit.instances:
            _terminate_unused(inst.interface)
```

magma/passes/drive_undriven.py
```diff
@@ -14,5 +15,6 @@ class DriveUndrivenPass(EditCircuitPass):
     def edit(self, circuit):
         if _drive_undriven(circuit.interface):
             circuit._is_definition = True
-        for inst in circuit.instances:
-            _drive_undriven(inst.interface)
+        if isdefinition(circuit):
+            for inst in circuit.instances:
+                _drive_undriven(inst.interface)
```
rsetaluri:
```suggestion
        if not isdefinition(circuit):
            return
        for inst in circuit.instances:
            _drive_undriven(inst.interface)
```

PR #647, opened 2020-03-18
Bug646

leonardt:
Add support for seq/comb phi for magma protocol.  This changes the code to use magma's new internal `Mux` primitive (rather than Mantle), so this way the fix for magma/mantle is the same.

The main change is
```
        if issubclass(T, MagmaProtocol):
            T = T._to_magma_()
```
inside the Mux primitive, but there were some other fixes needed to integrate into the entire code path (including updating the golds because we changed which generator we're using for the phi in the tests).


magma/backend/coreir_transformer.py
```diff
@@ -201,7 +201,7 @@ def connect(self, module_defn, port, value, non_input_ports):
         if value is None:
             if port.is_inout():
                 return  # skip inouts because they might be conn. as an input.
-            if self.defn._ignore_undriven_:
+            if getattr(self.defn, "_ignore_undriven_", False):
```
leonardt: Some circuits don't define this, so we should use `getattr`

magma/backend/coreir_utils.py
```diff
@@ -138,7 +138,7 @@ def is_clock_or_nested_clock(p):
     if issubclass(p, Array):
         return is_clock_or_nested_clock(p.T)
     if issubclass(p, Tuple):
-        for item in p.Ts:
+        for item in p.types():
```
leonardt: This is an old `hwtypes` branch related interface change that wasn't caught until now (this is an error code path, so doesn't show up in normal function)

magma/operators.py
```diff
@@ -24,23 +26,42 @@ def __init__(self, N: int, width: int):
 
 class Mux(Generator2):
     def __init__(self, height: int, T: Type):
+        T_str = str(T).replace("(", "").replace(")", "")\
+                      .replace(",", "_").replace("=", "_")\
+                      .replace("[", "").replace("]", "").replace(" ", "")
```
leonardt: For backwards compatibility with golds (same name as mantle mux)

magma/operators.py
```diff
@@ -24,23 +26,42 @@ def __init__(self, N: int, width: int):
...
         N = T.flat_length()
 
         io_dict = {}
         for i in range(height):
             io_dict[f"I{i}"] = In(T)
-        io_dict["S"] = In(Bits[clog2(height)])
+        if height == 2:
+            io_dict["S"] = In(Bit)
+        else:
+            io_dict["S"] = In(Bits[clog2(height)])
```
leonardt: Mantle uses `Bit` for select of two items, where coreir uses `Bits[1]`

magma/syntax/combinational.py
```diff
@@ -218,12 +219,10 @@ def _combinational(defn_env: dict, fn: types.FunctionType):
...
         combinational, tree.decorator_list, defn_env)
+    if "m" not in defn_env:
+        defn_env["m"] = m
     if "phi" not in defn_env:
-        tree = ast.Module([
-            ast.parse("import magma as m").body[0],
-            ast.parse("from mantle import mux as phi").body[0],
-            tree
-        ])
+        defn_env["phi"] = m.operators.mux
```
leonardt: Use env injection instead of inserting the imports into the tree

magma/operators.py
cdonovick: Relaying on this seems bad but ¯\\\_(ツ)\_/¯

PR #649, opened 2020-03-18
Try to maintain type names

cdonovick:
Try to maintain type names in sequential.


magma/syntax/sequential.py
```diff
@@ -237,31 +237,48 @@ def gen_register_instances(initial_value_map, async_reset, magma_name):
...
+            return k
+    else:
+        return None
+
+
 def get_io_from_annotations(f, tree, defn_env):
     inputs = []
     outputs = []
     tuple_out = False
+    type_map = {}
```
cdonovick: A cache of t -> name

PR #652, opened 2020-03-20
Reorg cleanup

rsetaluri:


magma/bit.py
```diff
@@ -179,20 +179,8 @@ def __int__(self) -> int:
         raise NotImplementedError("Converting magma bit to int not supported")
 
     def __repr__(self):
-        if self is VCC:
-            return "VCC"
-        if self is GND:
-            return "GND"
         return super().__repr__()
```
leonardt: Can just remove repr entirely

magma/bit.py
```diff
@@ -179,20 +179,8 @@ def __int__(self) -> int:
...
 
-
-VCC = Bit[Direction.Out](name="VCC")
-GND = Bit[Direction.Out](name="GND")
-
-HIGH = VCC
-LOW = GND
-
-
```
leonardt: Should we keep HIGH/LOW defined in this namespace for backwards compatibility? Probably not a major issue though

magma/circuit.py
```diff
@@ -151,13 +147,10 @@ def _has_definition(cls, port=None):
...
-            return None
+            return False
         return any(_has_definition(cls, p) for p in interface.ports.values())
-    if isinstance(port, Tuple) or isinstance(port, Array):
-        return any(_has_definition(cls, elem) for elem in port)
-    if not port.is_output():
-        return port.value() is not None
-    return False
+    flat = port.flatten()
+    return any([not f.is_output() and f.value() is not None for f in flat])
```
leonardt: Don't need the extra list comp before any I think `[...]`.  Ideally `flatten()` returns a generator object so we don't have to fully expand this in a lot of cases.

magma/circuit.py
```diff
@@ -218,17 +211,18 @@ def _get_intermediate_values(value, values):
...
+        for f in flat:
+            _get_intermediate_values(f, values)
+        return
+    while driver is not None:
+        _add_intermediate_value(driver, values)
+        if not driver.is_output():
+            value = driver
+            driver = driver.value()
+        else:
+            driver = None
```
leonardt: Maybe could simplify to 
```
if driver.is_output():
    return
value = driver
driver = driver.value()
```

magma/conversions.py
```diff
@@ -260,3 +260,112 @@ def replace(value, others: dict):
...
+        child_length = child.flat_length()
+        children.append(
+            from_bits(child, value[offset:offset + child_length])
+        )
+        offset += child_length
+    return cls(*children)
+
+
+def from_bits(cls, value):
+    return _dispatch(cls, _from_bits, cls, value)
```
leonardt: I think we should expose (and document) `_from_bits` and `_as_bits` so users can define their own types that implement this interface (that was the benefit of having it in the type definition themselves).  While moving them into a separate conversions module for organization makes sense, I think we should still allow users to define custom implementations of this logic.

magma/digital.py
```diff
@@ -2,16 +2,103 @@
...
+            return None
+        return self.driver.bit
+
+    def driven(self):
+        return self.trace() is not None
+
+    def wired(self):
+        return self.driver or self.driving
+
+
```
leonardt: Eventually we'd like to have Wire work with more complex types (e.g. when wiring two arrays together, we could connect corresponding Wire objects, rather than connecting the children's Wire's), so anticipating this, maybe we'd like Wire to be in a separate module?

magma/digital.py
```diff
@@ -194,28 +283,24 @@ def getgpio(self):
...
-        result = m.Bits[1]()
-        result[0] @= self
-        return result
-
-    @classmethod
-    def from_bits(cls, other):
-        if not isinstance(other, m.Bits) and not len(other) == 1:
-            raise TypeError("Can only convert from Bits[1] to Bit")
-        return other[0]
+        m.wire(DefineUndriven()().O, m.bit(self))
```
leonardt: Perhaps if we made `DefineUndriven` and `DefineUnused` parametrized by `type`, we could remove this circular dependency (on `m.bit`), also we could use `@=` to avoid `m.wire`?

magma/bit.py
rsetaluri: Done

magma/bit.py
rsetaluri: Doen't see either of these used outside of digital.py...will keep there for now for backward compat. but can be removed.

magma/circuit.py
rsetaluri: Done. Will make `flatten` return a generator in a future optimization

magma/circuit.py
rsetaluri: Done.

magma/conversions.py
rsetaluri: I agree. I'm actually wondering why we shouldn't just use `flatten` and something like `reshape`? Or even better a generic `reinterpret` function?

IMO, right now there's many ways to do the same thing of converting one type to another in magma. We should just decide on one way and make that the standard. For now I'll leave this as is, but I agree overriding this behavior should be possible.

magma/digital.py
rsetaluri: Agreed...wasn't sure what to name the file since there's a `wire.py` already. I'll figure out a reorg in a later commit.

magma/digital.py
rsetaluri: Agreed. Will work on this in the next batch of commits.

rsetaluri:
Thanks @leonardt for the comments. I want to make some more refactoring changes, so I'll continue to address your outstanding issues in those.

PR #658, opened 2020-03-26
Clean up type dependencies

rsetaluri:
Several changes to import structure and who own's what: all towards the effort of cleaning up package imports and circular dependencies.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -1,21 +1,15 @@
 import weakref
 from functools import reduce
 from abc import ABCMeta
-from hwtypes import BitVector
 from .common import deprecated
 from .ref import AnonRef, ArrayRef
 from .t import Type, Kind, Direction, In, Out
 from .compatibility import IntegerTypes
 from .digital import Digital
 from .bit import VCC, GND, Bit
-from .bitutils import int2seq, seq2int, clog2
+from .bitutils import int2seq, seq2int
 from .debug import debug_wire, get_callee_frame_info
 from .logging import root_logger
-from magma.generator import Generator2
-from magma.protocol_type import MagmaProtocol
-from magma.interface import IO
-from magma.tuple import Product
-from magma.circuit import coreir_port_mapping
 
 
 _logger = root_logger()
@@ -280,9 +274,6 @@ class Array(Type, metaclass=ArrayMeta):
     def flat_length(cls):
         return cls.N * cls.T.flat_length()
 
-    def dynamic_mux_select(self, key):
-        return Mux(len(self), self.T)()(*self.ts, key)
-
     def __getitem__(self, key):
         if isinstance(key, Type):
             # indexed using a dynamic magma value, generate mux circuit
@@ -461,56 +452,3 @@ class Array(Type, metaclass=ArrayMeta):
 
 
 ArrayType = Array
-
-
-class CoreIRCommonLibMuxN(Generator2):
-    def __init__(self, N: int, width: int):
-        self.name = f"coreir_commonlib_mux{N}x{width}"
-        MuxInT = Product.from_fields("anon", dict(data=Array[N, Array[width, Bit]],
-                                                  sel=Array[clog2(N), Bit]))
-        self.io = IO(I=In(MuxInT), O=Out(Array[width, Bit]))
-        self.renamed_ports = coreir_port_mapping
-        self.coreir_name = "muxn"
-        self.coreir_lib = "commonlib"
-        self.coreir_genargs = {"width": width, "N": N}
-        self.primitive = True
-        self.stateful = False
-
-        def simulate(self, value_store, state_store):
-            sel = BitVector[clog2(N)](value_store.get_value(self.I.sel))
-            out = BitVector[width](value_store.get_value(self.I.data[int(sel)]))
-            value_store.set_value(self.O, out)
-
-        self.simulate = simulate
-
-
-class Mux(Generator2):
-    def __init__(self, height: int, T: Type):
-        if issubclass(T, MagmaProtocol):
-            T = T._to_magma_()
-        T_str = str(T).replace("(", "").replace(")", "")\
-                      .replace(",", "_").replace("=", "_")\
-                      .replace("[", "").replace("]", "").replace(" ", "")
-        self.name = f"Mux{height}x{T_str}"
-        # TODO: Type must be hashable so we can cache.
-        N = T.flat_length()
-
-        io_dict = {}
-        for i in range(height):
-            io_dict[f"I{i}"] = In(T)
-        if height == 2:
-            io_dict["S"] = In(Bit)
-        else:
-            io_dict["S"] = In(Array[clog2(height), Bit])
-        io_dict["O"] = Out(T)
-
-        self.io = IO(**io_dict)
-        mux = CoreIRCommonLibMuxN(height, N)()
-        data = [Array[N, Bit](getattr(self.io, f"I{i}").flatten())
-                for i in range(height)]
-        mux.I.data @= Array[height, Array[N, Bit]](data)
-        if height == 2:
-            mux.I.sel[0] @= self.io.S
-        else:
-            mux.I.sel @= self.io.S
-        self.io.O @= T.unflatten(mux.O)
--- a/magma/array.py
+++ b/magma/array.py
@@ -1,16 +1,21 @@
 import weakref
 from functools import reduce
 from abc import ABCMeta
-import magma as m
+from hwtypes import BitVector
 from .common import deprecated
 from .ref import AnonRef, ArrayRef
-from .t import Type, Kind, Direction
+from .t import Type, Kind, Direction, In, Out
 from .compatibility import IntegerTypes
 from .digital import Digital
 from .bit import VCC, GND, Bit
-from .bitutils import seq2int
+from .bitutils import int2seq, seq2int, clog2
 from .debug import debug_wire, get_callee_frame_info
 from .logging import root_logger
+from magma.generator import Generator2
+from magma.protocol_type import MagmaProtocol
+from magma.interface import IO
+from magma.tuple import Product
+from magma.circuit import coreir_port_mapping
 
 
 _logger = root_logger()
@@ -205,12 +210,12 @@ class Array(Type, metaclass=ArrayMeta):
                         raise TypeError(f"Can only instantiate Array[N, Bit] "
                                         "with int, not Array[N, {self.T}]")
                     self.ts = []
-                    for bit in m.bitutils.int2seq(args[0], self.N):
+                    for bit in int2seq(args[0], self.N):
                         self.ts.append(VCC if bit else GND)
                 elif self.N == 1:
                     t = args[0]
                     if isinstance(t, IntegerTypes):
-                        t = m.VCC if t else m.GND
+                        t = VCC if t else GND
                     assert type(t) == self.T or type(t) == self.T.flip() or \
                         issubclass(type(type(t)), type(self.T)) or \
                         issubclass(type(self.T), type(type(t))), (type(t), self.T)
@@ -237,6 +242,10 @@ class Array(Type, metaclass=ArrayMeta):
         if cls.T is None:
             return False
         return cls.T.is_oriented(direction)
+
+    @classmethod
+    def is_clock(cls):
+        return False
 
     @classmethod
     @deprecated
@@ -272,7 +281,7 @@ class Array(Type, metaclass=ArrayMeta):
         return cls.N * cls.T.flat_length()
 
     def dynamic_mux_select(self, key):
-        return m.operators.Mux(len(self), self.T)()(*self.ts, key)
+        return Mux(len(self), self.T)()(*self.ts, key)
 
     def __getitem__(self, key):
         if isinstance(key, Type):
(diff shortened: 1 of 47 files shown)


rsetaluri:
Not sure why build's aren't showing up here, but here is the latest build: https://travis-ci.org/github/phanrahan/magma/builds/667380133

magma/interface.py
```diff
@@ -195,7 +188,7 @@ def __getitem__(self, key):
         if isinstance(key, int):
             return self.arguments()[key]
         if isinstance(key, slice):
-            return array([self[i] for i in range(*key.indices(len(self)))])
+            return [self[i] for i in range(*key.indices(len(self)))]
```
cdonovick: this seems like a pretty big change in semantics,

magma/array.py
```diff
@@ -388,14 +398,17 @@ def iswhole(self, ts):
 
         return True
 
+    def iswhole(self):
+        return Array._iswhole(self.ts)
```
cdonovick: Why this pattern? (Internally using `SpecificClass._static_method`)
If you are going to do this its seems better to minimally use `type(self)._static_method`

rsetaluri:
Thanks for taking a look @cdonovick. I actually did break them out into some semantically different changes (the individual commits are also self contained).

See #653, #654, #655, #656, #657. I just wanted to get them all in before someone overwrote changes. Part of the reason why I want to break up files more so we can do more individual development.

magma/array.py
```diff
@@ -272,7 +281,7 @@ def flat_length(cls):
         return cls.N * cls.T.flat_length()
 
     def dynamic_mux_select(self, key):
-        return m.operators.Mux(len(self), self.T)()(*self.ts, key)
+        return Mux(len(self), self.T)()(*self.ts, key)
```
leonardt: I don't see this Mux operator imported above, perhaps we don't have a test code path covering this?

magma/array.py
leonardt: This seems like a reasonable change to preserve backwards compatibility (the old `iswhole` method is still available, but from now on we should be using the _iswhole static method).

magma/array.py
leonardt: Ah, it's defined below.

magma/array.py
```diff
@@ -438,5 +460,57 @@ def is_mixed(cls):
...
+        self.io = IO(**io_dict)
+        mux = CoreIRCommonLibMuxN(height, N)()
+        data = [Array[N, Bit](getattr(self.io, f"I{i}").flatten())
+                for i in range(height)]
+        mux.I.data @= Array[height, Array[N, Bit]](data)
+        if height == 2:
+            mux.I.sel[0] @= self.io.S
+        else:
+            mux.I.sel @= self.io.S
+        self.io.O @= T.unflatten(mux.O)
```
leonardt: This seems like an odd place to have this, perhaps we can have a primitive directory and have a Mux file imported here?

magma/array.py
leonardt: Ah, scratch that, I see that the is whole signature has changed to not accept any parameters,  I think this is okay, the static method could also be a private helper but a static method is reasonable to me.

magma/bfloat.py
```diff
@@ -61,11 +63,11 @@ def declare_ite(cls, T):
...
-                                              coreir_genargs=coreir_genargs,
-                                              coreir_lib="float")
+        return DeclareCoreirCircuit(f"magma_BFloat_{N}_ite_{t_str}",
+                                    "I0", In(T),
+                                    "I1", In(T),
+                                    "S", In(Bit),
+                                    "O", Out(T),
+                                    coreir_name="mux",
+                                    coreir_genargs=coreir_genargs,
+                                    coreir_lib="float")
```
leonardt: I think Declare will generate a deprecation warning, so at some point we should update this to the subclass pattern.

magma/bits.py
```diff
@@ -505,23 +505,28 @@ def __lt__(self, other):
 
     @classmethod
     def get_family(cls):
-        return m.get_family()
+        return get_family()
+
+    @classmethod
+    def unflatten(cls, value):
+        # TODO(rsetaluri): Should we type check here?
+        return Bits[len(value)](value)
```
leonardt: I think it'd be fine to assume that the constructor type checks, no need to do it in two places.

magma/bits.py
```diff
@@ -532,7 +537,7 @@ def simulate(self, value_store, state_store):
 
     @lru_cache(maxsize=None)
     def Define(width):
-        return m.circuit.DeclareCoreirCircuit(
+        return DeclareCoreirCircuit(
```
leonardt: Another deprecation warning point

magma/common.py
```diff
@@ -17,6 +18,40 @@ def peek(self):
...
+    def add(self, elem):
+        self.refs.add(_Ref(elem))
+
+    def discard(self, elem):
+        self.refs.discard(_Ref(elem))
+
+    def __repr__(self):
+        return "%s(%s)" % (type(self).__name__, list(self))
+
+
```
leonardt: Are Python sets deterministically ordered? Otherwise this might mess up the the `__repr__` hashing logic if the ordered of the generated intermediates are different.

leonardt: (e.g. like dictionaries being insertion ordered).

magma/interface.py
leonardt: Hmm, intuitively I would think that slicing an array would return an array

leonardt: To avoid the circular dependency, what about `type(self)(...)`? This seems more extensible and proper than the old implementation too.

leonardt: Actually, I guess you would need to do `type(self).unsized_t[len(slice)](...)` (assuming we have an unsized_t, if not we can add it)

magma/protocol_type.py
```diff
@@ -67,5 +70,10 @@ def flatten(self):
...
         return self._get_magma_value_().name
 
+    def wire(self, other):
+        if isinstance(other, MagmaProtocol):
+            other = other._get_magma_value_()
+        self._get_magma_value_().wire(other)
+
     def __imatmul__(self, other):
-        m.wire(other, self._get_magma_value_())
+        self.wire(other)
```
leonardt: Since it's `wire(o, i)` and in the case of imatmul, it's `i @= o`, I think this should be `other.wire(self)`

leonardt: Scratch that, that's the signature of wire, but then internally wire calls `o.wire(i)`

magma/t.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, name=None):
             name = AnonRef()
         elif isinstance(name, str):
             name = NamedRef(name=name)
+            name.value = self
```
leonardt: Should just use the constructor interface here?

leonardt:
This definitely warrants some careful integration testing because there's a lot of changes that touch interfaces

magma/interface.py
rsetaluri: this feature allows slicing on an interface* to return a magma `Array` rather than a python array, which could be a handy feature but seems pretty arbitrary as far as the external interface (whether or not it's the caller's responsibility to wrap the returned value in an `Array`).

TBH, this was one of the points I was least sure about since it changes the external interface. However supporting this interface *absolutely* injects a circular dependency since Array -> Circuit -> Interface -> Array. If we must preserve this interface, we can and do a package import. But also note there was only once place in all the repo where this was used (`tests/test_wire/test_arg.py `).

magma/array.py
rsetaluri: Yeah, it was kind of a broken interface anyway and not really used outside, and self-contained.

magma/array.py
rsetaluri: Yes, ideally would have liked to have `mux.py` or a separate file, but unfortunately `Mux` class depends on `Array` and `Array` depends on `Mux` class through `dynamic_mux_select` function. Fixes the dependency issue to have them in the same file, but I agree it's bad organization.

The fundamental issue is that there are operators we want on the types that involve "higher level" things, e.g. the `__getitem__` on `Array` invoking mux. The two alternatives I see are, (a) allowing the operators to inject functions - I think this was how things were done before ala monkey patching: `Array.__getitem__ = fn` in `mux.py`; and (b) only *declaring* those functions at operator time so like

```python
def dynamic_mux_select(self, ...):
    Select = DeclareCircuit('Select', ...)
    select = Select()
    select.I @= ...
    return select.O
```

magma/common.py
cdonovick: Sets are still unordered.
```Python
s0 = set()
# Do a bunch of insertions / deletions from s0
s1 = set(s0) 

assert s0 == s1
# iteration order of a specific set that has not been mutated should be consistent 
# (pretty sure this is not just CPython implementation detail) 
assert list(s0) == list(s0) 
# However this may or may not hold
assert list(s0) == list(s1)
```

magma/bfloat.py
rsetaluri: Definitely; will put up a follow up PR to get rid of more of these.

magma/bits.py
rsetaluri: The issue is if we do `Bits[8].unflatten([x, y, z])`, i.e. size-mismatch. Right now it will indiscriminately return `Bits[3]()`. This doesn't seem to bad so just added a logger warning and got rid of TODO.

magma/bits.py
rsetaluri: Definitely; will put up a follow up PR to get rid of more of these.

magma/array.py
leonardt: Linking is probably the right long term solution (declare interface, get implementation later).  Monkey patching is probably a reasonable short-term solution since it's low code overhead and I think I'd prefer that over having the mux operator inside the array file (since mux should work with Bit/Tuple as well, so it seems odd to have it in array)

magma/common.py
leonardt: I think we'll need an ordered set then so we can preserve the uniqueness __repr__ invariant.

magma/interface.py
leonardt: Oh sorry, I misread, I thought this was slice of `Array`, yea I'm not sure about this.  Let me consider it, but I don't feel to strongly about either way.

magma/common.py
cdonovick: I should point out that even `dict(...).keys()` / `OrderedDict(...).keys()` do not guarantee iteration for keys generated in different orders, they just guarantee that the same sequence of events will always generate the same order. 

So if you want `__repr__` to be invariant between runs of the same program an insertion ordered structure will do that.  If you want `__repr__` to be invariant between equivalent sets you will need to sort them or something.

cdonovick: I know https://pypi.org/project/sortedcontainers/ is popular.

leonardt: Hmm, I think the issue is that two circuit definitions generating the same circuit should be identical, the question is whether generating the set of temporaries will produce the same string (this will be separate code paths generating a set, then comparing the generated strings), so whether those code paths construct the set in the same order I think should be true? So if the same sequence of events generates the same set iteration order, that should be fine.  I don't know if we want to generally sort the temporaries (e.g. circuits defined with the same structure but in different orderings could be viewed as unique or not).  At the end of the day, this is conservative so should not be a problem, but I think if two identical circuits (e.g. an uncached generator) are found, their temporary value sets should iterate the same way so it produces the same repr string for uniqueness checking

rsetaluri: Good point. Updated to use an ordered version of identity set.

cdonovick: An inefficient by quick hack to get an insertion ordered set would be to just wrap a dict.  E.g.:
```
class OSet(MutableSet):
    def __init__(self, init=()):
        self._s = {}
        for elem in init:
            self.add(elem)

    def add(self, elem):
        self._s[elem] = None

    def remove(self, elem):
        del self._s[elem]

    ...
```

magma/common.py
```diff
@@ -46,12 +46,40 @@ def add(self, elem):
...
 
 
+class IdentitySet(_IdentitySetBase):
+    def __init__(self, items=[]):
+        refs = set(map(_Ref, items))
+        super().__init__(refs)
+
+
+class OrderedIdentitySet(_IdentitySetBase):
+    def __init__(self, items=[]):
```
cdonovick: Don't know why my previous comment didn't link to this code

magma/common.py
rsetaluri: 1. It is necessary for iteration order to be deterministic across runs of the same program/instances of the same circuit. It is good for the keys to be further sorted, but as Lenny pointed out that is not necessary and w/o that guarantee we just have a conservative check. Therefore, all we need is a data-structure whose iteration order is insertion order.

2. @cdonovick: I basically wrapped an ordered dict in `OrderedIdentitySet`. Let me know if the way I did it is a good solution.

3. Will upgrade to `items=()` over `items=[]`.

magma/t.py
rsetaluri: Good catch. Also using a weakref to avoid circular reference

magma/array.py
rsetaluri: Moved to monkey patching, seems reasonable for now.

rsetaluri:
@leonardt agree re: integration tests. Kicked off a garnet flow test: https://github.com/StanfordAHA/aha/pull/420

rsetaluri:
See https://github.com/StanfordAHA/aha/pull/420 passing!

magma/interface.py
rsetaluri: @leonardt any final thoughts on this? Don't feel strongly, but given that the flow passed (see below), it seems safe to break this interface, and I think the new interface still makes sense.

PR #659, opened 2020-03-31
Copy locals for inline_verilog inside loops

leonardt:
Fixes a bug where locals are modified after inline_verilog is called (e.g. inside a loop) by copying the frames before storing it (since we are staging the logic until later).  Before this change, the resulting verilog would use the same (final) value of the loop variable for each inline string.

Depends on https://github.com/leonardt/ast_tools/pull/45

Also modifies the has_definition check to handle the case when there's only inline_verilog (used in the test)

### DIFF
--- a/magma/inline_verilog.py
+++ b/magma/inline_verilog.py
@@ -1,3 +1,4 @@
+import copy
 import string
 from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT, get_symbol_table
 from magma.circuit import _definition_context_stack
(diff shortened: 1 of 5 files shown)


magma/inline_verilog.py
```diff
@@ -1,3 +1,4 @@
+import copy
```
rsetaluri, line 1: looks unused?

leonardt:
ast_tools PR hasn't been merged yet, that's why

rsetaluri:
approved and merged ast_tools PR

PR #660, opened 2020-04-01
Coroutine sdram

leonardt:
Adds initial implementation of the coroutine SDRAM.

Future improvement will add support for desugaring for loops into while loops (will do this as a separate/generic AST tools pass)

### DIFF
--- a/magma/syntax/transforms/inline_yield_from.py
+++ b/magma/syntax/transforms/inline_yield_from.py
@@ -72,10 +72,7 @@ class YieldFromFunctionInliner(ast.NodeTransformer):
                     # TODO: Should use an ast_tools macro symbol replacer
                     # instead
                     tree = tree.body[0]
-            else:
-                assert len(tree.args) == 0, \
-                    "assumes no args for non-make variant"
-            # TODO: Assumes return is well formed
+            # TODO: Assumes function signature and return is well formed
             return tree.body[:-1]
         else:
             return node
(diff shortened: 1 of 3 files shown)


tests/test_syntax/test_coroutine/test_sdram.py
```diff
@@ -0,0 +1,285 @@
...
+    WRIT_CAS = m.bits(0b11010, 5)
+    WRIT_NOP2 = m.bits(0b11011, 5)
+
+    # CMD_PALL = "8'b10010001"
+    # CMD_REF = "8'b10001000"
+    # CMD_NOP = "8'b10111000"
+    # CMD_MRS = "8'b1000000x"
+    # CMD_BACT = "8'b10011xxx"
+    # CMD_READ = "8'b10101xx1"
+    # CMD_WRIT = "8'b10100xx1"
```
leonardt, line 87: Keeping these as reference.  The original verilog has `x` for some of the bits, in magma we don't support this yet, so we just make them 0 for now.

magma/syntax/transforms/inline_yield_from.py
```diff
@@ -61,14 +61,19 @@ def visit_YieldFrom(self, node):
...
-            symbol_table = {}
-            # Skip self arg
-            for arg, param in zip(func.args, tree.args.args[1:]):
-                symbol_table[param.arg] = arg
-            tree = replace_symbols(tree, symbol_table)
-            # TODO: Assumes make function is well formed
-            # TODO: Assumes return is well formed
-            return tree.body[0].body[:-1]
+            if (isinstance(tree, ast.FunctionDef) and
+                    tree.name[0:5] == "make_"):
```
rsetaluri, line 65: nit: indentation looks off

leonardt: Style checker forced this indentation (originally had it aligned with the paren, but it complained about same as next logical line and the next line was over indented)

Issue #651, opened 2020-03-20
Wiring different subclasses of Digital is allowed


rsetaluri:
the following is allowed. is that expected?

```python
import magma as m


class Foo(m.Circuit):
    io = m.IO(I=m.In(m.Clock), O=m.Out(m.Reset))
    m.wire(io.I, io.O)
```

Came across this when trying to clean up `magma/digital.py`. It seems there are references to it's subclasses, which introduces a circ. dependency:

https://github.com/phanrahan/magma/blob/4605be27ce38413ee192c24a7d8a1c779e99bb06/magma/digital.py#L205

https://github.com/phanrahan/magma/blob/4605be27ce38413ee192c24a7d8a1c779e99bb06/magma/digital.py#L80

...among others.

Even if the first snipped is expected to be allowed, I'd like to remove the circular dependency if possible

leonardt:
This is a bug, we should fix it.  the circular dependency is for backwards compatibility (VCC and GND should be of type Digital, but there's existing code that relies on them being Bit, so we can remove that once we fix existing internal code, I suspect that there's not much user code that relies on this though so that's not as much of a worry)

rsetaluri:
Got it. Taking a look at this code so I will try to kill both birds with one stone.

Can you point to code that depends on VCC/GND being a Bit?

leonardt:
Related discussion: https://github.com/phanrahan/magma/issues/499, revert PR: https://github.com/phanrahan/magma/pull/536

The old behavior of `Bit(0)` and `Bit(1)` was to return VCC and GND.  Here's one place that it's used: https://github.com/phanrahan/magma/blob/master/magma/backend/coreir_transformer.py#L221

Basically, there's code that relies on `is VCC` or `is GND` to check constants, which should be changed to check it another way.

I think changing VCC/GND to Digital makes sense semantically, and having `Bit(0)` and `Bit(1)` be different (just constants Bit instances) would make the most sense.

rsetaluri:
Got it, makes sense. What about constant values of Clock/Reset/etc.?

I see this from Pat:
> Thinking about this in hindsight, I think we should depreciate VCC and GND for values of Bit.

rsetaluri:
@leonardt take a look at bit.py, digital.py in https://github.com/phanrahan/magma/pull/652/

(some other stuff in there, but trying to clean up circular dependencies, and blanket package imports)

PR #661, opened 2020-04-02
Fix bit const

rsetaluri:
See https://github.com/cdonovick/peak/issues/147 for more discussion.

Also related is #651. Honestly I think we need to rework a few things in the type system; not w/ respect to type hierarchy or relations or anything. Just w.r.t. interfaces and construction. Right now it seems very ad-hoc and not super robust (e.g. see `Array::__init__`). Discussion re: https://github.com/cdonovick/peak/issues/146 also comes into play, regarding protocol types.

### DIFF
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -12,7 +12,7 @@ from hwtypes import AbstractBitVector, AbstractBitVectorMeta, AbstractBit, \
     InconsistentSizeError
 from .compatibility import IntegerTypes
 from .ref import AnonRef
-from .bit import Bit, VCC, GND
+from .bit import Bit
 from .array import Array, ArrayMeta
 from .debug import debug_wire
 from .t import Type, Direction, In, Out
@@ -94,7 +94,7 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
             self.ts = args[0].ts[:]
             # zext for promoting width
             for i in range(len(self) - len(args[0])):
-                self.ts.append(GND)
+                self.ts.append(Bit.GND)
             Type.__init__(self, **kwargs)
         else:
             Array.__init__(self, *args, **kwargs)
@@ -109,12 +109,13 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
         if not self.const():
             raise Exception("Not a constant")
 
-        def convert(x):
-            if x is VCC:
+        def _convert(x):
+            if x is type(x).VCC:
                 return True
-            assert x is GND
+            assert x is type(x).GND
             return False
-        return [convert(x) for x in self.ts]
+
+        return [_convert(x) for x in self.ts]
 
     def __int__(self):
         if not self.const():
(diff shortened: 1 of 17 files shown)


magma/bits.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, *args, **kwargs):
             self.ts = args[0].ts[:]
             # zext for promoting width
             for i in range(len(self) - len(args[0])):
-                self.ts.append(GND)
+                self.ts.append(Bit.GND)
```
leonardt, line 97: Should this be Bit(0) for consistency with elsewhere?

rsetaluri:
See garnet flow passing: https://buildkite.com/stanford-aha/aha-flow/builds/1226

magma/bits.py
rsetaluri: Fixed

rsetaluri:
@leonardt Agree with everything. It looks like there are 3 distinct issues at play; here they are along with my suggested solution:

1. Type constructors vs. type conversions, e.g. `Bits()` vs `bits()`. I think type constructors should be very simple, and be able to build values only from python POD's. I can write up a set of rules we can implement. Additionally, I think type conversions should all be from one magma type to another. That way these 2 mechanisms are completely separate. Ideally, the lower case "conversion" functions would be called "casts" or "reinterprets" but NBD. The only issue here is backwards compatibility since these constructors/conversions are used all over the place. May require a big overhaul.
2. Whether or not `GND` or `VCD` should be special named bits. I don't think they should be. We should just keep a `value` member which is `None` in the non-const case, and `0` or `1` otherwise. (See https://github.com/phanrahan/magma/blob/ee8da8eb670e9d6bf5878999b752f48eb0b139b3/magma/bit.py#L122-L137; already kind of doing this but seems inconsistent.) This somewhat ties into the next point, when it comes to checking equality...
3. The `==` operator can not be used for checking in the meta-language for equality to 0/1 because `==` instances a circuit. Well, this is only partly true because it does so for `Bit`, `Bits`, `SInt`, and `UInt` *always*, and *sometimes* for `Array`, `Tuple`, `Clock`, etc. (see https://github.com/phanrahan/magma/blob/ee8da8eb670e9d6bf5878999b752f48eb0b139b3/magma/array.py#L175-L178). It is a bit weird for `eq` to return a python value T/F sometimes, and a `Bit` other times. It makes it hard to have generic code without checking for these 2 cases. Instead, I think `==` operator on all types should return a Bit; for nested/complex types, we just dispatch on the subtypes. If you want to check that a bit is 0/1, you should do `x.const()` then `x.const_value()` which returns T/F (or 0/1). If you really want to get a python value for an equality op,  we can create a `def is(self, other)` which recursively does like `self is other or all(x.is(y) for x, y in zip(self, other))`.

What do you think?  Maybe I can start an issue to document all this type stuff and nail it down.

rsetaluri:
...but for now will merge to fix the present bug

PR #662, opened 2020-04-02
Fix pyverilog bugs

rsetaluri:


magma/frontend/pyverilog_importer.py
```diff
@@ -39,6 +39,10 @@ def _evaluate_node(node, params):
         l = _evaluate_node(node.left, params)
         r = _evaluate_node(node.right, params)
         return l // r  # floor division
+    if isinstance(node, pyverilog.vparser.ast.Times):
```
rsetaluri, line 42: this leaked in too...but it's useful :D

PR #663, opened 2020-04-02
Fix wrong argument to unflatten

rsetaluri:
unflatten expects a python list. An issue arose because we were
passing in a magma Array (instead of its 'ts').


rsetaluri:
@leonardt good question. to be quite honest, I'm not sure why this caused any issues, because the array itself is "as iterable" as its `.ts`

PR #664, opened 2020-04-02
Insert wire improvement

leonardt:
Fixes a bug where m.Bits are bit-blasted in the temporary wire generation.

Before, code such as

```
    class Main(m.Circuit):
        io = m.IO(I=m.In(m.Bit), O0=m.Out(m.Bit), O1=m.Out(m.Bit))

        x = m.Bits[2](name="x")
        for i in range(2):
            x[i] @= io.I
        io.O0 @= x[0]
        io.O1 @= x[1]
```
would produce the following verilog
```
module Main (
    input I,
    output O0,
    output O1
);
wire x_0;
wire x_1;
assign x_0 = I;
assign x_1 = I;
assign O0 = x_0;
assign O1 = x_1;
endmodule
```

Notice that x_0 and x_1 have been bit-blasted out even though it's a temporary of type Bits.

With these changes, the code produces

```
module Main (
    input I,
    output O0,
    output O1
);
wire [1:0] x;
assign x = {I,I};
assign O0 = x[0];
assign O1 = x[1];
endmodule
```

It does this by inspecting temporary values and if we find a member of a Bits value, we only emit a single wire for the parent Bits value.  Once the parent exists, we simply select the specified index.

The tests also include a more complex example for a nested array of bits, checking that
```
    class Main(m.Circuit):
        io = m.IO(I=m.In(m.Bits[2]), O0=m.Out(m.Bits[2]), O1=m.Out(m.Bits[2]))

        x = m.Array[2, m.Bits[2]](name="x")
        for i in range(2):
            x[i] @= io.I
        io.O0 @= x[0]
        io.O1 @= x[1]
```

produces

```
module Main (
    input [1:0] I,
    output [1:0] O0,
    output [1:0] O1
);
wire [1:0] x_0;
wire [1:0] x_1;
assign x_0 = I;
assign x_1 = I;
assign O0 = x_0;
assign O1 = x_1;
endmodule
```

Notice here we intend to get one level of flattening (so only leaf level Bits values are preserved).

### DIFF
--- a/magma/passes/insert_coreir_wires.py
+++ b/magma/passes/insert_coreir_wires.py
@@ -47,6 +47,9 @@ class InsertCoreIRWires(DefinitionPass):
     def _make_wire(self, driver, value, definition):
         is_bits = (isinstance(driver.name, ArrayRef) and
                    isinstance(driver, Digital))
+
+        # If this is a member of a bits, emit only one wire for the parent
+        # Bits, since it shouldn't be flattened
         if is_bits:
             index = driver.name.index
             driver = driver.name.array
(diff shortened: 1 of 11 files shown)


leonardt:
Yea I think there's an issue if not all the wires are connected (in your example, tmp[0] is unused), fortunately I think in the blocked use case everything is wired up.

PR #666, opened 2020-04-03
Add 'driving' API

rsetaluri:
Sometimes we want to know what values a bit is driving, similar to how `value()` tells us what value is driving a bit. Of course the difference is that `value()` is 1-to-1 (or None), where as a bit can be driving many other bits. Therefore I proposed 2 functions `driving()` and `driving_all()`, which can be described as follows:

- `T::driving()` returns a value of type `T` (flipped) if the value is driving *exactly* one other value (defined recursively of course for Array, Tuple, etc.). If it is not exactly one, then we return `None`.
- `T::driving_all()` returns a python data-type value, where leaf level objects are lists of driving values. So for example, `Bit::driving_all()` will return a list of all bits this bit is driving; `Array::driving_all()` will return a list of lists, where each element of the outer list is a list of all bits each bit of the array is driving; `Tuple::driving_all()` will return a python dict. We can't return the magma types as they exist (as we can in `driving`) because each element may be driving multiple bits, which can't be currently represented. Perhaps there's a more elegant way to do this, but ultimately this is the data-structure we want.

#667 is an alternative implementation which uses the existing `value()` function to provide the same functionality. See there for more discussion.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -350,16 +350,8 @@ class Array(Type, metaclass=ArrayMeta):
                 return False
         return True
 
-    def driving_all(self):
-        return [t.driving() for t in self]
-
     def driving(self):
-        ts = [t.driving() for t in self]
-        if any(t is None for t in ts):
-            return None
-        if Array._iswhole(ts):
-            return ts[0].name.array
-        return type(self).flip()(*ts)
+        return [t.driving() for t in self]
 
     def wired(self):
         for t in self.ts:
--- a/magma/array.py
+++ b/magma/array.py
@@ -350,6 +350,17 @@ class Array(Type, metaclass=ArrayMeta):
                 return False
         return True
 
+    def driving_all(self):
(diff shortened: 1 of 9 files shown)


leonardt:
Hmm both seem reasonable, and also aren't necessarily mutually exclusive? It seems there's a subtle difference between driving and value.  value will give you the unique "drivee" if it exists, while "driving" will give you all the drivees, I could imagine a use case for both.  Perhaps we can add "driving" for all the drivees (driving_all in this PR) and reuse value for "driving"?

If we want to choose one interface, I'm a little more partial to this one because it's more explicit in the sense that the driving/drivee relationship is distinguished by different methods, rather than overloading one interface, but I don't feel too strongly.

rsetaluri:
Makes sense. I went with the route of using `driving()` to mean a list of *all* driving bits. `value()` now has the semantics you mentioned (w/ updated docstring).

PR #667, opened 2020-04-03
value() returns driving value

rsetaluri:
Alternate implementation of #666 (😈)

Mechanically, this works mostly OOTB, since previously we raised an exception if the bit was an output. I don't think many places were checking for this anyway, so most code-paths are unaffected. The major one that is affected is that using `InOut`. The changes in `circuit.py` correspond to that.

This changes the semantics of `value()` to be "give me the value that drives me if there is one, otherwise give me the unique thing I drive if there is one". This somewhat stuffs multiple meanings into `value()` but I still like it since `value()` is kind of generic. Not tied to this however; would be ok reverting `value()` to the old semantics and adding `connection()` to mean this bidirectional thing...

Also, this PR doesn't include the `driving_all()` type of functionality. I was thinking of adding a `flat_value()` function which basically does something like `driving_all()` for **both** driver and driving (i.e. return a list of all things this bit is driving, or a POD structure of all things driving me, even if some are None).

### DIFF
--- a/magma/wire_container.py
+++ b/magma/wire_container.py
@@ -75,6 +75,13 @@ class Wire:
             return self.bit
         return None
 
+    def flat_value(self):
+        if self.driver is not None:
+            return self.driver.bit
+        if self.driving:
+            return self.driving
+        return None
+
     def value(self):
         """
         Return the driver of this wire
--- a/magma/wire_container.py
+++ b/magma/wire_container.py
@@ -79,11 +79,11 @@ class Wire:
         """
         Return the driver of this wire
         """
-        if self.bit.is_output():
-            raise TypeError("Can only get value of non outputs")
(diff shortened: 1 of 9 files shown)


magma/wire_container.py
```diff
@@ -79,11 +79,11 @@ def value(self):
...
-        if self.bit.is_output():
-            raise TypeError("Can only get value of non outputs")
-        if self.driver is None:
-            return None
-        return self.driver.bit
+        if self.driver is not None:
+            return self.driver.bit
+        if len(self.driving) == 1:
+            return self.driving[0].bit
+        return None
```
leonardt, line 93: We should update the docstring to include the new behavior

rsetaluri:
Agreed, closing this and working off #666

PR #669, opened 2020-04-06
Add methods to MagmaProtocol

rsetaluri:
@cdonovick incremental step towards getting Type <-> MagmaProtocol parity. Unblocks a specific peak issue (https://github.com/cdonovick/peak/issues/146).

### DIFF
--- a/magma/protocol_type.py
+++ b/magma/protocol_type.py
@@ -54,6 +54,9 @@ class MagmaProtocol(metaclass=MagmaProtocolMeta):
     def is_output(cls):
         return cls._to_magma_().is_output()
 
+    def iswhole(self):
+        return self._get_magma_value_().iswhole()
+
     def value(self):
         return self._get_magma_value_().value()
 
@@ -65,6 +68,9 @@ class MagmaProtocol(metaclass=MagmaProtocolMeta):
 
     def flatten(self):
         return self._get_magma_value_().flatten()
+
+    def __iter__(self):
+        return iter(self._get_magma_value_())
 
     @property
     def name(self):


rsetaluri:
Well, my idea is to unify `Type`/`MagmaProtocol`, whatever we call it in the end. There should be parity of the methods we can expect to call on a "magma value", and `iswhole()` is definitely one of them: we check for it everywhere. Whether or not `__iter__` should be an expected method on types or not is debatable, open to either way (we definitely expect it in some cases, would have to update those call sites).

My idea is that in order to specify a custom magma type, you need to implement all of these methods. Now, one way to do that is to always translate to and from a "base" magma type, which already has these methods defined, so the base class would just dispatch to those methods on the "base type" and the user would *only* have to implement to_magma/from_magma. Right now that is what we're calling "MagmaProtocol", but really "magma/t.py:Type" is closer to the protocol, and "MagmaProtocol" is a specific "WrapperType" partial implementation of Type protocol.

cdonovick:
@rsetaluri  https://github.com/phanrahan/magma/issues/671

cdonovick:
Okay this is not what I had envisioned for the protocol but could be workable for now.  However this is a mixin not a protocol.   Compare the sequence protocol with the Sequence mixin https://docs.python.org/3/library/functions.html#iter
https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence

if an type implements `__getitem__` and `__len__`  than `iter(obj)` works.  This is the sequence protocol.
if an type implements `__getitem__` and `__len__` and inherits `Sequence` then it will have `__iter__`,  and a bunch of other methods defined for free. 

The key difference being the sequence protocol doesn't actually add a `__iter__` method.  The function `iter` just knows how to handle the protocol.

If we go the mixin route it effectively reserves a bunch of method names and attributes in a way that is hard for me to enforce and will inevitably lead to annoying bugs.  It will also limit magma as the interface of type will have to be effectively static going forward as adding methods will reserve new names which may break existing code.

Further there are certain methods which I could imagine being expected of magma objects that I absolutely cannot support, specifically `__eq__`, `__ne__`, and `__hash__`.  `__iter__` is not currently used by anything I have but I am strongly opposed to adding it to the mixin as it would block wrapped types having an `__iter__` of their own.

rsetaluri:
@cdonovick that all makes sense to me. At some point we do have to have the structural information of `Type`, that is, very specific methods like `iswhole()` etc. I understand that not being able to have custom functions like `__iter__` etc. because they are reserved by the mixin is restrictive. But having to call `_value_()` before we use any "magma like value" add's significant overhead (though you could argue is more explicit). Are these the 2 alternatives? Are there any other alternatives?

cdonovick:
I know this would require significant refactoring but if we made all of the methods free functions instead of methods it would be pretty easy
For example:
```
def iswhole(x):
    if not isinstance(x, magma.t):
       # alternatively have magma.t implement the protocol and do this unconditionally
        try:
            x = x._get_magma_value_()
        except AttributeError:
            raise TypeError("iswhole expected a magma type or protocol")
    # do magma.t.iswhole impl
```
Now as we are going to have bunch of these functions I would probably turn this into a decorator:
```
def unpack_protocol(fn):
    @functools.wraps(fn):
    def wrapped(x)
        # do the try block from above
        return fn(x)

@unpack_protocol
def iswhole(x):
    # do magma.t.iswhole impl
```

cdonovick:
I would be willing to write a refactoring tool to help the transition if it would help.

rsetaluri:
I see; that makes sense. Wouldn't be all that bad, but one question is a lot of those functions are still type dependent, i.e. `Array.iswhole()` looks non-trivially different from `Tuple.iswhole()`. Are you suggesting the free-function `iswhole()` checks for the types, or we still have the member methods `Array.iswhole()`, `Tuple.iswhole()`, etc.?

On another note, is this pythonic? What is the principle that makes this more pythonic than the other options?

Finally, if this doubles as an opportunity to clean up/refactor some of the interfaces between types/backends/etc. I'm in favor.

rsetaluri:
@cdonovick are you ok merging this as temporary and doing a design review soon and establishing the "right" thing for types/protocol?

cdonovick:
Yes lets merge for now

PR #675, opened 2020-04-07
Fix regression in const logic

leonardt:
The const change causes nested arrays of constants to be treated as constants in the coreir compiler, but really we should only be handling constant Bits, since the coreir const primitive is defined as a bit vector type, so we should recurse unless it's filled with Bit constants, not generally constant.


tests/test_coreir/test_const.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_const_logic():
+    class Main(m.Circuit):
+        io = m.IO(O=m.Out(m.Array[1, m.Bits[3]]))
+
+        io.O @= m.array([m.bits(0, 3)])
+
+    m.compile("build/test_const", Main)
+    assert check_files_equal(__file__, f"build/test_const.v",
```
rsetaluri, line 12: can we test against json rather than verilog?

magma/backend/coreir_transformer.py
```diff
@@ -207,7 +207,7 @@ def connect(self, module_defn, port, value, non_input_ports):
         def get_source():
             if isinstance(value, Wireable):
                 return value
-            if isinstance(value, Array) and is_const(value):
+            if isinstance(value, Array) and is_const_digital(value):
```
rsetaluri, line 210:
```suggestion
            if isinstance(value, Array) and issubclass(type(value).T, Digital) and value.const():
```

rsetaluri: we can just capture all that logic inline here. Then we can get rid of `is_const()` and the import of it. 

`all(x.const() for x in array)` duplicates `Array.const()` implementation.

rsetaluri: Actually we should probably do 

```python
if isinstance(value, Bits) and value.const()
```

because either (a) if you do `Array[N, Bit]` then you should really just do `Bits[N]`, or (b) if you do `Array[N, Clock](0)` or something then that should be different in the backend than a const bits (b/c clock types are differentiated). But not sure if external code is doing (a) or (b) and expecting it to work.

leonardt: don't know if const Clock exists and doubt it's being used (b)
with this change if you do Array[N, Bit](0) you'll get an array of Bit(0) consts, which doesn't seem unreasonable, but one could argue that the old logic generates more readable code.

rsetaluri: Yeah. I'm ok with either

`if isinstance(value, Bits) and value.const():`

or 

`if isinstance(value, Array) and issubclass(type(value).T, Bit) and value.const():`

PR #680, opened 2020-04-13
Make unused/undriven passes only operator on Definitions

leonardt:
Otherwise, they will recurse into primitives and modify their definitions (e.g. the terminate unused pass descends into the unused definition to terminate its unused input port, same goes for the binary ops).  If the circuit doesn't have a definition, we don't have to terminate the unused ports since coreir will skip compiling anyways (even if we did, the user probably won't care because this would imply there's nothing else in the definition, so compiling it as if it's a declaration is fine too, the main issue with definitions is that this will trigger an error in coreir, but for declarations it won't matter).

### DIFF
--- a/magma/passes/passes.py
+++ b/magma/passes/passes.py
@@ -108,7 +108,7 @@ class InstanceGraphPass(Pass):
                 self(vert, edges)
 
 
-class EditCircuitPass(DefinitionPass):
+class EditDefinitionPass(DefinitionPass):
     @abstractmethod
     def edit(self, circuit):
         raise NotImplementedError()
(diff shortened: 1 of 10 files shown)


magma/passes/passes.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, main):
                 self(vert, edges)
 
 
-class EditCircuitPass(CircuitPass):
+class EditCircuitPass(DefinitionPass):
```
rsetaluri, line 111: maybe rename to `EditDefinitionPass` then?

PR #681, opened 2020-04-13
Bind after uniquification

leonardt:
Unique modules should have unique binds, and binds should not be re-run when cacheing occurs.

Also fixes minor undriven/unused logic bug for mixed direction ports that came up in the new  test.

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -15,7 +15,7 @@ class _ClockType(Digital):
         Bit.unused(self)
 
     def undriven(self):
-        Bit.unused(self)
+        Bit.undriven(self)
 
 
 class Clock(_ClockType, metaclass=DigitalMeta):
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -45,12 +45,14 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
     opts = kwargs.copy()
     compiler = _make_compiler(output, main, basename, opts)
 
-    # Steps to process bind's and inline verilog generation.
+    # Steps to process inline verilog generation.
     ProcessInlineVerilogPass(main).run()
-    BindPass(main, compile).run()
 
     # Default behavior is to perform uniquification, but can be overriden.
     uniquification_pass(main, opts.get("uniquify", "UNIQUIFY"))
+
+    BindPass(main, compile).run()
+
     if opts.get("drive_undriven", False):
         DriveUndrivenPass(main).run()
     if opts.get("terminate_unused", False):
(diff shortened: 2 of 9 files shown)


magma/clock.py
```diff
@@ -15,7 +15,7 @@ def unused(self):
         Bit.unused(self)
 
     def undriven(self):
-        Bit.unused(self)
+        Bit.undriven(self)
```
rsetaluri, line 18: oops :)

magma/compile.py
```diff
@@ -45,12 +45,14 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
...
 
-    # Steps to process bind's and inline verilog generation.
+    # Steps to process inline verilog generation.
     ProcessInlineVerilogPass(main).run()
-    BindPass(main, compile).run()
 
     # Default behavior is to perform uniquification, but can be overriden.
     uniquification_pass(main, opts.get("uniquify", "UNIQUIFY"))
+
+    BindPass(main, compile).run()
```
rsetaluri, line 54: Add a comment explaining the ordering and the dependencies between these steps.

magma/passes/drive_undriven.py
```diff
@@ -2,12 +2,20 @@
 from ..is_definition import isdefinition
 
 
+def _drive_if_undriven_input(port):
+    if port.is_mixed():
+        # Sum so it doesn't short circuit
+        return sum(_drive_if_undriven_input(p) for p in port)
```
rsetaluri, line 8:
```suggestion
        undrivens = (_drive_if_undriven_input(p) for p in port)
        return any(undrivens)
```

This is a little more readable

magma/passes/terminate_unused.py
```diff
@@ -2,12 +2,20 @@
 from ..is_definition import isdefinition
 
 
+def _terminate_if_unwired_output(port):
+    if port.is_mixed():
+        # Sum so it doesn't short circuit
+        return sum(_terminate_if_unwired_output(p) for p in port)
```
rsetaluri, line 8:
```suggestion
        terminated = (_terminate_if_unwired_output(p) for p in port)
        return any(terminated)
```

This is a little more readable

magma/passes/drive_undriven.py
leonardt: Any doesn't work because it short circuits

leonardt: Used list comp to prevent short circuiting.

PR #682, opened 2020-04-13
Short circuit array logic

leonardt:


magma/clock.py
```diff
@@ -90,17 +90,31 @@ def ClockInterface(has_enable=False, has_reset=False, has_ce=False,
...
-    if isinstance(port, (Array, Tuple)):
-        return sum([_get_clocks(elem, clocktype) for elem in port], [])
+    if isinstance(port, Tuple):
+        return sum((_get_clocks(elem, clocktype) for elem in port), [])
+    if isinstance(port, Array):
+        result = _get_clocks(port[0], clocktype)
+        # Only traverse all children circuit if first child has a clock
+        if result:
+            result += sum((_get_clocks(elem, clocktype) for elem in port[1:]), [])
+        return result
```
rsetaluri, line 117: oh another option here is to make sure the common case is fast. presumably the common case is connecting to a top-level clock port (not a clock in a nested type). We could first check if any port is itself an instance of `clocktype` before doing the recursive checks.

maybe in `wireclocktype` the first few lines should be:

```python
clks = (port if isinstance(port, clocktype) else None
        for port in defn.interface.ports.values())
defnclk = _first(clks)
if defnclk is None:
  # Call _get_first_clock() as described in above comment.
```

magma/clock.py
```diff
@@ -90,17 +90,31 @@ def ClockInterface(has_enable=False, has_reset=False, has_ce=False,
...
+            for elem in port[1:]:
+                _wire_clock_port(elem, clocktype, defnclk)
+    elif isinstance(port, clocktype) and not port.driven():
         wire(defnclk, port)
+        wired = True
+    return wired
 
 
 def _get_clocks(port, clocktype):
-    if isinstance(port, (Array, Tuple)):
```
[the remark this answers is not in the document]
leonardt: I'm sure we can generalize this for other use cases (get all clocks), but since this is internal, I think it's okay to specialize it to the use case of finding the first clock

magma/backend/coreir_.py
```diff
@@ -82,18 +82,26 @@ def simulate(self, value_store, state_store):
...
-                self.wrap_if_named_type(t, definition)
-            return
+                wrapped |= self.wrap_if_named_type(t, definition)
+            return wrapped
+        if isinstance(port, Array):
+            wrapped = self.wrap_if_named_type(port[0], definition)
+            if not wrapped:
+              return False
+              for t in port[1:]:
+                  self.wrap_if_named_type(t, definition)
```
rsetaluri: oops think the formatting from my suggestion got a little screwed up here

PR #684, opened 2020-04-14
Hotfix issue with length zero array

leonardt:
Regression introduced by https://github.com/phanrahan/magma/pull/682 where working with an array of length 1 will cause the code path to try to iterate over an array of length 0.  For this to work, magma needs to construct an array of length 0, which causes an exception (we don't allow length 0 arrays https://github.com/phanrahan/magma/blob/f04540f7403641e6d8f32be07da85252504b26f8/magma/array.py#L92).  I think ideally we'd work out some pythonic behavior where slicing off the end works since this is a common python pattern, but for now this change adds a guard to the new code path so we can unblock @mbstrange2

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -91,8 +91,12 @@ class InsertWrapCasts(DefinitionPass):
             wrapped = self.wrap_if_named_type(port[0], definition)
             if not wrapped:
                 return False
-            for t in port[1:]:
-                self.wrap_if_named_type(t, definition)
+            # TODO: Magma doesn't support length zero array, so slicing a
+            # length 1 array off the end doesn't work as expected in normal
+            # Python
+            if len(port) > 1:
+                for t in port[1:]:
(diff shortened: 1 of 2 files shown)


magma/backend/coreir_.py
```diff
@@ -91,8 +91,12 @@ def wrap_if_named_type(self, port, definition):
             wrapped = self.wrap_if_named_type(port[0], definition)
             if not wrapped:
                 return False
-            for t in port[1:]:
-                self.wrap_if_named_type(t, definition)
+            # TODO: Magma doesn't support length zero array, so slicing a
```
rsetaluri, line 94: it might be cleaner to just change this to `for t in port.ts[1:]` rather than the explicit check here. Since we never really need a handle on the array returned by `port[1:]`, just its underlying `ts` this might be ok. Not perfect either but seems slightly better than the len check. If you do stick with the len check, maybe prefer

```python
if len(port) == 1:
    return True
for t in port[1:]:
    ...
return True
```

PR #685, opened 2020-04-15
Memoize first clock type of defn

rsetaluri:


magma/clock.py
```diff
@@ -126,16 +127,22 @@ def _get_first_clock(port, clocktype):
...
+        return defnclk
+    # Check recursive types
+    clks = (_get_first_clock(port, clocktype)
+            for port in defn.interface.ports.values())
+    return _first(clks)
+
+
+@functools.lru_cache(maxsize=None)
+def wireclocktype(defn, inst, clocktype):
+    defnclk = _get_first_clock_of_defn(defn, clocktype)
```
leonardt, line 145: I think we want to memoize the _get_first_clock function right? But I'm not sure if this is really where the best savings will occur, since for one definition, we'll only need to get the first clock once.

I think where it becomes useful is in the wiring of instances.  We may have a large number of instances, and we have to check each of their interfaces to find the first clock.  If we can memoize the first clock for a definition, then for each instance we can simply lookup the default clock rather than traversing the interface.  I suspect most of the savings will come here, since will be repeating the clock lookup logic for instances of the same definition, while definitions only need to lookup their internal clock once

leonardt: Also, `wireclock` will repeat this logic for each clock type.  I think it would be more efficient to wire a set of clock types and handle them all as we're traversing the interface once (rather than traversing the interface each time for each clock type).

rsetaluri: (a) yes, oops, sorry put the decorator on the wrong function.

(b) with the decorator on the right function, we will only traverse the interface *once* per (definition, type) pair. If I understand correctly, you're suggesting we simultaneously do the traversal for all the types?

(c) yes, good point about instances. Even if we find we can find the first clock of a defn fast, we still find *all* the clocks for each instance, and this is repeated even for multiple instances of the same defn. Will try to implement this.

I think the best way to do this is to have an acceleration data structure associated with each defn which is Map[Type, List[String]], where for each type we have the name of all the ports which match that type. We can make this a query and then cache the result per definition. I think it would also be prudent to have an option to stop on the first one if needed.

rsetaluri: @leonardt I can try to implement the above.

rsetaluri: Actually, I realize (c) is more complicated for nested types. If for example, defn.x[0].y[1] happens to be a clock type, re-targetting that reference to an instance is non-trivial (we basically need to capture all the clock types "by reference"). Will think about the best way to do this.

If we avoid nested types this becomes much easier as all we have to do is save the port name and do `getattr` on instances. But it sounds like nested clocks are definitely important.

PR #687, opened 2020-04-16
Inline verilog symbol map

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/868

Updates the magma inline verilog logic to use a module instantiation to manage inline verilog references.  This removes the requirement on magma to manage the name mapping of flattened types since the connection references are updated by the fact that the inline verilog instance is explicit in the graph.  For now, magma flattens nested array references in the connection structure and reconstructs them in the verilog with a concat node.  This issue will be avoided once we start code generating n-d arrays.  We don't handle interpolation of tuples into the verilog since there would be no analog verilog structure anyways.


magma/backend/coreir_.py
```diff
@@ -96,21 +96,22 @@ def wrap_if_named_type(self, port, definition):
...
         else:
-            T = type(value).flip()
+            T = In(type(value))
 
         with definition.open():
             port.unwire(value)
             inst = self.define_wrap(T, type(port), type(value))()
-            wire(getattr(inst, "in"), convertbit(value, type(port)))
+            wire(convertbit(value, type(port)), getattr(inst, "in"))
             wire(inst.out, convertbit(port, type(value)))
```
leonardt, line 115: Some bug fixes that came up when I started wiring clock signals to unused for the inline verilog logic

magma/inline_verilog.py
```diff
@@ -1,15 +1,133 @@
...
+                    with defn.open():
+                        temp_name = "_magma_inline_wire"
+                        temp_name += f"{cls.inline_verilog_wire_counter}"
+                        temp = Wire(type(value.port))(name=temp_name)
+                        cls.inline_verilog_wire_counter += 1
+                        temp.I @= value.port
+                        temp = PortView(temp.O, parent)
+                        value._magma_inline_wire_ = temp
+                    value = value._magma_inline_wire_
+            wire(value, getattr(inst, key))
```
leonardt, line 130: Will refactor this to reduce the code repetition

magma/placer.py
```diff
@@ -141,7 +141,7 @@ def place(self, inst):
         inst.defn = LazyCircuit
         # Setup view now because inline strings might use it during defn
         for sub_inst in getattr(type(inst), "instances", []):
-            setattr(inst, sub_inst.name, InstView(sub_inst, inst))
+            setattr(inst, sub_inst.name, InstView(sub_inst, InstView(inst)))
```
leonardt, line 144: Make top-level parent an instview so it's consistent (otherwise we have to check whether it's an instview or inst when walking up the tree)

PR #690, opened 2020-04-22
Pass lib dependencies to verilog compile

rsetaluri:


rsetaluri:
See aha tests at https://github.com/StanfordAHA/aha/pull/526

PR #692, opened 2020-04-23
Inline verilog bug fixes

leonardt:
Uses default getattr logic for backwards compatability with old-style circuits that don't have inline_verilog_modules fields.

Fixes uniquification of inline verilog modules to use rename API instead of just setting the name field.

Appends _inline_verilog_ to module name for easier debug/readability

### DIFF
--- a/magma/uniquification.py
+++ b/magma/uniquification.py
@@ -57,7 +57,8 @@ class UniquificationPass(DefinitionPass):
                 # Inline modules use a naming scheme based on the container
                 # name, so we update their names by replacing the name
                 # substring
-                for module in definition.inline_verilog_modules:
+                for module in getattr(definition, "inline_verilog_modules",
+                                      []):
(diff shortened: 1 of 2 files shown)


magma/uniquification.py
```diff
@@ -37,6 +37,16 @@ def _hash(definition):
...
 
 
+def _rename_inline_verilog(definition, new_name):
+    # Inline modules use a naming scheme based on the container
+    # name, so we update their names by replacing the name
+    # substring
+    for module in getattr(definition, "inline_verilog_modules",
+                          []):
+        type(module).rename(module, module.name.replace(definition.name,
+                                                        new_name))
```
rsetaluri, line 47:
```suggestion
    modules = getattr(definition, "inline_verilog_modules", [])
    for module in modules:
        new_name = module.name.replace(definition.name, new_name)
        type(module).rename(module, new_name)
```
just a lil more readable :)

PR #694, opened 2020-04-25
Revert to old coreir-verilog compilation logic

rsetaluri:


rsetaluri:
Ok this library stuff has caused enough headache. I've reverted to the old logic, with a env flag that allows enabling "fast" coreir-verilog compile. That makes it pretty much an experimental feature that's turned off by default. This should unblock everything...

Now to figure out how to actually fix the library stuff...

PR #695, opened 2020-04-27
Fix zext bug

leonardt:
Magma uses last index ([-1]) for msb (see https://github.com/phanrahan/magma/blob/master/magma/conversions.py#L241)


leonardt:
¯\_(ツ)_/¯ unused code path? It works fine if you do an explicit zext/sext, so perhaps most people were using that pattern rather than using `uint` to do the implicit zext

PR #696, opened 2020-04-28
Run inline verilog logic after uniquification

leonardt:
There was an issue where inline verilog modules were not being renamed properly in the uniquification flow.  This simplifies the logic (and fixes the bug) by just running the inline verilog logic after uniquification (so the generated inline verilog modules use the already uniquified name)

### DIFF
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -45,11 +45,11 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
     opts = kwargs.copy()
     compiler = _make_compiler(output, main, basename, opts)
 
-    # Steps to process inline verilog generation.
-    ProcessInlineVerilogPass(main).run()
-
     # Default behavior is to perform uniquification, but can be overriden.
     uniquification_pass(main, opts.get("uniquify", "UNIQUIFY"))
+
+    # Steps to process inline verilog generation.
+    ProcessInlineVerilogPass(main).run()
 
     # Bind after uniquification so the bind logic works on unique modules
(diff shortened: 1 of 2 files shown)


magma/uniquification.py
```diff
@@ -37,16 +37,6 @@ def _hash(definition):
...
 
 
-def _rename_inline_verilog(definition, new_name):
-    # Inline modules use a naming scheme based on the container
-    # name, so we update their names by replacing the name
-    # substring
-    modules = getattr(definition, "inline_verilog_modules", [])
-    for module in modules:
-        new_name = module.name.replace(definition.name, new_name)
-        type(module).rename(module, new_name)
```
leonardt, line 47: The bug seems to be that new_name was overwritten each iteration of the loop, which was unintended

magma/compile.py
```diff
@@ -45,12 +45,12 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
     opts = kwargs.copy()
     compiler = _make_compiler(output, main, basename, opts)
 
-    # Steps to process inline verilog generation.
-    ProcessInlineVerilogPass(main).run()
-
     # Default behavior is to perform uniquification, but can be overriden.
     uniquification_pass(main, opts.get("uniquify", "UNIQUIFY"))
 
+    # Steps to process inline verilog generation.
```
rsetaluri, line 51:
```suggestion
    # Steps to process inline verilog generation. Required to be run after uniquification.
```

PR #702, opened 2020-05-26
Add blackbox and stub passes

rsetaluri:


magma/passes/stub.py
```diff
@@ -0,0 +1,30 @@
...
+
+def _drive_if_undriven_input(port):
+    if port.is_mixed():
+        for p in port:
+            _drive_if_undriven_input(p)
+        return
+    # NOTE(rsetaluri): This may override previously wired inputs, resulting in a
+    # warning. In this case, this warning is benign.
+    if port.is_input():
+        port.undriven()
```
leonardt, line 15: Perhaps we should disable warnings in this pass? One simple way would be to set a magma "global" that disables warning reporting, which we can enable/disable in this pass.  A more fine-grained way might be to pass a flag through undriven that says ignore drivers or something.  Ideally we can avoid generating warnings internally, even if they are benign

tests/test_passes/test_stub.py
```diff
@@ -0,0 +1,63 @@
...
+                             f"gold/test_passes_stub_noop.json")
+
+
+def test_compiler_args(ckts):
+    _Foo, _Bar, _Top = ckts
+    m.compile("build/test_passes_stub_compiler_args", _Top,
+              output="coreir", stubs=(_Foo,))
+    assert check_files_equal(__file__,
+                             f"build/test_passes_stub_compiler_args.json",
+                             f"gold/test_passes_stub_basic.json")
```
leonardt, line 63: Can you add a test that stubs a circuit with an internal instance?  I believe that you also need to remove those (since the compilation algorithm will iterate over the instances and generate the coreir for them)

rsetaluri: This is a good point. In fact, this breaks the abstraction that we shouldn't be deleting from circuits, and it looks like all downstream compilers place even floating instances. One option would be to make the compilers all deal with floating instances differently (not place them), then all we would have to do is unwire everything. However, this changes the semantics of the compiler and may affect other codepaths.

Instead, I think stubbing/black-boxing should be ephemeral -- the circuits should not be modified and only the invocation of the compiler for which these options are passed should consider those options. The only downside is that each compiler (verilog, coreir, etc.) would have to duplicate the logic to deal with stubs/black-boxes (rather than have this fall out of the modified circuits). I think this is still better though, esp. since we really only use CoreIR. I'll modify the PR to do this.

rdaly525:
For stubbing, would it also be desirable to instance "term" instances for the inputs? 

If you want this behavior in CoreIR, blackboxing can be done by just removing the module definition. Stubbing is similar but just with creating a new definition with the stub properties you want (instances of "undriven"/"term").

rsetaluri:
That's right. I actually think doing it in CoreIR might be safer/make more sense...

PR #705, opened 2020-05-29
Initial CoreIR Slice Support

leonardt:
For now, we discover slices in the backend and pack them into coreir slices.  The hope is that this will still reduce the number of API calls for connections as well as avoid coreir logic for traversing the bit-blasted slices and generating the bit-blasted coreir.

A follow-up PR will incorporate slices natively into the Bits __getitem__ syntax so we can represent them natively in python/magma (then the backend logic should be simplified to just translate these objects into coreir).


leonardt:
Bump, this is getting a bit old, would like to merge it before it get's out of sync or anything and would be good to start testing it in the wild.

PR #706, opened 2020-05-29
Avoid repr call in backend

leonardt:
When we generate the logging debug string, we're actually invoking repr on the type of the circuit (possibly many times for multiple instances).  For now, I think it's okay if we simply just log the name of the instance type.


leonardt:
Nothing major seen yet, but I'll keep poking around.  This one was particularly bad because it was run multiple times for each instance of each module (and expensive for large modules with lots of instances)

PR #707, opened 2020-05-29
Use is instead of == in equality check

leonardt:
We are cacheing types so we can and should be using `is` instead of __eq__ (this is for comparing type equality)


leonardt:
Yes and yes, this one showed up as being a nuisance in the profile particularly, but we should go and audit all the __eq__ logic for kinds and make sure they're using `is` where possible for performance.  The main issue arises in the issubclass check, which uses __eq__ (why we need to define it), but inside __eq__ we should be using is since we know that was can use object equality for kind equality.

Issue #708, opened 2020-05-31
coreir-verilog inline issue


splhack:
#709, enable `inline=True` test with

    @pytest.mark.parametrize("inline", [True, False])

Yosys and verilator with `--language 1364-2005` don't compile generated verilog due to syntax error.

    %Error: tests/test_issues/build/test_708_inline_True.v:50:12: syntax error, unexpected $undefined
       50 |     .I1_x(8'(self_a_O_x + 8'h01)),
          |            ^
    %Error: Exiting due to 1 error(s)

leonardt:
Hmm, I think we'll need to add a flag to coreir to control verilog code generation compatibility

splhack:
FYI https://github.com/YosysHQ/yosys/pull/2179

PR #709, opened 2020-05-31
test for #708

splhack:
test for coreir-verilog inline=True issue

### DIFF
--- a/tests/test_issues/gold/test_708_inline_False.v
+++ b/tests/test_issues/gold/test_708_inline_False.v
@@ -1,0 +1,140 @@
+module coreir_reg #(
+    parameter width = 1,
+    parameter clk_posedge = 1,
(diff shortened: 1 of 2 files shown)


splhack:
File ".magma/make_Test.py", line 1, in <module>
        from mantle import DefineRegister
    ModuleNotFoundError: No module named 'mantle'

Hmm, I can't repro this Travis-CI error locally. The test should be similar with test_sequential.py.

leonardt:
Ah yes, we define a stub register generator for use in the magma test suite (so magma doesn't depend on mantle).  Here's an example of how you can import it for your test: https://github.com/phanrahan/magma/blob/master/tests/test_issues/test_545.py#L4-L11 (it doesn't appear locally because you likely have mantle installed so the import works)

PR #711, opened 2020-06-03
Prune infeasible paths

leonardt:
Related to https://github.com/phanrahan/magma/pull/710

This adds a primitive logic in the coroutine compiler to prune infeasible paths.  In this case, it merely searches the condition list for the inverse of the condition to be added (e.g. `if tms` and `if ~tms`).  Eventually we'll want to generalize this to a more sophisticated algorithm, e.g. using SMT to check for a possible solution to the current set of conditions.

### DIFF
--- a/magma/syntax/coroutine.py
+++ b/magma/syntax/coroutine.py
@@ -87,7 +87,16 @@ def is_yield(statement):
             and isinstance(statement.value, ast.Yield))
 
 
-def collect_paths_to_yield(start_idx, block):
+def infeasible(conds, next_cond):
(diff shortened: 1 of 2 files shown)


cdonovick:
Im not really familiar with this code but is this being done post ssa?  If not I'd be worried about the following case:
```python
if cond:
    ...
...
cond = ~cond
...
if ~cond:
    ...
```

leonardt:
Hmm that is a problematic case, doing it post SSA would resolve that, I'll look into moving the logic there

PR #712, opened 2020-06-05
Add m.reduce

leonardt:
Adds an api for generating the unary coreir/verilog reduce operators on Bits

### DIFF
--- a/magma/primitives/reduce.py
+++ b/magma/primitives/reduce.py
@@ -10,7 +10,7 @@ from magma.generator import Generator2
 from magma.t import In, Out
 
 
-def reduce_factory(coreir_name, operator):
+def _reduce_factory(coreir_name, operator):
     class Reduce(Generator2):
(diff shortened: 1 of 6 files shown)


magma/primitives/reduce.py
```diff
@@ -0,0 +1,46 @@
...
+            def simulate(self, value_store, state_store):
+                I = BitVector[width](value_store.get_value(self.I))
+                O = functools.reduce(operator, I.bits())
+                value_store.set_value(self.O, O)
+
+            self.simulate = simulate
+    return Reduce
+
+
+OP_MAP = {
```
rsetaluri, line 34: this stuff is elsewhere too I believe. fine for now, just making a note that we should refactor these probably.

rsetaluri, line 34: make `OP_MAP` and `reduce_factory` private ideally

PR #713, opened 2020-06-06
Fix sequential `__call__` for no args

splhack:
When sequential `__call__` doesn't have any arguments aside from `self` like the test,

    def __call__(self) -> m.Bits[3]:

combinational logic falls into Python function path due to the type check

                not any(isinstance(x, (Type, MagmaProtocol)) for x in args +
                        tuple(kwargs.values()))):
                    # If not called with at least one magma value, use the Python
                    # implementation
    >               return fn(*args, **kwargs)
    E               TypeError: unsupported operand type(s) for +: 'reg_PR' and 'UInt[(2, Out(Bit))]'

it requires AnonymousCircuitType for sequential Registers?


splhack:
@leonardt @rsetaluri could you merge this?

PR #714, opened 2020-06-09
Add insert prefix pass

leonardt:


leonardt:
This was mainly to show a demonstration to our FB users how to do it, they will likely just copy/paste the code into their codebase since they want to avoid doing a version upgrade for their frozen RTL.  We can/should add it to the pass standard library, but it's not urgent so if you want to refactor rename and move this into the passes directory before merging that's fine.  Also, we should start a doc page with the collection of standard passes provided by the core and what they do/how to use them.

leonardt:
Ok, just got word they are using it for the next version, so it's non-blocking so we should move this into the stdlib and provide the documentation on how to use it.

rsetaluri:
Closing in favor of #734

PR #716, opened 2020-06-10
Fix warning

cdonovick:


cdonovick:
Wouldn't normally merge my own commit without approval but this is safe.

Issue #598, opened 2020-02-25
Provide Magma Primitive Generators Parametrized by Type Needed by Aetherling


David-Durst:
Aetherling has a number of extensions to standard Magma/Mantle modules that handle any Magma type:
- [x] [LUTs](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/lut_any_type.py) - https://github.com/phanrahan/magma/pull/720
- [x] [MUXs](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/mux_any_type.py) - https://github.com/phanrahan/magma/pull/722
- [x] [Registers](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/register_any_type.py) - https://github.com/phanrahan/magma/pull/691
- [x] [ROMs](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/rom_any_type.py) - https://github.com/phanrahan/magma/pull/727
- [x] [Terms](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/term_any_type.py) - provided by `unused` interface
- [x] [SIPOs](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/sipo_any_type.py) - https://github.com/phanrahan/mantle/pull/183
- [x] [RAMs](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/ram_any_type.py) - https://github.com/phanrahan/magma/pull/726
  - [x] Add latency parameter to coreir mantle generator so we can emit specialized verilog (see https://github.com/phanrahan/magma/pull/726 for context w.r.t. to FPGA BRAM synthesis) - https://github.com/phanrahan/magma/pull/738

They use the [hydrate/dehydrate modules](https://github.com/David-Durst/aetherling/blob/master/aetherling/modules/hydrate.py) for flattening types. Those call out to CoreIR.

Can we merge them with Facebook's implementation? Alternatively, can we replace them with Facebook's if Facebook's produces higher quality Verilog?

leonardt:
Let's use this issue for tracking the generator support for AnyTypes

leonardt:
We should provide the above listed generators as magma primitives, similar to how we implement the new Mux primitive here: https://github.com/phanrahan/magma/blob/master/magma/mux.py#L35-L64

leonardt:
[...]

PR #720, opened 2020-06-13
Add LUT(T) primitive

leonardt:
For https://github.com/phanrahan/magma/issues/598

### DIFF
--- a/magma/primitives/lut.py
+++ b/magma/primitives/lut.py
@@ -69,4 +69,3 @@ class LUT(Generator2):
             luts.append(_CoreIRLUT(BitVector[n]([elem[i] for elem in
                                                  contents]))())
         self.io.O @= from_bits(T, fork(luts)(as_bits(self.io.I)))
-
--- a/magma/primitives/lut.py
+++ b/magma/primitives/lut.py
@@ -39,7 +39,7 @@ def _to_int(value):
         if not value.const():
             raise ValueError(f"Unexpected Bit value: {value}")
         return 1 if value is Bit.VCC else 0
-    elif not isinstance(value, Bits):
+    if not isinstance(value, Bits):
         value = as_bits(value)
     return int(value)
 
--- a/magma/primitives/lut.py
+++ b/magma/primitives/lut.py
@@ -34,6 +34,16 @@ class _CoreIRLUT(Generator2):
         self.coreir_configargs = {"init": contents}
 
 
+def _to_int(value):
+    if isinstance(value, Bit):
+        if not value.const():
+            raise ValueError(f"Unexpected Bit value: {value}")
+        return 1 if value is Bit.VCC else 0
+    elif not isinstance(value, Bits):
+        value = as_bits(value)
+    return int(value)
+
+
 class LUT(Generator2):
     """
     Generate a LUT containing entries of a generic Type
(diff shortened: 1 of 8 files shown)


magma/primitives/lut.py
```diff
@@ -0,0 +1,71 @@
...
+        self.coreir_genargs = {"N": n}
+        self.coreir_configargs = {"init": contents}
+
+
+def _to_int(value):
+    if isinstance(value, Bit):
+        if not value.const():
+            raise ValueError(f"Unexpected Bit value: {value}")
+        return 1 if value is Bit.VCC else 0
+    elif not isinstance(value, Bits):
```
rsetaluri, line 42: nit: can be `if`

magma/primitives/lut.py
```diff
@@ -0,0 +1,71 @@
...
+        n: nubmer of entries in the LUT
+
+        contents: the contents of the LUT (a tuple of length `n` containing
+                  elements of type T that are constant)
+        """
+        num_bits_per_entry = T.flat_length()
+        self.io = IO(I=In(Bits[clog2(n)]), O=Out(T))
+
+        contents = tuple(BitVector[num_bits_per_entry](_to_int(c)) for c in
+                         contents)
```
rsetaluri, line 65: nit

```suggestion
        contents = tuple(BitVector[num_bits_per_entry](_to_int(c))
                         for c in contents)
```

magma/primitives/lut.py
```diff
@@ -0,0 +1,71 @@
...
+        # create LUT for each bit in T
+        luts = []
+        for i in range(num_bits_per_entry):
+            luts.append(_CoreIRLUT(BitVector[n]([elem[i] for elem in
+                                                 contents]))())
```
rsetaluri, line 70: nit
```suggestion
            luts.append(_CoreIRLUT(BitVector[n]([elem[i]
                                                 for elem in contents]))())
```
or
```suggestion
            bv = BitVector[n]([elem[i] for elem in contents])
            luts.append(_CoreIRLUT(bv)())
```

magma/primitives/lut.py
```diff
@@ -0,0 +1,71 @@
...
+
+        n: nubmer of entries in the LUT
+
+        contents: the contents of the LUT (a tuple of length `n` containing
+                  elements of type T that are constant)
+        """
+        num_bits_per_entry = T.flat_length()
+        self.io = IO(I=In(Bits[clog2(n)]), O=Out(T))
+
+        contents = tuple(BitVector[num_bits_per_entry](_to_int(c)) for c in
```
rsetaluri, line 64: should we check the type of contents early here? afraid it might lead to some weird/unexpected behavior if wire w/o the check.

leonardt: The type check is done by the BitVector conversion, so if it doesn't satisfy that interface we'll get an error, I think that's a pretty reasonable choice.  We could be more strict (enforcing it's an instance of T), but I would imagine there's case when the user would want to pass a raw BitVector or int rather than constructing the full type.  So the current interface is "can it be converted to a BitVector of the required number of bits", if not we'll get a type error.  This is in the spirit of Python duck typing style (can it act as a BitVector[n]?)

rsetaluri: That makes sense, and I agree. I think I was originally confused as to why the type annotation for `contents` was a `magma.Tuple` but I realized thats `typing.Tuple` actually :+1:

PR #722, opened 2020-06-15
Add mux tests

leonardt:
Mux primitive was already implemented, moved it into the primitives sub-package and added some unit tests.

For https://github.com/phanrahan/magma/issues/598


magma/operators.py
```diff
@@ -12,7 +12,7 @@
 from .t import Type, In, Out
 from .tuple import Product
 from .wire import wire
-from magma.mux import Mux
+from magma.primitives import Mux
 
 
 def mux(I, S, **kwargs):
```
rsetaluri, line 18: not a big deal, but since we're re-org'ing, what's the best place for these 2 functions? They seem similar to `magma.reduce()` so perhaps move `mux()` to `primitives/mux.py` and `slice()` to `primitives/slice.py`. Also fine with them here.

PR #725, opened 2020-06-17
Return protocols in tuples

cdonovick:
fixes downstream bug https://github.com/cdonovick/peak/issues/176

### DIFF
--- a/magma/syntax/sequential.py
+++ b/magma/syntax/sequential.py
@@ -17,6 +17,13 @@ import itertools
 from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT, SymbolTable
 from ast_tools import gen_free_name
 
+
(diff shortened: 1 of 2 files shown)


magma/syntax/sequential.py
```diff
@@ -17,6 +17,13 @@
 from ast_tools.stack import _SKIP_FRAME_DEBUG_STMT, SymbolTable
 from ast_tools import gen_free_name
 
+
+def _get_magma_value(val):
+    if isinstance(val, MagmaProtocol):
+        return val._get_magma_value_()
+    else:
+        return val
```
rsetaluri, line 25:
```suggestion
def _get_magma_value(val):
    if isinstance(val, MagmaProtocol):
        return val._get_magma_value_()
    return val
```

PR #726, opened 2020-06-17
Add Memory(T) (RAM) primitive

leonardt:
Adds a generic memory primitive based on the coreir.mem

The init (and therefor ROMs) parameter is not supported by coreir yet (the verilog codegen for coreir.mem will ignore the init parameter).  Will follow up with a second PR for the ROM variant after adding the feature to CoreIR.  For now we will throw an exception if readonly is set to True.

### DIFF
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -1,0 +1,105 @@
+from hwtypes import BitVector
+
+from magma.bit import Bit
+from magma.bits import Bits
+from magma.bitutils import clog2
+from magma.clock import Clock, Enable
+from magma.clock_io import ClockIO
+from magma.conversions import from_bits, as_bits
+from magma.generator import Generator2
+from magma.interface import IO
+from magma.primitives.register import Register
+from magma.t import In, Out, Kind
+
+
+class CoreIRMemory(Generator2):
+    def __init__(self, depth, width, init=None):
+        self.name = f"coreir_mem{depth}x{width}"
+        addr_width = clog2(depth)
+        self.io = IO(
+            raddr=In(Bits[addr_width]),
+            rdata=Out(Bits[width]),
+            waddr=In(Bits[addr_width]),
+            wdata=In(Bits[width]),
+            clk=In(Clock),
+            wen=In(Enable)
+        )
+        self.verilog_name = "coreir_mem"
+        self.coreir_name = "mem"
+        self.coreir_lib = "coreir"
+        self.coreir_genargs = {"width": width, "depth": depth,
+                               "has_init": init is not None}
+        self.coreir_configargs = {}
+        if init is not None:
+            raise NotImplementedError("CoreIR's memory primitive does not "
+                                      "compile init arg to verilog yet")
+            # self.coreir_configargs["init"] = <convert to json arg>
+
+        def _simulate(self, value_store, state_store):
+            cur_clk = value_store.get_value(self.clk)
+
+            if not state_store:
+                state_store['mem'] = [
+                    BitVector[width](0) for _ in range(depth)
+                ]
+                state_store['prev_clk'] = cur_clk
+
+            prev_clk = state_store['prev_clk']
+            clk_edge = cur_clk and not prev_clk
+            rdata = value_store.get_value(self.rdata)
+
+            if clk_edge:
+                index = int(
+                    BitVector[addr_width](value_store.get_value(self.raddr))
+                )
+                rdata = state_store['mem'][index].as_bool_list()
+
+            if clk_edge:
+                if value_store.get_value(self.wen):
+                    index = int(
+                        BitVector[addr_width](value_store.get_value(self.waddr))
+                    )
+                    state_store['mem'][index] = BitVector[width](
+                        value_store.get_value(self.wdata)
+                    )
+
+            state_store['prev_clk'] = cur_clk
+            value_store.set_value(self.rdata, rdata)
+
+        self.simulate = _simulate
+
+
+class Memory(Generator2):
+    def __init__(self, height, T: Kind,
+                 read_latency: int = 0, read_only: bool = False):
+        if read_latency < 0:
+            raise ValueError("read_latency cannot be negative")
+        addr_width = clog2(height)
+        data_width = T.flat_length()
+        # For ROM, we use coreir's generic mem and tie off the write signals
+        self.io = IO(
+            RADDR=In(Bits[addr_width]),
+            RDATA=Out(T),
+        ) + ClockIO()
+        if not read_only:
+            self.io += IO(
+                WADDR=In(Bits[addr_width]),
+                WDATA=In(T),
+                WE=In(Enable)
+            )
+            waddr = self.io.WADDR
+            wdata = as_bits(self.io.WDATA)
+            wen = self.io.WE
+        else:
+            waddr = Bits[addr_width](0)
+            wdata = Bits[data_width](0)
+            wen = Bit(0)
+        coreir_mem = CoreIRMemory(height, data_width)()
+        coreir_mem.waddr @= waddr
+        coreir_mem.wdata @= wdata
+        coreir_mem.wen @= wen
+        coreir_mem.raddr @= self.io.RADDR
+        rdata = from_bits(T, coreir_mem.rdata)
+        for _ in range(read_latency):
+            rdata = Register(T)()(rdata)
+        self.io.RDATA @= rdata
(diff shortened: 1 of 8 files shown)


magma/primitives/memory.py
```diff
@@ -0,0 +1,105 @@
...
+            wdata = Bits[data_width](0)
+            wen = Bit(0)
+        coreir_mem = CoreIRMemory(height, data_width)()
+        coreir_mem.waddr @= waddr
+        coreir_mem.wdata @= wdata
+        coreir_mem.wen @= wen
+        coreir_mem.raddr @= self.io.RADDR
+        rdata = from_bits(T, coreir_mem.rdata)
+        for _ in range(read_latency):
+            rdata = Register(T)()(rdata)
```
David-Durst, line 104: Is there a way that the read_latency can be passed to CoreIR? I don't remember if this approach, a 0-latency (aka asynchronous) memory followed by a register in a separate module, will synthesize to the right result in Vivado, a BRAM with the right settings, rather than a lower clock rate solution, such as a BRAM followed by a register.

leonardt: Should be hard to implement, but not sure if @rdaly525 meant to keep it out of the memory primitive's parameter set?

leonardt: An alternative, but slightly more complex approach would be to run a pass that collects this pattern (register chain after memory read port) and packs it into a special module that generates the desired verilog (registers inside the memory module) so it's more of a technology specific code generation option

rsetaluri: I think a good pattern, which has already been followed for register primitives is to have keep the coreir memory "primitive" latency agnostic and have a "mantle" library generator which has a latency parameter. This can then be lowered into [coreir_mem |-> coreir_reg] or [fpga_bram(latency = 1)].

magma/primitives/memory.py
```diff
@@ -0,0 +1,105 @@
...
+                        BitVector[addr_width](value_store.get_value(self.waddr))
+                    )
+                    state_store['mem'][index] = BitVector[width](
+                        value_store.get_value(self.wdata)
+                    )
+
+            state_store['prev_clk'] = cur_clk
+            value_store.set_value(self.rdata, rdata)
+
+        self.simulate = _simulate
```
rsetaluri, line 69: I think we can just directly make `_simulate` a member function on the class? (or a classmethod maybe?).

magma/primitives/memory.py
rdaly525: The ideal approach in my mind is like @rsetaluri said, to keep the latency parameter in the mantle library generator. If the generator itself is staged, then we can lower it to appropriate verilog or fpga bram just using specialization on that generator. If things are not staged, then a pass to look for memories followed by registers would be needed.

magma/primitives/memory.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        self.verilog_name = "coreir_mem"
+        self.coreir_name = "mem"
+        self.coreir_lib = "coreir"
+        self.coreir_genargs = {"width": width, "depth": depth,
+                               "has_init": init is not None}
+        self.coreir_configargs = {}
+        if init is not None:
+            raise NotImplementedError("CoreIR's memory primitive does not "
+                                      "compile init arg to verilog yet")
```
rsetaluri, line 35:
```suggestion
            raise NotImplementedError("CoreIRMemory generator does not yet "
                                      "support 'init' argument")
```

magma/primitives/memory.py
rdaly525: That being said I do understand the utility of a set of coreir primitives that are parameterized by latency. If we do that approach, we should do a separate set of generators  from the base set (like coreir_w_latency.add)

magma/primitives/memory.py
```diff
@@ -0,0 +1,105 @@
...
+
+        self.simulate = _simulate
+
+
+class Memory(Generator2):
+    def __init__(self, height, T: Kind,
+                 read_latency: int = 0, read_only: bool = False):
+        if read_latency < 0:
+            raise ValueError("read_latency cannot be negative")
+        addr_width = clog2(height)
```
rsetaluri, line 77: should we check for power-of-2-ness here or in the coreir generator? alternatively, the interface could be to take in the "addressability" rather than the height, i.e. the size of `addr_width`. Think it depends on what kind of API people are used to using.

magma/primitives/memory.py
leonardt: I tried that before but it didn't work, perhaps class method will work, didn't look deeply into why we can't just `def ` it like a normal member

magma/primitives/memory.py
leonardt: Hmm, not quite though, it supports the argument, it just doesn't pass it through to verilog (so you can instantiate the CoreIR representation with the argument and it should compile to json file)

magma/primitives/memory.py
David-Durst: Where will the fpga_bram primitive suggested by @rsetaluri come from? CoreIR?

rsetaluri: @David-Durst I meant it more to generically say we should raise to mantle library, then lower to desired backends. I'm not clear specifically how to lower mem w/ latency to fpga.

Is there a specific verilog structure which you know would produce the desired fpga bram in the presence of latency? Or do we have to instance the fpga primitive directly?

magma/primitives/memory.py
leonardt: FWIW this is the interface provided by both the existing mantle generator and the coreir primitive (height/depth rather than address width).  I would imagine it's probably simpler for the user to decide the number of entries they'd like in their memory rather than the width of the address space.  For example, you may have a non-power of two number of entries, so your address space width has the requisite bit width (rounding up), but you wouldn't actually need the extra registers/storage elements for the invalid addresses.  With this in mind, height seems better than address space.  Then, I would imagine most designers would want to forgo any "runtime" check on the address validity, since that would incur a cost, and instead rely on something like an assertion/monitor to prove/test that the address is always valid.

magma/primitives/memory.py
rsetaluri: Ah I see, then I would revise to "Verilog code generation for CoreIRMemory does not yet support 'init'" (not a big deal either way though, I misinterpreted the point).

Also, consider making a warning rather than exception, since the "error" is quite downstream of this entry-point. I.e. you could generate json that gets used by mapper, and the issue of verilog code gen never materializes.

magma/primitives/memory.py
rdaly525: I think we could do either. Ideally we would just instantiate the appropriate BRAM primitive for the particular tech lib (which FPGA) but we could also go the route of just producing verilog that the FPGA tools can interpret as a BRAM.

leonardt:
Sounds good, I'll submit a CoreIR PR with a mantle memory generator and support for the init parameter to verilog, then update this to use those features.

magma/primitives/memory.py
David-Durst: In my experience, BRAMs are instanced using the right verilog in Vivado, not instantiation with a TCL call to create a BRAM IP. This is the approach used by chisel at https://www.chisel-lang.org/chisel3/memories.html#syncreadmem-sequentialsynchronous-read-sequentialsynchronous-write.

The verilog looks like https://github.com/David-Durst/embeddedHaskellAetherling/blob/79c5f6101f1c91ad2d9a180bd8a8ed6e39124b28/test/verilog_examples/aetherling_copies/big_real_32_conv2d/big_real_32_conv2d_2073600.v#L3974

leonardt: @David-Durst would you just be looking for a synchronous read (registered read port) module rather than a general latency parameter? That would be a simpler requirement to meet versus the more general latency parameter solution.

David-Durst: @leonardt Yep, that's exactly what I want.

leonardt: @rdaly525 @rsetaluri Looking at Chisel, it looks like they have two variants of their memory primitive, comb/async read and synch read.  Would you be opposed to adding a "synchronous_read" boolean flag to the memory primitive? This would significantly reduce the required code changes to achieve David's requirements (we can just augment the magma primitive to use the synch read parameter for latency 1, and chain registers for greater than 1.  and the verilog code generation can handle the synchronous read logic).

This is one possible middle ground between keeping all latency related logic in the mantle memory generator and adding latency as a parameter to the memory primitive.

Chisel's motivation for the two variants are:
> These SyncReadMems will likely be synthesized to technology SRAMs (as opposed to register banks).

> Writes to Mems are combinational/asynchronous-read, sequential/synchronous-write. These Mems will likely be synthesized to register banks, since most SRAMs in modern technologies (FPGA, ASIC) tend to no longer support combinational (asynchronous) reads.

So one argument for having sync mem be a primitive (or parameter of the primitive) is that it would be a common technology target.

PR #727, opened 2020-06-17
Add rom primitive

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/900

### DIFF
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -32,10 +32,7 @@ class CoreIRMemory(Generator2):
                                "has_init": init is not None}
         self.coreir_configargs = {}
         if init is not None:
-            init_bv = BitVector[width](init[0])
-            for i, elem in enumerate(init[1:]):
-                init_bv = init_bv.concat(BitVector[width](elem))
-            self.coreir_configargs["init"] = init_bv
+            self.coreir_configargs["init"] = [int(x) for x in init]
 
         def _simulate(self, value_store, state_store):
             cur_clk = value_store.get_value(self.clk)
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from hwtypes import BitVector
 
 from magma.bit import Bit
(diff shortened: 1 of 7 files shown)


magma/primitives/memory.py
```diff
@@ -71,7 +73,8 @@ def _simulate(self, value_store, state_store):
 
 class Memory(Generator2):
     def __init__(self, height, T: Kind,
-                 read_latency: int = 0, read_only: bool = False):
+                 read_latency: int = 0, read_only: bool = False, init:
```
rsetaluri, line 76: nit: prefer `init` on a newline entirely

magma/primitives/memory.py
```diff
@@ -31,9 +32,10 @@ def __init__(self, depth, width, init=None):
                                "has_init": init is not None}
         self.coreir_configargs = {}
         if init is not None:
-            raise NotImplementedError("CoreIR's memory primitive does not "
-                                      "compile init arg to verilog yet")
-            # self.coreir_configargs["init"] = <convert to json arg>
+            init_bv = BitVector[width](init[0])
+            for i, elem in enumerate(init[1:]):
+                init_bv = init_bv.concat(BitVector[width](elem))
+            self.coreir_configargs["init"] = init_bv
```
rsetaluri, line 38:
```suggestion
            functools.reduce(BitVector[width].concat, (BitVector[width](i) for i in init))
            init_bv = BitVector[width](init[0])
            for i, elem in enumerate(init[1:]):
                init_bv = init_bv.concat(BitVector[width](elem))
            self.coreir_configargs["init"] = init_bv
```
(import `functools` of course)

magma/primitives/memory.py
```diff
@@ -71,7 +70,8 @@ def _simulate(self, value_store, state_store):
 
 class Memory(Generator2):
     def __init__(self, height, T: Kind,
-                 read_latency: int = 0, read_only: bool = False):
+                 read_latency: int = 0, read_only: bool = False,
+                 init: Optional[tuple] = None):
```
rsetaluri, line 74: I never remember what the functional difference between `x: tuple` and `x: Tuple` is, but just making sure this is ok

leonardt: Hmm, they're really just cosmetic in this context so I don't think it matters, but Tuple is the actually Type (from typing) and tuple is the conversion operator (e.g. tuple(list(foo))), so if we were using something like mypy I think we would want Tuple

PR #729, opened 2020-06-21
Allow concat to use Bit

splhack:
`concat` doesn't allow to use `Bit`, it requires `bits` cast.

    bits0 = m.concat(m.bit(0), m.bit(1))
    
            def concat(*arrays):
    >               ts.extend(a.ts)
    E               AttributeError: 'Bit' object has no attribute 'ts'

    bits0 = m.concat(m.bits(m.bit(0)), m.bits(m.bit(1)))

whereas `repeat` supports `Bit`, also Verilog concatenate `{}` supports scalar.

### DIFF
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -192,7 +192,7 @@ def concat(*arrays):
         if isinstance(a, hwtypes.BitVector):
             ts.extend(a.bits())
         elif isinstance(a, Bit):
-            ts.extend(bits(a).ts)
+            ts.extend([a])
         else:
             ts.extend(a.ts)
     return array(ts)
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -191,6 +191,8 @@ def concat(*arrays):
     for a in arrays:
         if isinstance(a, hwtypes.BitVector):
             ts.extend(a.bits())
+        elif isinstance(a, Bit):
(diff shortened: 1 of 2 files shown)


leonardt:
CC @rsetaluri this seems fine for convenience, but is possibly related to your width inference proposal that's forthcoming, any opinions on merging this?

magma/conversions.py
```diff
@@ -191,6 +191,8 @@ def concat(*arrays):
     for a in arrays:
         if isinstance(a, hwtypes.BitVector):
             ts.extend(a.bits())
+        elif isinstance(a, Bit):
+            ts.extend(bits(a).ts)
```
rsetaluri, line 195:
```suggestion
        elif isinstance(a, Bit):
            ts.extend([a])
```

rsetaluri: @leonardt I think we can merge this for now. that will be a bigger change and will likely operate on completely separate types. Perhaps the long run is to only use those types, but this is a bit orthogonally useful

rsetaluri:
@leonardt I think we can merge this for now. that will be a bigger change and will likely operate on completely separate types. Perhaps the long run is to only use those types, but this is a bit orthogonally useful

rsetaluri:
Build is green, idk why it isn't showing here...merging

PR #734, opened 2020-06-22
Add option to add user-level namespace for coreir compilation

rsetaluri:
This passes introduces all user modules into a new user-specified namespace. Therefore, if specified (as a top-level compile option ala `m.compile("foo", foo, user_namespace="my_namespace")`), the output coreir will be structured as:

```
namespace my_namespace {
    Circuit foo;
    Circuit bar;
    ...
}

namespace some_external_lib {
    ...
}
```

This effectively replaces namespace `global` with namespace `my_namespace`. Furthermore, this will result in the verilog circuits having the namespace as a prefix (for free, per CoreIR verilog codegen spec).

I think this is a more robust and more "correct" way to achieve #714. It also has added benefits of avoiding name collisions even at the CoreIR level (#714 was to avoid name collisions at generated verilog level). As future work, we could allow specifying namespaces at a finer granularity, i.e. per python file or per verilog file. Then namespaces could even be a magma-level construct (right now it's just at CoreIR level). Further work could be to even support nested namespaces.

Issues/Concerns:
* Do we need to pass down the user_namespace to other parts of the compiler? `test_passes` was meant to make sure that the passes are run on the user namespace. I'm concerned because it seems to work even though we do not include that namespace when we specify on what namespace to run the pass (see inline comment).

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -46,9 +46,9 @@ _context_to_modules = {}
 
 class CoreIRBackend:
     def __init__(self, context=None):
-        self.__init(context)
+        self._init(context)
 
-    def __init(self, context):
+    def _init(self, context):
         singleton = CoreIRContextSingleton().get_instance()
         if context is None:
             context = singleton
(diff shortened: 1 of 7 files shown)


magma/backend/coreir_.py
```diff
@@ -46,9 +46,9 @@ def __init__(self):
 
 class CoreIRBackend:
     def __init__(self, context=None):
-        self.__init(context)
```
rsetaluri, line 49: unrelated style change

magma/backend/coreir_compiler.py
```diff
@@ -47,7 +55,8 @@ def compile(self):
         InsertCoreIRWires(self.main).run()
         InsertWrapCasts(self.main).run()
         backend = self.backend
-        backend.compile(self.main)
+        opts = _make_opts(backend, self.opts)
+        backend.compile(self.main, opts)
         backend.context.run_passes(self.passes, self.namespaces)
```
rsetaluri, line 60: do we need to add the user_namespace to `self.passes` here?

rsetaluri:
Yes, I closed that one

PR #736, opened 2020-06-24
Bugfixes for peak sequential2 support

leonardt:
Will add unit tests before merging

### DIFF
--- a/magma/protocol_type.py
+++ b/magma/protocol_type.py
@@ -31,6 +31,9 @@ class MagmaProtocolMeta(type):
     def flip(cls):
         return cls._flip_magma_()
 
+    def __len__(cls):
+        return len(cls._to_magma_())
+
 
 class MagmaProtocol(metaclass=MagmaProtocolMeta):
     @abc.abstractmethod
--- a/magma/bit.py
+++ b/magma/bit.py
@@ -169,13 +169,15 @@ class Bit(Digital, AbstractBit, metaclass=DigitalMeta):
 
     def ite(self, t_branch, f_branch):
         type_ = type(t_branch)
-        if type_ != type(f_branch):
+        if type_ is not type(f_branch):
             raise TypeError(f"ite expects same type for both branches: {type_} != {type(f_branch)}")
         if self.const():
             if self is type(self).VCC:
                 return t_branch
             assert self is type(self).GND
             return f_branch
+        if issubclass(type_, tuple):
+            return tuple(self.ite(t, f) for t, f in zip(t_branch, f_branch))
         # Note: coreir flips t/f cases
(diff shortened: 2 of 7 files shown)


magma/protocol_type.py
```diff
@@ -31,6 +31,9 @@ def _from_magma_value_(cls, val: Type):
     def flip(cls):
         return cls._flip_magma_()
 
+    def __len__(cls):
+        return len(cls._to_magma_())
+
```
leonardt, line 36: This was the simplest fix for getting the length of an arbitrary protocol, but it reserves the len method, an alternative is to update everywhere in the code where we call len(T) and replace it with len(T._to_magma_()) for protocols, this would avoid needing to reserve the method but add quite a bit of code changes, so I'll only do this if we anticipate the need for a custom len method

cdonovick: Hmmm, this isn't ideal but will work for now.  I think in the long term if we want to have good support for protocols and allow magma to extend its interface we will need to use a sort of "dereference" pattern where before interacting with objects they are first unpacked. like is done in sequential: https://github.com/phanrahan/magma/blob/master/magma/syntax/sequential.py#L21

I realize to will be annoying to always need to call protocol methods before use.  However, it would allow magma to seamlessly extend its interfaces without requiring updates from protocol types.   It would also allow protocols to not have the mixin methods.  Finally it would remove the need to actually inherit the protocol and instead allow structural subtyping similar to the bitvector protocol (I could see this being especially import for the metaclass).  

https://github.com/leonardt/hwtypes/blob/master/tests/test_bv_protocol.py#L38

We could create a decorator which could ease some cases:
```Python
def unpack(value_or_type):
    if isinstance(value_or_type, MagmaProtocol):
        return value_or_type._get_magma_value_()
    elif isinstance(value_or_type, MagmaProtocolMeta):
        return value_or_type._to_magma_()
    else:
        return value_or_type

def cast_to_magma(fn):
    @wraps(fn)
    def wrapper(*args, **kwargs):
        args = (unpack(x) for x in args)
        kwargs = {k: unpack(v) for k, v in kwargs.items()}
        return fn(*args, **kwargs)
    return wrapper
```

cdonovick: It should be noted I definitely can't add most `__dunders__` I don't currently use `__len__` but I do use others.

leonardt: Yea I agree that would be the most extensible, it's mainly a matter of laziness at this point ;) I think the proper long term solution is to have all these magma type level interactions go through the protocol (and have the protocol implemented by built-in types) so we won't need to do any of the "isinstance(..., protocol)" checks

Issue #731, opened 2020-06-22
sequential2: Register access shortcut?


splhack:
sequential can use register like this (`self.count` instead of `self.count.I` and `self.count.O`)

    @m.circuit.sequential(async_reset=False)
    class Test1:
        def __init__(self):
            self.count: m.SInt[16] = m.sint(0, 16)
        def __call__(self) -> m.SInt[16]:
            self.count = self.count + 1
            return self.count

sequential2 allows to write similar code, but `self.count.O + 1` instead of `self.count + 1`. can it be like sequential's? (or am I wrong about `__init__` Register declaration?)

    @m.sequential2()
    class Test2:
        def __init__(self):
            self.count = m.Register(T=m.SInt[16], init=m.sint(0, 16))()
        def __call__(self) -> m.SInt[16]:
            self.count = self.count.O + 1
            return self.count

leonardt:
This is a bug, working on a fix

leonardt:
Should be fixed by #737, thanks for testing this out, let us know if you run into more problems!

splhack:
ran into #739 😄

PR #737, opened 2020-06-24
Improve register getattr logic

leonardt:
Pulling into peak-seq2 to avoid merge conflicts (depends on some of those changes).

This augments the sequential2 getattr logic to support both the standard circuit pattern (calling syntax) and the setattr/getattr syntax for referencing the current value of the register and updating the next value.

For https://github.com/phanrahan/magma/issues/731

The way it works is:
When calling getattr, check if its an instance of the register primitive, if so, wrap it in a special magma protocol type that will implement a __call__ method that supports the circuit call style syntax, while otherwise behaving as the output of the register (i.e. register.O is the underlying magma value).  The main annoyance in the implementation is having to pass through things like operators to the underlying magma value, but I'm not sure if there's an alternative to doing this if we want to support both styles.

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -5,7 +5,9 @@ from ast_tools.passes import apply_ast_passes, ssa, if_to_phi, bool_to_bit
 
 from ..circuit import Circuit, IO
 from ..clock_io import ClockIO
-from ..t import Type
+from ..t import Type, Direction
+from ..protocol_type import MagmaProtocol, MagmaProtocolMeta
+from ..primitives.register import Register
 
 from .util import build_io_args, build_call_args, wire_call_result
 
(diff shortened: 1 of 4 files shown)


cdonovick:
This is a tricky problem.  In peak, I have the same issue with assembled adts where `AsmTuple[Bitvector[16], BitVector[8]](...)[0]` should return an "assembled" `BitVector[16]` (which could actually be meaningful if the assembled version had different endianness or something).   However because the forwarding operator to the leaf type would be super annoying I just return the type without assembling it.    But you can't do that here cause you want to support call.

Would it be easier to make register a descriptor instead of trying to do magic in `{get,set}attr`? 

https://docs.python.org/3/reference/datamodel.html#implementing-descriptors
https://docs.python.org/3/howto/descriptor.html

leonardt:
I'm not sure if descriptors will help, they could help organize the code so we don't have to generally modify get/set attr, but we still run into the issue of what "get" means (do we want the output, for the implicit syntax, or the circuit, for the call syntax)

PR #739, opened 2020-06-25
Adding sequential2 test

splhack:
Encountered type mismatch error on seq2

    if sel:
        self.count = self.count + 1

    E           TypeError: ite expects same type for both branches: Out(SInt[16]) != <class 'magma.syntax.sequential2._SequentialRegisterWrapperOut(SInt[16])'>


leonardt:
Pushed a fix in https://github.com/phanrahan/magma/pull/740

PR #740, opened 2020-06-25
Fix seq2 ite protocol bug

leonardt:
Fix for https://github.com/phanrahan/magma/pull/739

### DIFF
--- a/magma/protocol_type.py
+++ b/magma/protocol_type.py
@@ -86,3 +86,10 @@ class MagmaProtocol(metaclass=MagmaProtocolMeta):
 
     def __imatmul__(self, other):
         self.wire(other)
+
+
+def get_type(value):
+    T = type(value)
+    if issubclass(T, MagmaProtocol):
+        return T._to_magma_()
+    return T
--- a/magma/bit.py
+++ b/magma/bit.py
@@ -17,6 +17,7 @@ from magma.circuit import Circuit, coreir_port_mapping
 from magma.family import get_family
 from magma.interface import IO
 from magma.language_utils import primitive_to_python
+from magma.protocol_type import get_type
 
 
 def bit_cast(fn: tp.Callable[['Bit', 'Bit'], 'Bit']) -> \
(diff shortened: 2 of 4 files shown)


magma/protocol_type.py
```diff
@@ -86,3 +86,10 @@ def wire(self, other):
 
     def __imatmul__(self, other):
         self.wire(other)
+
+
+def get_type(value):
+    T = type(value)
+    if issubclass(T, MagmaProtocol):
+        return T._to_magma_()
```
cdonovick, line 94: Do you want to allow?

```Python
x : ProtocolT
assert ProtocolT._to_magma_() is m.Bits[8]
y : m.Bits[8]
c : m.Bit 
c.ite(x, y)
```
 I do not allow that with the `BitVectorProtocol`

leonardt: Hmm, I think that's the current use case.  Basically, the RegisterWrapper for sequential (implements the call logic) is an implementation of protocol, so this register can be used as a return value (output of any generic type), or you could say, return the register output + 1, or some constant, so for these cases to work, we need the RegisterWrapper (and protocol) to type check with the regular operation (addition result, or constant type)

cdonovick: I think we may want to do a design review at some point about this.  This seems like a very unsafe / surprising coercion.

```Python3
x: Protocol
y: Protocol 
c: Bit

if c:
   z = x
else:
   z = y

assert isinstance(z, Bits)
```

Is not a behaviour I would want.

leonardt: Yea I see the problem with that, let me see how we can workaround this.  Perhaps using an alternative pattern to Protocol would avoid this issue, allowing us to make the type check stricter.  The one issue with sub-classing Bits is that we can't do that in general (the type of a Register can be any arbitrary magma type T), so we would want to dynamically generate subclasses of T that implement the __call__ method, might be complicated but could work and avoid having to relax this type check.

cdonovick: Ahh I see.  That is tricky.  

Perhaps protocols could optionally define a field like `_coercible_` or something if they want to be treated like that?

```Python
def get_type(value):
    T = type(value)
    if issubclass(T, MagmaProtocol) and getattr(T, '_coercible_', False) :
        return T._to_magma_()
    return T
```

leonardt: Yea that could work, basically in this case the register wrapper just wants to behave exactly like it's wrapped typed, it just needs to add some special logic, so "casting" it to a T when it's used in other contexts is totally fine (e.g. the output of the ite), in general a protocol type may not want this, so we could have coercible false be the default

Issue #735, opened 2020-06-22
pytest slow


rsetaluri:
Finding that running full pytest is slow. Just did a quick profile. Most of the tests are ~1s, except for the following:

```
test_log.prof | 4.451536
test_jtag.prof | 9.470764
test_array_of_bits[coreir].prof | 9.599056
test_array_of_bits[coreir-verilog].prof | 9.834311
test_sdram.prof | 49.562007
```

The remainder of the suite takes ~40s cumulatively; these 5 tests take ~80s.

Is there anything we can do to get these down? We should try to chop any easy time off. But if they're fundamentally this long, that's fine, but we should mark them long tests or something, so that for local dev iteration we can skip them. Ideally, in general we could use https://docs.pytest.org/en/stable/example/markers.html to mark which tests are "core" or "regression" or "smoke." For example, I often just want to make sure that I didn't break any core feature.

@leonardt thoughts?

leonardt:
array of bits tests will likely be improved by porting to sequential2 and using the new register(T) primitive, so I would suggest porting that code over and deprecating the old functionality.  The log test may have to do with slowness with writing to disk, not sure what we could improve there, maybe using a named pipe? But we'd have to pass that through to the verilog TB some how.  Not sure about the SDRAM code gen off the top of my head, we'll have to profile that one.

leonardt:
We've addressed the issues for the major sources of problems, test_log hasn't been sped up but looking into it, most of the time is spent inside waiting for the verilator test benches to run.  I'm guessing they're slow because of IO (printing to stdout or to a file), and I'm not sure if we can really improve this without making the tests less useful. Should we close this for now? 4s isn't so bad compared to the others I would say.

leonardt:
Going to close this for now, but feel free to reopen if it's still an issue.

PR #743, opened 2020-06-26
Use single return statement in coroutine generated code

leonardt:
This was done to address the performance issue here: https://github.com/phanrahan/magma/issues/735

Basically, a majority of time was in dealing with some complex code generated for handling return statements inside if statements.  This simplifies the coroutine code to simply assign the return value and return it at the end.  This avoids the complexity of handling return statements (speeding up the code), but also improves the complexity of the generated hardware (simpler phi logic).  We can do this because we know that only one control flow path will be taken in the coroutine body (so we won't get clobbered by another assignment to the return values).

### DIFF
--- a/magma/syntax/coroutine.py
+++ b/magma/syntax/coroutine.py
@@ -182,6 +182,13 @@ def _coroutine(defn_env, fn):
                     ).rstrip()
                 yield_encoding_map[statement] = encoding
 
+    return_type = fn.__call__.__annotations__["return"]
+    if isinstance(return_type, tuple):
+        return_target = [f"__magma_coroutine_return_value_{i}"
+                         for i in range(len(return_type))]
+    else:
+        return_target = ["__magma_coroutine_return_value"]
+
     # add yield state register declaration
     yield_state_width = clog2(len(yield_encoding_map))
     if not manual_encoding:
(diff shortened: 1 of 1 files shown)


leonardt:
This brings the cumulative runtime of those tests down to 7.8s
```
❯ pytest tests/test_syntax/test_coroutine --profile
============================================== test session starts ===============================================
platform darwin -- Python 3.7.2, pytest-4.2.0, py-1.8.0, pluggy-0.9.0
rootdir: /Users/lenny/repos/magma, inifile: setup.cfg
plugins: profiling-1.7.0, parallel-0.0.9, cov-2.6.1, codestyle-2.0.0
collected 3 items

tests/test_syntax/test_coroutine/test_jtag.py .                                                            [ 33%]
tests/test_syntax/test_coroutine/test_sdram.py .                                                           [ 66%]
tests/test_syntax/test_coroutine/test_uart.py .                                                            [100%]
Profiling (from /Users/lenny/repos/magma/prof/combined.prof):
Thu Jun 25 18:14:10 2020    /Users/lenny/repos/magma/prof/combined.prof

         4552468 function calls (4206535 primitive calls) in 7.864 seconds
```

PR #744, opened 2020-06-26
Add sequential2 test with Product

splhack:
coreir aborts with this error.

> ERROR: magma_Bit_ite_Tuple_a=Out_Bit_inst0 21 is not a valid coreIR name!. Needs to be = ^[a-zA-Z_\-\$][a-zA-Z0-9_\-\$]*

pycoreir: 2.0.81
coreir: master (https://github.com/rdaly525/coreir/commit/9642814dd847dc82ff99c1c83670e6087dcd934a)
verilogAST: blacklist-slice-inline

### DIFF
--- a/tests/test_syntax/test_sequential2.py
+++ b/tests/test_syntax/test_sequential2.py
@@ -242,6 +242,7 @@ def test_sequential2_counter():
     assert check_files_equal(__file__, f"build/TestSequential2Counter.v",
                              f"gold/TestSequential2Counter.v")
 
+
 def test_sequential2_counter_if():
     @m.sequential2()
     class Test2:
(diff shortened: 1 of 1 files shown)


rsetaluri:
@splhack is this an example of an issue w/ a request for a fix? If so, can we move to issues with this snippet? And then we can add the test to the PR that fixes the issue.

splhack:
@rsetaluri yep, just minimal code to repro the issue.

leonardt:
FWIW, the PR model of submitting this repro code is useful.  I can check out the branch locally, push a fix, verify it works with the test case, then pull the fix branch in.  Basically saves me having to copy the snippet from the issue to the test suite.  But, perhaps opening a corresponding issue would be useful for tracking the problem

leonardt:
Closing in favor of 745 which has the fix.

PR #747, opened 2020-06-26
Add helpful warning message for working with undirected values

leonardt:
This enables a warning message that should've been on in the first place (the error guard was too conservative).  Basically, there's cases when people use `m.wire` that don't respect the `m.wire(driver, drivee)` pattern when working with undirected values.  This ensures that there is a useful error message reported (was not the case before).  For now, we tell the user that they are overriding the previous driver, which should be quick to indicate that their wiring of the undirected values was not working as intended.

We may consider a way to trace back to the original wiring command and report some suggestion such as "maybe reverse this wiring statement?", but I think for now at least you get something useful.  Note that this is a more rare case, because when using `x @= y` to indicate y drives x, it's a more natural/common pattern to handle the wiring relationship between undirected values.

### DIFF
--- a/tests/test_wire/test_undirected_error.py
+++ b/tests/test_wire/test_undirected_error.py
@@ -1,20 +1,19 @@
 import magma as m
-from magma.testing.utils import has_warning
+from magma.testing.utils import has_warning, magma_debug_section
 
 
 def test_undirected_value_warning_1(caplog):
-    m.config.set_debug_mode(True)
-    class Main(m.Circuit):
-        io = m.IO(I=m.In(m.Bits[5]), O=m.Out(m.Bits[5]))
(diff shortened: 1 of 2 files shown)


tests/test_wire/test_undirected_error.py
```diff
@@ -0,0 +1,42 @@
+import magma as m
+from magma.testing.utils import has_warning
+
+
+def test_undirected_value_warning_1(caplog):
+    m.config.set_debug_mode(True)
```
rsetaluri, line 6: we might want to use `with magma_debug_section(): ...` here (from `magma.testing.utils`) so as to not keep debug mode on for other tests unexpectedly. Not sure if we already turn it off before each test elsewhere (e.g. conftest).

PR #749, opened 2020-06-28
coroutine2

splhack:
wanted to see how coroutine works with sequential2😄

### DIFF
--- a/tests/test_syntax/test_coroutine/test_sdram.py
+++ b/tests/test_syntax/test_coroutine/test_sdram.py
@@ -94,15 +94,12 @@ def test_sdram():
     CMD_READ = m.bits(0b10101001, 8)
     CMD_WRIT = m.bits(0b10100001, 8)
 
-    @m.syntax.coroutine
+    @m.coroutine(manual_encoding=True, reset_type=m.AsyncResetN)
     class SDRAMController:
-        _manual_encoding_ = True
-        _reset_type_ = m.AsyncResetN
-
         def __init__(self):
-            self.yield_state: m.Bits[5] = INIT_NOP1
-            self.command: m.Bits[8] = CMD_NOP
-            self.i: m.UInt[4] = m.uint(15, 4)
+            self.yield_state = m.Register(T=m.Bits[5], init=INIT_NOP1)()
+            self.command = m.Register(T=m.Bits[8], init=CMD_NOP)()
+            self.i = m.Register(T=m.UInt[4], init=m.uint(15, 4))()
 
         def __call__(self, refresh_cnt: m.UInt[10], rd_enable: m.Bit, wr_enable: m.Bit) -> (m.Bits[5], m.Bits[8]):
             yield from self.init()
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -96,6 +96,7 @@ from .log import info, debug, warning, error
 
 from .syntax.sequential2 import sequential2
 from .syntax.combinational2 import combinational2
+from .syntax.coroutine import coroutine
 
 from magma.primitives import (LUT, Mux, mux, Register, get_slice, set_slice,
                               slice, reduce, Memory)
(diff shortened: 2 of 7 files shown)


splhack:
yes, I'm happy to finish it up! To pass the original tests, need to deal with the last blocker, `self.attr.prev()`.

- One idea is, migrating it to ssa https://github.com/leonardt/ast_tools/pull/52
- if we may not want to do it, adding pre-pass to convert `prev()` with initial value.

original code

    def __call__(self, val):
        self.cnt = self.cnt + val
        return self.cnt.prev()

prev conversion pre-pass in coroutine

    def __call__(self, val):
        __magma_coroutine_init_cnt = self.cnt   # inserted by this conversion pass
        self.cnt = self.cnt + val
        return __magma_coroutine_init_cnt  # replaced with the init value


----
here are nice-to-have things with coroutine/sequential2 I've noticed

implicit `reset_type` setting. It can be automated by coroutine/sequential2 conversion pass.

    @m.coroutine
    class SDRAMController:
        _reset_type_ = m.AsyncResetN

        def __init__(self):
            self.command = m.Register(T=m.Bits[8], init=CMD_NOP, reset_type=m.AsyncResetN)()
                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^

    @m.coroutine
    class UART:
        def __init__(self):
            self.message = m.Register(T=m.Bits[8], init=0, reset_type=m.AsyncReset)()
                                                           ^^^^^^^^^^^^^^^^^^^^^^^

implicit type conversion. IIRC, the following code may cause type conflict between `int` vs `magma type (like Bits)`. The code already know the type of member variable (`self.cnt`), so it can cast the type of `1` to `self.cnt` type.

    if sel:
        self.cnt = 1
    else:
        self.cnt = self.cnt + 1

    test_sdram.py example
    - self.i = 7
    + self.i = m.uint(7, 4)

leonardt:
See https://github.com/leonardt/ast_tools/pull/52/files#r447218751, I think we should do it as a pass in magma or if possible figure out a way to do without a pass.

splhack:
ok, figured out the way without `inspect_enclosing_env` nor `compile_function_to_file`.

1. coroutine decorator builds `__call__` method AST tree (`call_method`) with 
https://github.com/phanrahan/magma/blob/c0a50856f93c2732e392f87920cb0eff0e99ea70/magma/syntax/coroutine.py#L145-L273
2. create a `ast_tools.passes.Pass` class to replace `__call__` method `tree` with 1's `call_method`
3. call `sequential2` with 2 in `pre_passes`

splhack:
depends on https://github.com/phanrahan/magma/pull/756

splhack:
depends on https://github.com/phanrahan/magma/pull/757

- adopt register's `reset_type`, `has_enable`, and `reset_priority`
- convert `manual_encoding` and `reset_type` into `coroutine` decorator arguments

tests/test_syntax/test_coroutine/test_sdram.py
```diff
@@ -138,7 +135,7 @@ def init(self, refresh_cnt: m.UInt[10], rd_enable: m.Bit, wr_enable: m.Bit) -> (
             self.yield_state = INIT_REF1
             yield self.yield_state.prev(), self.command.prev()
             # for _ in range(7, -1, -1):
-            self.i = 7
+            self.i = m.uint(7, 4)
```
leonardt, line 138: I'm guessing with https://github.com/phanrahan/magma/pull/761, we'll be able to revert these changes? Not a big deal, but would make it a little easier to read/write

splhack: yes exactly

splhack:
rebased onto https://github.com/phanrahan/magma/pull/761 to clean up test_sdram.py

leonardt:
Thanks for working on this! Glad this was moved to sequential2 and we could clean up some of the logic.

PR #751, opened 2020-06-29
Unwrap protocol types in get_source

rsetaluri:


magma/backend/coreir_transformer.py
```diff
@@ -21,17 +21,49 @@
...
+            value[i].name.array is value[i - 1].name.array and
+            value[i].name.index == value[i - 1].name.index + 1
+        ):
+            drivers.append(value[start_idx:i])
+            start_idx = i
+    drivers.append(value[start_idx:])
+    return drivers
+
+
+def _unwrap(x):
```
cdonovick, line 59: Maybe this should be in some sort of util file or in the protocol_type module it self.  There is an effectively equivalent function here
https://github.com/phanrahan/magma/blob/master/magma/syntax/sequential.py#L21

magma/backend/coreir_transformer.py
```diff
@@ -21,17 +21,49 @@
...
+    if isinstance(x, MagmaProtocol):
+        return x._get_magma_value_()
+    if isisstance(x, MagmaProtocolMeta)
```
cdonovick, line 62: :

magma/backend/coreir_transformer.py
rsetaluri: when we do a nuclear "deprotocol" pass I think that will consume all of this

cdonovick: Im not sure.  Wouldn't the use in sequential still be necessary?

rsetaluri: yeah that's fair. think it would be a good idea to separate out. already merged to accelerate unblocking, so I will come back and make a PR for this.

cdonovick: Oh for sure I didn't think the PR should be blocked on this.  Just a thought

Issue #753, opened 2020-06-30
Setting a slice of a bit vector (feature request)


kavyasreedhar:
@leonardt @cdonovick 

Is it possible to add setting a part of a register instead of having to specify the whole signal? I was talking with Caleb about a NotImplementedError I was getting for set_item today. It would be nice if I could do something like

self.mem[x : x + 16] = data_in 

(where self.mem is a BitVector) instead of having to do 

self.mem = concat(self.mem[ : x], data_in, self.mem[ x + 16 : ]) 

because as Caleb pointed out, I would also have to watch out for the case when I have a 0 length slice which would cause errors, so this code would actually need to be converted into an if statement where I check if (in this example) x is 0 or the max address and separate out my code into 3 separate cases with different concat scenarios. Caleb suggested asking for the feature on "SequentialRegister", if that's possible? Thanks!

leonardt:
This should be possible, but to clarify, is x here a python constant value? Or is it a dynamic index?

kavyasreedhar:
x is an input to the module (like addr). The actual code I had if that's helpful was  self.mem[addr * data_width : (addr + 1) * data_width] = data_in, where addr is an input, data_width is a parameter, and data_in is also an input.

cdonovick:
@leonardt I think we could special case dynamic indices for `getitem :: BV[n] -> BV[k] -> Bit where n == 1 << k`  Its a total function that can be defined with shifts so it shouldn't be a problem.   Similar for setitem.

cdonovick:
Obviously magma could something smarter if it wants to

cdonovick:
This wouldn't address how this precise scenario where a slice being extracted but I think we could something similar.

leonardt:
Right now magma provides an `m.slice` operator for these dynamic offset, constant width slices, see the documentation here https://github.com/phanrahan/magma/blob/c0a50856f93c2732e392f87920cb0eff0e99ea70/magma/primitives/slice.py#L5, I think this would work for this use case?

[...]

PR #755, opened 2020-07-01
Add set_slice operator

leonardt:
For https://github.com/phanrahan/magma/issues/753

### DIFF
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -98,4 +98,4 @@ from .syntax.sequential2 import sequential2
 from .syntax.combinational2 import combinational2
 
 from magma.primitives import (LUT, Mux, mux, Register, get_slice, set_slice,
-                              reduce, Memory)
+                              slice, reduce, Memory)
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -97,4 +97,5 @@ from .log import info, debug, warning, error
 from .syntax.sequential2 import sequential2
 from .syntax.combinational2 import combinational2
 
-from magma.primitives import LUT, Mux, mux, Register, slice, reduce, Memory
+from magma.primitives import LUT, Mux, mux, Register, get_slice, set_slice, \
+    reduce, Memory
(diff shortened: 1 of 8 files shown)


leonardt:
The proposed slice operator interface: https://github.com/phanrahan/magma/pull/755/files#diff-eec6d6b4f4c6633a0e9af0ad4c2f53b6R316-R325

magma/__init__.py
```diff
@@ -97,4 +97,5 @@ def set_mantle_target(t):
 from .syntax.sequential2 import sequential2
 from .syntax.combinational2 import combinational2
 
-from magma.primitives import LUT, Mux, mux, Register, slice, reduce, Memory
+from magma.primitives import LUT, Mux, mux, Register, get_slice, set_slice, \
```
rsetaluri, line 100: prefer parentheses over `\` break

magma/primitives/slice.py
```diff
@@ -1,8 +1,12 @@
...
-from magma.conversions import array
+from magma.bit import Bit
+from magma.bits import Bits, UInt
+from magma.bitutils import clog2
+from magma.t import Direction
+from magma.conversions import array, zext, uint
+from magma.protocol_type import MagmaProtocol
 
 
-def slice(value: Bits, start: Bits, width: int):
```
rsetaluri, line 5: are we sure there were no external users of this function? can we provide a deprecated backward-compatible fn. for `slice` in case?

rsetaluri:
can you explain the `set_slice` semantics a bit?

leonardt:
See #753 for the motivation.

set_slice takes a value, and returns a new value with a dynamic slice of the value replaced with a new value.  It basically emulates setting a constant width slice, e.g. `self.mem[x : x + 16] = data_in`, but restricts the operation to a simpler to handler case (as discussed in the issue).

We use a functional style operator (take a value, return a new "mutated" value) because it's easier to implement and restrict the semantics (compared to a more general setitem/getitem slice operator which brings up many issues)

PR #756, opened 2020-07-01
sequential2: support self.attr.prev()

splhack:
def __call__(self) ->m.UInt[3]:
        self.cnt = self.cnt + 1
        return self.cnt.prev()

    def __call__(self) ->m.UInt[3]:
        __magma_sequential2_cnt = self.cnt
        self.cnt = self.cnt + 1
        return __magma_sequential2_cnt

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -1,7 +1,9 @@
-from typing import Optional
+import ast
+from typing import Optional, MutableMapping
 
 from ast_tools.stack import SymbolTable
-from ast_tools.passes import apply_ast_passes, ssa, if_to_phi, bool_to_bit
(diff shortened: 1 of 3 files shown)


magma/syntax/sequential2.py
```diff
@@ -197,6 +199,56 @@ def _seq_phi(s, t, f):
     return s.ite(t, f)
 
 
+class PrevReplacer(ast.NodeTransformer):
+    """Replace self.attr.prev() with __magma_sequential2_attr"""
+    def __init__(self, env):
+        self.env = env
```
leonardt, line 205: Looks like `env` is unused, can we remove this attribute/constructor arg?

splhack: sure. also renamed private classes with `_` prefix.

PR #757, opened 2020-07-01
sequential2: support register params

splhack:
same as sequential, Register reset_type is set implicitly

    self.yield_state = m.Register(T=m.Bits[5], init=INIT_NOP1)()

    self.yield_state = m.Register(T=m.Bits[5], init=INIT_NOP1, reset_type=m.AsyncResetN)()

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -6,7 +6,8 @@ from ast_tools.passes import (PASS_ARGS_T, Pass, apply_ast_passes, ssa,
                               if_to_phi, bool_to_bit)
 
 from ..circuit import Circuit, IO
-from ..clock_io import ClockIO
+from ..clock import AbstractReset
+from ..clock_io import ClockIO, gen_clock_io
 from ..t import Type, Direction
 from ..protocol_type import MagmaProtocol, MagmaProtocolMeta
 from ..primitives.register import Register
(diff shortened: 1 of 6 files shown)


splhack:
- sequential2 `m.sequential2(has_async_reset=True)`
- coroutine `_reset_type_ = m.AsyncReset`
- Register `m.Register(reset_type=m.AsyncReset)`

for the consistency, sequential2/coroutine can use args like this?

    m.sequential2(reset_type=m.AsyncReset)
    m.coroutine(reset_type=m.AsyncReset, manual_encoding=True)

magma/syntax/sequential2.py
```diff
@@ -205,11 +297,30 @@ def sequential2(pre_passes=[], post_passes=[],
...
+    if ("has_async_reset" in clock_io_kwargs and
+            clock_io_kwargs["has_async_reset"]):
+        reset_type = ast.parse("m.AsyncReset").body[0].value
+    elif ("has_async_resetn" in clock_io_kwargs and
+            clock_io_kwargs["has_async_resetn"]):
+        reset_type = ast.parse("m.AsyncResetN").body[0].value
+    elif "has_reset" in clock_io_kwargs and clock_io_kwargs["has_reset"]:
+        reset_type = ast.parse("m.Reset").body[0].value
+    elif "has_resetn" in clock_io_kwargs and clock_io_kwargs["has_resetn"]:
+        reset_type = ast.parse("m.ResetN").body[0].value
```
leonardt, line 316: Yes, I think it would be good to update the sequential2 interface to support the reset_type parameter for consistency with the new register interface.

magma/syntax/sequential2.py
```diff
@@ -197,6 +199,96 @@ def _seq_phi(s, t, f):
...
+        self.reset_type = reset_type
+
+    def visit_Call(self, node: ast.Call):
+        # find m.Register with no reset_type
+        if (
+            isinstance(node.func, ast.Attribute)
+            and isinstance(node.func.value, ast.Name)
+            and node.func.value.id == "m"
+            and node.func.attr == "Register"
+        ):
```
leonardt, line 216: This should work in the simple case, but there might be some other strange cases that we may want to support, for example, the user could say `from magma import Register` so there's no `m` prefix.  One thing we could do is try to serialize and evaluate the call func inside `env` and see if it's equal to m.Register, e.g. `eval(astor.to_source(node.func).rstrip()) == m.Register`

leonardt: This won't cover everything though, but I'm not sure if we can support something like:
```
def __init__(self, ...):
    Foo = m.Register
    self.x = Foo(...)
```

leonardt: Now that I think about it, maybe it's just simpler to **force** users of sequential2 to declare registers with `m.Register`, then this code should work and it would vastly simplify the task for us.

splhack:
replace sequential2 `clock_io_kwargs` with register's `reset_type`, `has_enable`, and `reset_priority`

PR #760, opened 2020-07-01
Add notimplemented logic to bv coerce

leonardt:
This allows other constructs to handle the operator if we can't coerce the other type to a Bits


cdonovick:
I think generally a bare except clause is a bad idea.  Could we just convert type errors to NotImplemented?

PR #761, opened 2020-07-02
seq2 ite: implicit int conversion

splhack:
convert int to the other type implicitly since test_sdram.py with sequential2

            self.i = 1

causes `ite expects same type for both branches` error. requires

            self.i = m.uint(1, 4)


leonardt:
CC @cdonovick , this seems like a reasonable choice.  Basically this allows `int` values to be coerced into the type of the other branch if possible. Thoughts?

cdonovick:
This seems reasonable as ints are generally coercible.

cdonovick:
I think the  only issue is if there ints on both sides of an if.

Issue #754, opened 2020-07-01
[RFC] SmartBits


rsetaluri:
# SmartBits

## Problem

Magma users have frequently complained about the ergonomics (specifically, lack-there-of) of the strict type-system. In particular, where as Verilog provides implicit width conversions for arithmetic and logic operators, Magma has much stricter rules around operand widths. This results in a lot of verbosity for even simple operations. **Note** that this issue is two-fold, as verbosity appears both at the python (user) level, and in the generated Verilog.

For example, to simply do an addition of mismatched widths in Verilog we can happily write:

```
wire [7:0] A;
wire [15:0] B;
wire [23:0] C;

assign C = A + B;
```

Similar code in Magma would be:

```
A = m.UInt[8]()
B = m.UInt[16]()
C = m.UInt[24]()
C @= A + B
```

but this results int the error: `hwtypes.bit_vector_abc.InconsistentSizeError: Inconsistent size`, since the strictness of the width rules disallows addition of an 8-bit value with a 16-bit value. The following is working magma code, that produces Verilog semantically equivalent to the hand-written Verilog above:

```
A = m.UInt[8]()
B = m.UInt[16]()
C = m.UInt[24]()
C @= m.zext(A, 16) + m.zext(B, 8)
# The following is also working code, but has different semantics:
#   C @= m.zext(m.zext(A, 8) + B, 8)
```

This by itself may not seem too egregiously cumbersome, but for a complicated datapath with several cascaded operations, this becomes prohibitively verbose and unreadable. For example, imagine a complex operation like `Z = ((X * Y) + U) * W`; since the verbosity increases linearly with the number of operands and operations in an expression, we may have up to 5 extension operators in this simple expression. Furthermore, beyond just the un-readability, there is extra burden on the programmer to do the arithmetic necessary to figure out the zero-extension argument (e.g. in the above example 24 - 8 = 16, and 24 - 16 = 8).

### This problem seems self-inflicted?

[...]

PR #767, opened 2020-07-10
SmartBits prototype

rsetaluri:
See #754.

Still WIP. @leonardt @cdonovick early feedback welcome. Especially re: usage of magma protocol, and tree traversal algorithms.


magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+    @abc.abstractmethod
+    def resolve(self):
+        raise NotImplementedError()
+
+    @property
+    def determination(self):
+        return self._determination
+
+    def __add__(self, other):
+        assert isinstance(other, _SmartExpr)
```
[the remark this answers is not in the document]
rsetaluri: Good pointer, will convert to that.

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+        if self.assignee is not None:
+            max_width = max(max_width, len(self.assignee.bits))
+        return max_width
+
+
+class _SmartExprMeta(MagmaProtocolMeta):
+    pass
+
+
+class _SmartOpExprMeta(_SmartExprMeta):
```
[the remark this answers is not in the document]
rsetaluri: good point, could get rid of these, and just point straight to `_SmartExprMeta`

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def __len__(self):
+        return len(type(self)._T)
+
+    def force_width(self, width):
+        diff = len(self) - width
+        if diff == 0:
+            return self
+        value = self._get_magma_value_()
+        value = value.zext(-diff) if diff < 0 else value[:-diff]
```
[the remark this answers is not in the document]
rsetaluri: signed-ness is still WIP.

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def __invert__(self):
+        return _SmartUnaryOpExpr(operator.invert, self)
+
+
+class _SmartOpExpr(_SmartExpr, metaclass=_SmartOpExprMeta):
+    def __init__(self, determination, op, *args):
+        super().__init__(determination)
+        self._op = op
+        self._args = tuple(args)
```
leonardt, line 90: Minor, but I think it might be safe to just assign args without the tuple call?

```
Python 3.7.2 (default, Dec 29 2018, 00:00:04)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def f(*args):
...     print(type(args))
...
>>> f(1, 2, 3)
<class 'tuple'>
```

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+    @property
+    def args(self):
+        return self._args
+
+    @property
+    def op(self):
+        return self._op
+
+    def _update(self, *args):
+        self._args = tuple(args)
```
leonardt, line 101: See above re: *args being a tuple already

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def __str__(self):
+        if inspect.isbuiltin(self._op):
+            op_str = self._op.__name__
+        else:
+            op_str = str(self._op)
+        args = ", ".join(str(arg) for arg in self._args)
+        return f"{str(op_str)}({args})"
```
leonardt, line 109: Minor Do we need the extra str call on op_str here? Could either call str in the first if body to match the else, or remove the else from the else body?

magma/smart/smart_bits.py
cdonovick: yes `*args` is always a tuple

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+                         op, loperand, roperand)
+
+    def resolve(self, context):
+        context = Context(None, self)
+        for arg in self._args:
+            arg.resolve(context)
+        to_width = max(arg._width_ for arg in self._args)
+        args = (_extend_if_needed(arg, to_width) for arg in self._args)
+        self._update(*args)
+        self._width_ = 1
```
leonardt, line 173: Seems like there might be some way to share some of the common code with _SmartBinaryOpExpr.resolve

tests/test_smart_bits.py
```diff
@@ -0,0 +1,79 @@
...
+        print (io.I0)
+        print ("=======================")
+
+        io.O @= x
+        io.O2 @= y
+        io.O3 @= io.I0
+
+    print (repr(_Foo))
+
+    m.compile("Foo", _Foo, output="coreir-verilog", inline=True)
```
leonardt, line 31: It could imagine it being useful to have "pretty print" (maybe an extension of default circuit repr) that annotates the final widths that were determined

leonardt: (which would also be useful for testing)

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class Context:
+    def __init__(self, assignee, root):
+        self.assignee = assignee
+        self.root = root
+
+    def max_width(self):
+
+        def _visit(node):
```
rsetaluri, line 28: TODO: refactor this into visitor pattern

magma/smart/smart_bits.py
rsetaluri: I actually racked my brain on this for a while. Couldn't find a good abstraction, though the structure of the two are very similar. Probably implies that the overall pattern smells a bit (i.e. should be revisited). Open to suggestions if a refactoring is evident to you.

tests/test_smart_bits.py
rsetaluri: Makes sense. Will propose something in a upcoming commit.

magma/smart/smart_bits.py
rsetaluri: Not needed since we don't use this much.

tests/test_smart_bits.py
rsetaluri: Done, attached `_smart_expr_` to the smart bits that get wired to. See `test_complex`.

rsetaluri:
@leonardt this version of smart bits is ready for final review. There are a few things missing, but I think getting users will be more useful than hammering out all the details right now

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+There are two smart type constructors:
+* `SmartBits[bit_width, signed=False]`: returns a SmartBits type with bit-width `bit_width` and signedness `sign`. `sign` defaults to `False`.
+* `SmartBit` which is a literal alias of `SmartBits[1]`.
+
+Expressions built on top of these base smart types are of an intermediate type `SmartExpr`, which can only (a) be used in another expression, such as `dbl = expr + expr`, or (b) be on the right hand side of a wiring assignment. Besides these two operations (and calling `str()` for debugging), these intermediate objects are not meant to be introspected or operated on by the user.
+
+## Operators
+Arithmetic, bit-wise logical, comparison, shift, bit-wise reduction, and concat operators are defined for all smart types and intermediates. All operators are implemented as python-style infix operators (e.g. `a + b`), except for the following:
+* Concat: the function `concat` is a free, variadic function; e.g. `long = concat(x, y, z)`
+* Bit-wise reduce: bit-wise reductions are defined as a member function `reduce`, which takes the reduction operator as an argument; e.g. `short = x.reduce(operator.or_)`
```
leonardt, line 30: We could add `m.reduce()` support for this pattern, but this should work for now

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+
+See the end of the document for a table summarizing all operators.
+
+## Inference
+We first define a *partial expression* to be any (potentially compound) expression with smart operands, without wiring, and we define a *complete expression* to be any wiring statement where the left hand side is a base smart type and the right hand side is a partial expression. Every complete expression is dynamically resolved using a scheme to infer the bit-width and signedness of each operand. This is inspired from the [Verilog standard](https://www.eg.bucknell.edu/~csci320/2016-fall/wp-content/uploads/2015/08/verilog-std-1364-2005.pdf) (see Sec. 5.2).
+
+In general, every partial expression appearing in a complete expression is either *self-determined* or *context-determined*. This distinguishes expressions which require the entirety of the complete expression in-order to do inference, or only the partial expression. Each operator may have further specialized inference rules.
```
leonardt, line 37: Can a partial expression be used more than once? (e.g. wired to two different endpoints)

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Bit-width
+The rule of thumb for context-determined expressions is that all it's operands get extended to the **longest** operand appearing in the complete expression (including the LHS of the assignment). On the other hand, the scope for self-determined expressions is only the partial expression itself. Note that since a context-determined expression can appear within a self-determined expression, each self-determined expression must still be procssed by the inference step, except that the context is limited to just the expression itself, and in particular, the LHS of the assignment is excluded.
```
leonardt, line 40: Typo: "it's operands" -> "its operands"

leonardt: Also, "procssed" -> "processed"

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+
+The bit-width of any partial expression's result is determined by the specific operator (see table below). For example the result of addition has bit-width equal to the maximum of all of it's operand bit-widths, where as the result of comparison always has bit-width equal to 1.
+
+### Signedness
+The signedness of operands can both determine the semantics of operators, as well as how operands are extended. For example, signed-comparison has different semantics than unsigned-comparison, and operands can either be sign-extended or zero-extended (during bit-width inference).
+
+The rule of thumb for both self-determined and context-determined expressions is that all it's operands must be signed for the operator to be interpreted as signed (and therefore produce a signed output). There are the following exceptions:
```
leonardt, line 47: Same typo: "it's operands" -> "its operands"

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+
+## TODOs
+* Integer operands
+* Explicit extend operator
+* SmartBits select operator
+* Signedness for SmartBit
+
+## Further reading
+* [Bit-width and signedness gotchas](http://www.deepchip.com/items/0466-05.html)
+
```
leonardt, line 71: Not for this PR, but in the future, it would be interesting to add some sort of context manager pattern for entering "SmartBits" syntax regions, where any incoming variables are converted to smartbits (this would make it convenient for interfaces with standard bits, but easily converting inputs into smart bits without having to explicitly case all of them by name).

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,526 @@
...
+        }
+
+        def __init__(self, op):
+            cls = _SmartReductionOpExpr._ReductionOp
+            if op not in cls._OP_TO_NAME:
+                raise ValueError(f"Reduction operator {op} not supported")
+            self._op = op
+
+        def __call__(self, operand):
+            return reduce(self._op, operand)
```
leonardt, line 241: You may want to call `m.reduce` here since it will generate the coreir bitwise reduce primitive (which in turn generates something like `&x` in verilog)

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,526 @@
...
+        self._update(*args)
+        self._width_ = to_width
+        self._signed_ = all(arg._signed_ for arg in self._args)
+
+    def eval(self):
+        args = self._eval_args()
+        if self._signed_:
+            args = (sint(arg) for arg in args)
+        else:
+            args = (uint(arg) for arg in args)
```
leonardt, line 336: The mapping of `_signed_` to `uint` or `sint` can probably be wrapped up in a helper function since it's the same for all these `eval` methods.

leonardt: Also it looks like `_SmartReductionOpExpr`, `_SmartNAryContextualOpExr`, and `_SmartShiftOpExpr` all have the same `eval` function that could be reused.  

`_SmartComparisonOpExpr` looks similar, but `signed_args` is used instead of `_signed_` (so could share some helper function like `_eval(self._signed_)` and `_eval(signed_args)`

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,526 @@
...
+
+    def eval(self):
+        args = self._eval_args()
+        args = [uint(arg) for arg in args]
+        return uint(self.op(*args))
+
+
+def concat(*args):
+    if not all(isinstance(arg, _SmartExpr) for arg in args):
+        return NotImplemented
```
leonardt, line 368: Maybe this should raise NotImplementedError instead?

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+a @= expr
+```
+
+Note that using raw magma types (i.e. `Bits`, `UInt`, `SInt`), we would have to do several explicit extensions and explicit promotion to `UInt` (for example, because addition is not defined on plain `Bits`). The smart types allow for writing this simple syntax while still compiling to raw magma types (and therefore the generated code still adheres to the typing rules of the raw magma types).
+
+It should also be noted that due to the implementation details, expressions on smart types are not actually operable magma types. In fact, the only thing you can do with them is wire them to other smart types, as in `a @= expr`. At the time of wiring, bit-widths and signedness of operands are resolved as a function of the expression itself and the left hand side of the assignment (see below for details). So, a statement like `b @= expr`, means that the generated RTL could different based on the bit-width of `b`.
+
+## Types
+There are two smart type constructors:
+* `SmartBits[bit_width, signed=False]`: returns a SmartBits type with bit-width `bit_width` and signedness `sign`. `sign` defaults to `False`.
```
cdonovick, line 22: `SmartBits[bit_width, signed=False]`

(x) Doubt

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+### Summary of bit-width and signedness rules
+| Expression             | Operands           | Output bit-width | Output signedness       |
+|------------------------|--------------------|------------------|-------------------------|
+| Binary arith.,logical  | Context-determined | max(operands)  | all(signed(operands)) |
+| Unary logical          | Context-determined | max(operands)  | all(signed(operands)) |
+| Shift                  | Shiftee context-determined, shifter self-determined, | Shiftee bit-width | all(signed(operands)) |
+| Comparison             | Context-determined | 1                | unsigned                |
+| Reduction              | Self-determined    | 1                | all(signed(operands)) |
+| Concat                 | Self-determined    | sum(operands)  | unsigned |
+
```
cdonovick, line 62: What is the signedness of `int` constants?

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,526 @@
...
+    def resolve(self, context):
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def eval(self):
+        raise NotImplementedError()
+
+    # Binary arithmetic operators.
+    def __add__(self, other: '_SmartExpr'):
+        if not isinstance(other, _SmartExpr):
```
cdonovick, line 61: Using `not isinstance(other, type(self))` make this more robust with inheritance

magma/smart/smart_bits.py
cdonovick: Agreed.  `NotImplemented` is used for operators to signal that the lhs doesn't know how to handle the rhs and is converted to a type error if the rhs also does not how handle the left.   It should not be returned to a user.

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,526 @@
...
+        self._width_ = len(self._bits)
+        self._signed_ = type(self._bits)._signed
+
+    def eval(self):
+        return self._bits.typed_value()
+
+
+class _SmartBitsMeta(_SmartExprMeta):
+    def __getitem__(cls, key):
+        assert cls is SmartBits
```
cdonovick, line 429: Don't want people extending SmartBits?

magma/smart/smart_bits.py
```diff
@@ -0,0 +1,526 @@
...
+        assert isinstance(value, Bits)
+        signed = isinstance(value, SInt)
+        return SmartBits[len(value), signed](value)
+
+    def __str__(self):
+        signed = type(self)._signed
+        return f"SmartBits[{len(self)}, {signed}]({str(self._value)})"
+
+
+class SmartBit(SmartBits[1]):
```
cdonovick, line 518: Nit: this contradicts the docs which say:

> `SmartBit` which is a literal alias of `SmartBits[1]`

should be 
```Python
SmartBit = SmartBits[1]
``` 
Unless you are some sort of liar

docs/smart_bits.md
```diff
@@ -0,0 +1,71 @@
...
+### Bit-width
+The rule of thumb for context-determined expressions is that all it's operands get extended to the **longest** operand appearing in the complete expression (including the LHS of the assignment). On the other hand, the scope for self-determined expressions is only the partial expression itself. Note that since a context-determined expression can appear within a self-determined expression, each self-determined expression must still be procssed by the inference step, except that the context is limited to just the expression itself, and in particular, the LHS of the assignment is excluded.
+
+The bit-width of any partial expression's result is determined by the specific operator (see table below). For example the result of addition has bit-width equal to the maximum of all of it's operand bit-widths, where as the result of comparison always has bit-width equal to 1.
+
+### Signedness
+The signedness of operands can both determine the semantics of operators, as well as how operands are extended. For example, signed-comparison has different semantics than unsigned-comparison, and operands can either be sign-extended or zero-extended (during bit-width inference).
```
cdonovick, line 45: Given:
```Python
x = SmartBits[8, True]()
y = SmartBits[16, False]()
```
`x+y` is unsigned but is x sign extend or zero extended?

magma/smart/smart_bits.py
cdonovick: If you are just blocking `SmartBits[8][16]`  raise an appropriate error.

docs/smart_bits.md
rsetaluri: Good point. Wanted to make this code path decoupled from rest of existing magma. Once this is more mainline, there should be more integrated first class support, such as in `m.reduce()`.

docs/smart_bits.md
rsetaluri: Yes. Added clarification about this point earlier in the doc.

docs/smart_bits.md
rsetaluri: Yes, something like that would be super useful. I think in general (a) how we interoperate with non-smart code will be important for adoption and backwards compatibility, and (b) avoiding the use of "smart" literally in the code everywhere will be useful. Added a note about this to the docs TODO.

magma/smart/smart_bits.py
rsetaluri: This is actually that function, there is `from magma.primitives.reduce import reduce` at the top of the file (maybe a `as magma_reduce` would be clearer).

magma/smart/smart_bits.py
rsetaluri: ah ok missed that subtlety. thanks. fixed.

docs/smart_bits.md
rsetaluri: Unfortunately, `int` constants aren't supported operands of smart operators (presumably would be a blocker for actual usage...). But when it is added (soon), they will likely be unsigned. From the Verilog spec:

> * Expression type depends only on the operands. It does not depend on the left-hand side (if any).
> * Decimal numbers are signed.
> * Based_numbers  are  unsigned,  except  where  the  s  notation  is  used  in  the  base  specifier  (as  in"4'sd12").

magma/smart/smart_bits.py
rsetaluri: the problem with that is we can call the operator on two separate sub-classes, so

```python
class Sub1(Super): ...
class Sub2(Super): ...

Sub1() + Sub2()  # not isinstance(Sub2(), type(Sub1())), but we want this to work
```

docs/smart_bits.md
cdonovick: Gotcha.  I would advocate making them signed.  That way there treated as unsigned in an unsigned context and signed in a signed one.  Probably my most commonly used int constant, other than 0, is -1 (for building masks) and I would definitely want it be sign extended.

magma/smart/smart_bits.py
cdonovick: Do we want that to work?  What is the return type?

cdonovick: Further what is the type of `Sub1() + Sub1()`

cdonovick: If the issue is the mixing of unsigned and signed I would advocate checking for that case explicitly like have a way to get the non-signed (i.e. the type sans signedness information) from a SmartBits type and checking if other is an instance of that.

docs/smart_bits.md
rsetaluri: Sorry for the confusion, clarified the interface.

magma/smart/smart_bits.py
rsetaluri: 😂 good catch. I still need to fix this.

Actually, as you can see even `SmartBit = SmartBits[1]` excludes the possibility of a signed `SmartBit` type (which of course you could do just as `SmartBits[1, True]`.

What about a class `SmartBit` which subclasses and overrides `getitem` to just take in an optional bool for sign? The more I think about it, the more this seems kinda clunky. I'm not sure what you could do besides make a full separate metaclass for `SmartBit`

docs/smart_bits.md
rsetaluri: Great question. From my reading of the standard it should be **zero-extended**:

> The following are the steps for evaluating an expression:
> * Determine the expression size based upon the standard rules of expression size determination.
> * Determine the sign of the expression using the rules outlined in 5.5.1.
> * Propagate the type and size of the expression (or self-determined sub expression) back down to the context-determined  operands  of  the  expression.  In  general,  any  context-determined  operand  of  an operator  shall  be  the  same  type  and  size  as  the  result  of  the  operator.  However,  there  are  two exceptions:
>   ...
> * When propagation reaches a simple operand as defined in 5.2 (a primary as defined in A.8.4), then that  operand  shall  be  converted  to  the  propagated  type  and  size.  If  the  operand  must  be  extended,then it shall be sign-extended only if the propagated type is signed.

I actually just checked and the code was **not** following this^. So I've now fixed it in a recent commit.

docs/smart_bits.md
rsetaluri: That's a good point. Since we do not have the luxury of having base specifiers for integral constants e.g. `4'sd12` vs `4'd12`, it makes sense to make them signed by default. I will do that when the feature is added.

magma/smart/smart_bits.py
rsetaluri: The issue is not of signed-ness; rather its just a matter of the implementation design. That is `_SmartExpr` is an abstract base class...each "ast node" inherits from that (just colloquially calling them AST nodes, in reality they're just these staged values, but each type of op has it's own node, e.g. AddNode, ShiftNode, etc.). So we define these operators on the base class and get them for free. It's kind of like:

```python
class Node:
    def __add__(self, other):
        assert isinstance(other, Node)
        return AddNode(self, other)

   def __sub__(self, other):
        assert isinstance(other, Node)
        return SubNode(self, other)

class AddNode(Node): ...
class SubNode(Node): ...

class Leaf(Node): pass

x = Leaf()
y = Leaf()
z = (x + y) - (x + x - y) + ...  # whatever you want to do on x and y
```

All the intermediates, such as `(x + y)`, are indirectly instances of `Node` base cls, but we they could be any specialized node type.

magma/smart/smart_bits.py
rsetaluri: Agreed. In general the `eval` and `resolve` methods could be cleaned up and unified I think. However, given that it is a little bit fragile, I want to make sure the logic is right, and then I will think about a unification of these.

magma/smart/smart_bits.py
rsetaluri: Resolving for now; if you think there is better design, would be good to discuss. I'm not tied to this way, but it seems to be working.

magma/smart/smart_bits.py
rsetaluri: For now, going to go ahead with `SmartBit = SmartBits[1]` literally. Will think about this more. One option is to just get rid of `SmartBit` all together, though I suspect users will want it.

magma/smart/smart_bits.py
cdonovick: Gotcha, it might be possible to have a subtype of SmartBits automatically generate its own ast type so that it could be safely subtyped.  Let me think about it.

docs/smart_bits.md
cdonovick: Hmmm I suspect I like the old code better.  It feels more natural to me to have signedness of the sub expression determine the extension.  But I suspect other people would disagree (which is why I don't like these sort of coercion rules).   Anyway, is there some way to force extension so I could get this to work:
```Python
x = SmartBits[16, True](-1)
y = SmartBits[32, False](1)
assert x + y == 0
```

rsetaluri: Yes, that's the reason for the `signed()` and `unsigned()` coercion functions, so if you had

```python
x = SmartBits[16, True](-1)
y = SmartBits[32, False](1)
assert x + signed(y) == 0
```
it would work.

cdonovick: But that's not quite the same.  As that would return a signed expression and do a signed version of the operator (not that it matters for plus).

What I was wondering if there was some way to do: `x.ext(y.size-x.size) + y`

PR #772, opened 2020-07-16
Fix usernamespace decl bug

leonardt:
Fixes issue with user_namespace and declarations, also updates the golds for upstream coreir bug fix so that usernames-ace prefix is actually used in the generated verilog

### DIFF
--- a/tests/test_compile/gold/test_user_namespace_verilog_prefix.v
+++ b/tests/test_compile/gold/test_user_namespace_verilog_prefix.v
@@ -1,11 +1,20 @@
+// Module `_Decl` defined externally
 module my_namespace_Bar (
     input I_x,
     input I_y,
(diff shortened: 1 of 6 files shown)


tests/test_compile/gold/test_user_namespace_verilog_prefix.v
```diff
@@ -16,7 +25,7 @@ module my_namespace_Foo (
 );
 wire Bar_inst0_O_x;
 wire Bar_inst0_O_y;
-Bar Bar_inst0 (
+my_namespace_Bar Bar_inst0 (
```
leonardt, line 28: Turns out we weren't properly prefixing the instance module names, fixed by https://github.com/rdaly525/coreir/pull/920

Issue #759, opened 2020-07-01
[RFC] Inline Combinational Syntax


leonardt:
# Overview
One barrier to adoption of the combinational syntax is the overhead of defining the logic outside of a circuit definition.  For example, inside a magma circuit we may declare a list of sub-modules with various inputs and outputs.  To use combinational to manage these inputs/outputs, we would need to declare input arguments for each signal, assign variables to the returned output arguments, and wire the output arguments up to the right place.  One way we can make this easier to provide a way for the combinational syntax to refer to an outer scope.  This would allow us to avoid having to pass inputs and return outputs.

# Example
```python
foo_inst, bar_inst = Foo(), Bar()  # circuit instances
foo_inst.I @= bar_inst.O

@m.combinational2
def logic():
    if io.cond:
        bar_inst.I @= foo_inst.O + 1
    else:
        bar_inst.I @= foo_inst.O - 1
```
In the above example, the combinational code refers directly to the `foo_inst` and `bar_inst` ports in the enclosing scope, so the input/output ports don't need to be explicitly passed as arguments and returned.

# Issues
The existing combinational syntax logic is pretty tied to being a standalone function and compiling to a circuit.  I think it might be best to preserve this pattern since it's still useful for defining standalone, reusable combinations logic.  However, we could introduce a method of defining an "inline" combinational syntax block, such as using the with statement.  Here's one option:
```python
foo_inst, bar_inst = Foo(), Bar()  # circuit instances
foo_inst.I @= bar_inst.O

with m.combinational2() as syntax:
    if io.cond:
        bar_inst.I @= foo_inst.O + 1
    else:
        bar_inst.I @= foo_inst.O - 1
```
We could run an ast processing pass inside the circuit definition logic to transform this, but I'm not exactly sure if there's any issues with the current circuit definition pipeline that doesn't use the staged definition method (do we need to use a class decorator or can we find a hook in the meta class stages to do this?)

If we use a function style decorator (no with statement), then perhaps we could manage this logic inside the decorator, but the code might look a little less natural, for example
```python
foo_inst, bar_inst = Foo(), Bar()  # circuit instances
foo_inst.I @= bar_inst.O

@m.inline_combinational2()  # would rewrite the code then execute it
def logic():
    if io.cond:
        bar_inst.I @= foo_inst.O + 1
    else:
        bar_inst.I @= foo_inst.O - 1
```

[...]

PR #773, opened 2020-07-17
Prototype inline combinational

leonardt:
See https://github.com/phanrahan/magma/issues/759 for some background.

The goal is to provide the ability to write combinational syntax while referring to the enclosing scope.  This avoids having to pass and return values (tedious when there's a large number inside the module definition).

The minimum working example for prototyping:
```python
    class Main(m.Circuit):
        io = m.IO(invert=m.In(m.Bit), O=m.Out(m.Bit))
        io += m.ClockIO()
        reg = m.Register(m.Bit)()

        @m.inline_combinational(debug=True, file_name="inline_comb.py")
        def logic():
            if io.invert:
                reg.I = ~reg.O
            else:
                reg.I = reg.O

        io.O @= reg.O
```

The basic algorithm:
1. rewrite the tree so that all wiring is done as assignments to temporaries, return the temporaries at the end.  AugAssignment targets are mapped to a temporary value based on their immutable hash (structure).  This allows us to run generic SSA (since it's defined for variable assignments, rather than wiring to general objects).
1. run standard combinational passes on the tree
1. exec the function, call it to get the return values (the execution should use the enclosing scope so references will work), then wire the return values (temporaries) to the actual values stored in the mapping (original AugAssign AST node target hash).

For step 1, the above example is rewritten into:
```python
def logic():
    _auto_prefix_000 = ~reg.O
    _auto_prefix_001 = reg.O
    _auto_prefix_002 = __phi(io.invert, _auto_prefix_000, _auto_prefix_001)
    __return_value0 = _auto_prefix_002
    return __return_value0
```
(the first digit is from gen_free_prefix, the second digit is the unique id for each new temporary added with prefix (only 1 in this case), the third digit is from ssa)

I chose this path rather than rewriting to a standard comb2 function because there were some non-trivial analysis required.  For example, we would need to analyze the source to determine the input types (e.g. reg.O) and the output types (e.g. the final value of reg.I).  By simply doing the rewrites and executing them in the enclosing scope (without the comb2 wrapper), we avoid having to do this analysis.

Open to suggestions on how we could improve the algorithm as well as edge cases that we should watch out for.

### DIFF
--- a/magma/syntax/inline_combinational.py
+++ b/magma/syntax/inline_combinational.py
@@ -3,6 +3,7 @@ from typing import MutableMapping, Optional
 
 import astor
 
+import ast_tools
 from ast_tools.stack import SymbolTable
 from ast_tools.passes import (apply_ast_passes, Pass, PASS_ARGS_T, ssa,
                               if_to_phi, bool_to_bit)
@@ -27,7 +28,7 @@ class _ToCombinationalRewriter(ast.NodeTransformer):
 
     def visit_AugAssign(self, node):
         if not isinstance(node.op, ast.MatMult):
-            return
+            return node
         key = immutable(node.target)
         if key not in self.target_map:
             retval_name = self._gen_new_name()
@@ -72,8 +73,11 @@ class inline_combinational(apply_ast_passes):
                  ):
         self.target_map = {}
         passes = (pre_passes + [_RewriteToCombinational(self.target_map),
+                                ast_tools.passes.debug(dump_src=True),
                                 ssa(strict=False), bool_to_bit(),
-                                if_to_phi(lambda s, t, f: s.ite(t, f))] +
+                                if_to_phi(lambda s, t, f: s.ite(t, f)),
+                                ast_tools.passes.debug(dump_src=True),
+                                ] +
                   post_passes)
         super().__init__(passes=passes, env=env, debug=debug, path=path,
                          file_name=file_name)
--- a/magma/syntax/inline_combinational.py
+++ b/magma/syntax/inline_combinational.py
@@ -88,8 +88,8 @@ class inline_combinational(apply_ast_passes):
         if not isinstance(result, tuple):
             result = (result,)
         for key, value in zip(self.target_map.keys(), result):
-            # Eval origin target in env to do wiring, not sure if there's a way
-            # we can avoid having to do eval here (other option is to codegen
-            # the original augassign in a new tree, but that's effectively the
-            # same as evaling)
+            # Eval original target in env to do wiring, not sure if there's a
+            # way we can avoid having to do eval here (other option is to
+            # codegen the original augassign in a new tree, but that's
+            # effectively the same as evaling)
             wire(eval(astor.to_source(mutable(key)).rstrip(), {}, env), value)
--- a/magma/syntax/inline_combinational.py
+++ b/magma/syntax/inline_combinational.py
@@ -82,8 +82,8 @@ class inline_combinational(apply_ast_passes):
         super().__init__(passes=passes, env=env, debug=debug, path=path,
                          file_name=file_name)
 
-    def exec(self, etree, stree, env, metadata):
-        fn = super().exec(etree, stree, env, metadata)
+    def exec(self, *args, **kwargs):
+        fn = super().exec(*args, **kwargs)
         result = fn()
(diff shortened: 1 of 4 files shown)


cdonovick:
Could you write out how your example would be transformed by steps 2-4?

magma/syntax/inline_combinational.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, pre_passes=[], post_passes=[],
+                 debug: bool = False,
+                 env: Optional[SymbolTable] = None,
+                 path: Optional[str] = None,
+                 file_name: Optional[str] = None
+                 ):
+        self.target_map = {}
+        passes = (pre_passes + [_RewriteToCombinational(self.target_map),
+                                ssa(strict=False), bool_to_bit(),
+                                if_to_phi(lambda s, t, f: s.ite(t, f))] +
```
cdonovick, line 76: `if_to_phi(Bit.ite)` should probably work but its not actually equivalent.

leonardt:
After step 1 (rewrite wires)
```python
BEGIN SRC
@m.inline_combinational(debug=True, file_name='inline_comb.py')
def logic():
    if io.invert:
        _auto_prefix_00 = ~reg.O
    else:
        _auto_prefix_00 = reg.O
    return _auto_prefix_00
END SRC
```

step 2 (standard comb):
```python
BEGIN SRC
@m.inline_combinational(debug=True, file_name='inline_comb.py')
def logic():
    _auto_prefix_000 = ~reg.O
    _auto_prefix_001 = reg.O
    _auto_prefix_002 = __phi(io.invert, _auto_prefix_000, _auto_prefix_001)
    __return_value0 = _auto_prefix_002
    return __return_value0
END SRC
```

I changed step 3 to actually happen during step 1, will update the original post (so you see the return introduced above)

Step 4 doesn't have an "after" since it's executing code rather than rewriting the tree, but it's done by exec here: https://github.com/phanrahan/magma/pull/773/files#diff-3c29ac4b4e52ecb1cf3f6d008f8a0178R81-R91
Basically, the rewriter in step 1 stores a mapping from the original wire target to the temporary value, here the returned temporary value from executing the rewritten function above is wired up to the eval of the original wire target in the env.

magma/syntax/inline_combinational.py
```diff
@@ -0,0 +1,95 @@
...
+        self.name_count = 0
+
+    def _gen_new_name(self):
+        new_name = f"{self.prefix}{self.name_count}"
+        self.name_count += 1
+        return new_name
+
+    def visit_AugAssign(self, node):
+        if not isinstance(node.op, ast.MatMult):
+            return
```
cdonovick, line 30: This will delete aug assigns that are not matmul.  Probably want to raise an error because ssa doesn't transform them or minimally return node.

magma/syntax/inline_combinational.py
```diff
@@ -0,0 +1,95 @@
...
+        fn = super().exec(etree, stree, env, metadata)
+        result = fn()
+        if not isinstance(result, tuple):
+            result = (result,)
+        for key, value in zip(self.target_map.keys(), result):
+            # Eval origin target in env to do wiring, not sure if there's a way
+            # we can avoid having to do eval here (other option is to codegen
+            # the original augassign in a new tree, but that's effectively the
+            # same as evaling)
+            wire(eval(astor.to_source(mutable(key)).rstrip(), {}, env), value)
```
cdonovick, line 91: Could you try to look the name up in the env?  If its a name look up the name in env, if its and attribute getattr on the recursive call.

magma/syntax/inline_combinational.py
leonardt: Woops, yea should just return for now I think

magma/syntax/inline_combinational.py
leonardt: That would work, but in theory there could also be a function call? At that point, we're effectively parsing and evaluating the code, so perhaps just using eval of the source is fine?

cdonovick: Yeah in the face of a function we do need to do an eval, but I like trying to avoid evals where possible.

cdonovick: So I suppose eval is fine.  The algorithm I am defining is basically an eval anyway.

PR #776, opened 2020-07-20
Add support for normal assign to inline_combinational

leonardt:
Extends inline_combinational to support normal assignment statements.  These are handled similarly to wiring where the final value (determined by SSA) is returned from function and an assignment statement is "execed" (ugh, but not sure if there's a way around this without rewriting the AST from the enclosing scope, which probably isn't much better) with the return value from the rewritten inline_combinational function call

### DIFF
--- a/magma/syntax/inline_combinational.py
+++ b/magma/syntax/inline_combinational.py
@@ -4,10 +4,10 @@ from typing import MutableMapping, Optional
 import astor
 
 import ast_tools
-from ast_tools.stack import SymbolTable
+from ast_tools.stack import SymbolTable, _SKIP_FRAME_DEBUG_STMT, get_symbol_table
 from ast_tools.passes import (apply_ast_passes, Pass, PASS_ARGS_T, ssa,
                               if_to_phi, bool_to_bit)
-from ast_tools.common import gen_free_prefix
+from ast_tools.common import gen_free_prefix, gen_free_name
 from ast_tools.immutable_ast import immutable, mutable
 
 from magma.syntax.combinational2 import combinational2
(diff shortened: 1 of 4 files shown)


leonardt:
Also added some documentation

cdonovick:
A rewrite from the enclosing scope wouldn't be unreasonable if theres a way to do it cleanly.

magma/syntax/inline_combinational.py
```diff
@@ -84,9 +102,15 @@ def exec(self, etree, stree, env):
...
             # Eval original target in env to do wiring, not sure if there's a
             # way we can avoid having to do eval here (other option is to
             # codegen the original augassign in a new tree, but that's
             # effectively the same as evaling)
             wire(eval(astor.to_source(mutable(key)).rstrip(), {}, env), value)
+        for key, value in zip(self.assign_map.keys(),
+                              result[len(self.wire_map):]):
+            value_name = gen_free_name(ast.NameConstant(False), env)
+            env.locals.update({value_name: value})
+            exec(f"{astor.to_source(mutable(key)).rstrip()} = {value_name}",
```
cdonovick, line 115: The more evaling / execing there is the more I am afraid of our attempts at recreating the env will break.  Maybe this could move to a pattern similar to peak where users must pass locals / globals. E.g. https://github.com/cdonovick/peak/blob/master/examples/riscv/sim.py#L23

Having to write `m.comb(locals(), globals())` would be somewhat annoying but it rules out a very hard to debug class of bugs.

leonardt: Yea I think that's a reasonable idea, let me see if there's any safer way to do this.  I looked into rewriting from the enclosing scope but it's not so easy.  The main problem is that the raw source code is actually invalid (for the assign case, since you'll get a name error when referring to a variable assigned inside the inline_comb. block), so we can't use a decorator pattern on the class even if we wanted to (not ideal anyways since we'd have to decorate instead of subclass circuit).  Other options face similar issues because of the potential for a name error from the invalid code.

I'm going to think about whether there's an alternative syntax that avoids the name error, perhaps using a `with m.inline_combinational as syntax:` block could work.

leonardt: Basically if we can find a way to let the code evaluate without an exception, we could then just hijack part of the class pipeline to introspect the source, do the rewrite, then return the evaluation of the rewrite (but this has other potential problems, as we know with getting source from a class).

PR #777, opened 2020-07-22
Add prototype for NDArray

leonardt:
This minimum working example subclasses the Array/ArrayMeta classes to
minimize the amount of code to write.  One problem with this approach is
that the Array/ArrayMeta logic depend on the `Array[N, T]` interface for
dealing with various magma operations such as flipping types.  The
NDArray type does not satisfy this interface for it's type constructor,
so for now it just assumes invalid parameters are to be passed through
to the Array super class logic.  The downside to this approach is that
in theory, a user could do `NDArray[N, T]` and get back an `Array[N, T]`
rather than error message (for now NDArray parameters should be a tuple
of values convertible to int).

Two options I can immediately see for cleaning up this pattern:
1. Move the generic Array/ArrayMeta logic into a base class, inherit
from Array and NDArray.  I'm not exactly sure how this would work, but
this seems like the natural choice for reusing the core logic code with
different interfaces.  I have a feeling this will be an non-trivial
undertaking just based on the complexity of the Array/ArrayMeta logic.
2. Use the MagmaProtocol pattern for NDArray so the underlying value
uses the Array logic and the NDArray interface just dispatches
conveneniences like tuple indexing to the underlying value.  This seems
like a simpler choice that doesn't involve messing with the
Array/ArrayMeta pipeline


leonardt:
Based on feedback from https://github.com/phanrahan/magma/issues/752, I've updated this to use `Array[(M, N), T]` as the NDArray pattern, which solves the code organization issue reported in the original PR writeup.

magma/array.py
```diff
@@ -275,6 +288,31 @@ def flat_length(cls):
         return cls.N * cls.T.flat_length()
 
     def __getitem__(self, key):
+        if isinstance(key, tuple):
+            if len(key) == 1:
+                return self[key[0]]
+            # ND Array key
+            if not (key[-1] == slice(None) or key[-1] == slice(0, None) or
```
rsetaluri, line 295: what is this condition? can you augment the comment to explain this a bit?

magma/array.py
```diff
@@ -89,6 +89,19 @@ def __getitem__(cls, index: tuple) -> 'ArrayMeta':
 
         # index[0] (N) can be None (used internally for In(Array))
         if index[0] is not None:
+            if isinstance(index[0], tuple):
```
rsetaluri, line 92: technically, I don't think anything would break...but should we constrain what `T` can be? Specifically if `T` is itself some array or nested type, its unclear what the semantics should be. if `T` is array, should that be sliceable in the same way too? basically are `Array[N, Array[M, T]]` and `Array[(N, M), T]` the same?

magma/array.py
leonardt: Will leave a comment, it checks if the key is any of `x[:], x[0:], x[:len(x)], x[0:len(x)]`, basically checks if it's trying to slice the entire array

magma/array.py
leonardt: Right now those are treated as the same, so the ndarray logic just assumes a container of type `Array[M, Array[N, ...]]`.  The meta class logic unpacks `Array[(M, N)]` into the above form and the tuple indexing logic assumes it has the corresponding structure.  So the above example will work, and while confusing, not sure if we need to constrain it

rsetaluri: I think that's ideal actually. It would be even more confusing if they resulted in different Thanks for clarifying.

leonardt:
Now depends on https://github.com/rdaly525/coreir/pull/928

PR #780, opened 2020-07-24
Fix #778

### LINKED ISSUE #778: sequential2 ite issue with namedtuple + Register combination
a bit tricky

    @m.sequential2()
    class Test:
        def __init__(self):
            self.v = m.Register(T=m.Bit, init=m.bit(0))()

        def __call__(self, sel: m.Bit) -> m.AnonProduct[dict(a=m.Bit)]:
            if sel:
                return m.namedtuple(a=self.v)
            else:
                return m.namedtuple(a=self.v)

Both branches are the same type but `ite` can't handle it

    E           TypeError: ite expects same type for both branches: Tuple(a=<class 'magma.syntax.sequential2._SequentialRegisterWrapperOut(Bit)'>) != Tuple(a=<class 'magma.syntax.sequential2._SequentialRegisterWrapperOut(Bit)'>)

splhack:
Convert MagmaProtocol type/value in `_tuple` to fix #778.

(Another option would be keeping MagmaProtocol inside Tuple/Product and dealing it with later. Most likely it requires a lot of code changes, but no particular benefits?)

### DIFF
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -14,6 +14,7 @@ from .bfloat import BFloat
 from .digital import Digital
 from .tuple import Tuple, Product
 from .bitutils import int2seq
+from .protocol_type import get_type, get_value
 import hwtypes
 
 __all__ = ['bit']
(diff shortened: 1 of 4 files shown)


leonardt:
Hmm, let me check out how hard it would be to keep the magmaprotocol inside the tuple/product, I think this would be the ideal thing to do (in the same way you could define a tuple/product type with a magma protocol field)

splhack:
keeping MagmaProtocol #784

PR #782, opened 2020-07-25
inline_combinational: support m.wire

### LINKED ISSUE #781: inline_combination without return value?
Is this code valid as an inline combinational code?
(just deleted `O1 = reg.O` lines from tests/test_syntax/test_inline_comb.py)

    @m.inline_combinational()
    def logic():
        if io.invert:
            reg.I @= ~reg.O
        else:
            reg.I @= reg.O

If so, how about this? using `m.wire` instead of `@=`.

    @m.inline_combinational()
    def logic():
        if io.invert:
            m.wire(~reg.O, reg.I)
        else:
            m.wire(reg.O, reg.I)

both are same code but the latter causes error, because the logic function doesn't return value.

    >       retval = ast.Tuple(elts, ast.Load()) if len(elts) > 1 else elts[0]
    E       IndexError: list index out of range

splhack:
Fix #781

### DIFF
--- a/magma/syntax/inline_combinational.py
+++ b/magma/syntax/inline_combinational.py
@@ -10,16 +10,18 @@ from ast_tools.passes import (apply_ast_passes, Pass, PASS_ARGS_T, ssa,
 from ast_tools.common import gen_free_prefix, gen_free_name
 from ast_tools.immutable_ast import immutable, mutable
 
+import magma as magma_module
 from magma.syntax.combinational2 import combinational2
 from magma.bit import Bit
 from magma.wire import wire
 
 
 class _ToCombinationalRewriter(ast.NodeTransformer):
-    def __init__(self, free_prefix, wire_map, assign_map):
+    def __init__(self, free_prefix, wire_map, assign_map, env):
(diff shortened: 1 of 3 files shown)


magma/syntax/inline_combinational.py
```diff
@@ -44,6 +49,12 @@ def visit_Assign(self, node):
             node.targets[i] = ast.Name(self.assign_map[key], ast.Store())
         return node
 
+    def visit_Call(self, node):
+        if node.func.value.id in self.magma_ids and node.func.attr == "wire":
```
leonardt, line 53: Hmm, rather than using this "magma_ids" pattern, how about just passing the "env" to this visitor and calling "eval" on node.func inside the env? This should support more general patterns and we're passing around/evaling inside the env anyways.

leonardt: You could use "astor.to_source" to dump the AST to a string that you can eval with env.

splhack: sure, seems `eval(node.func.value.id, {}, self.env)` is working

PR #783, opened 2020-07-25
Support non-"m" magma module for Register updater

splhack:
This unblocks to use non-"m" magma module, like `magma.Register` instead of `m.Register`

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -1,4 +1,5 @@
 import ast
+import magma as magma_module
 from typing import Optional, MutableMapping
 
 from ast_tools.stack import SymbolTable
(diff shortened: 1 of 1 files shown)


magma/syntax/sequential2.py
```diff
@@ -216,12 +218,15 @@ def __init__(self, reset_type, has_enable, reset_priority):
...
+    def set_magma_ids(self, magma_ids):
+        self.magma_ids = magma_ids
+
     def visit_Call(self, node: ast.Call):
         # find m.Register
         if (
             isinstance(node.func, ast.Attribute)
             and isinstance(node.func.value, ast.Name)
-            and node.func.value.id == "m"
+            and node.func.value.id in self.magma_ids
```
leonardt, line 229: See this comment: https://github.com/phanrahan/magma/pull/782#discussion_r461002871

I think we can use a similar pattern here where we just "eval" the node.func in the environment (it's available below in the rewrite function).  It provides the same feature but should generalize to some more patterns (e.g. a function call that returns magma)

PR #784, opened 2020-07-28
support MagmaProtocol in Array/Tuple/Product

### LINKED ISSUE #778: sequential2 ite issue with namedtuple + Register combination

splhack:
Fix #778
Close #780 
Depends on #789

~AnonProduct works with this PR. Tuple doesn't, yet.~


splhack:
if sel:
        return m.tuple_(m.bit(0))
    else:
        return m.tuple_(self.v.prev())

    E           TypeError: ite expects same type for both branches: Tuple[Bit[Out]] != Tuple[_SequentialRegisterWrapperOut(Bit)]

need more hacks..

leonardt:
Hmm yea, let me think about how to handle that.  Basically in this case I think we should be coercing the register wrapper to the bit type to match the expected type, but let's see if we can come up with a reasonable scheme for this.

rsetaluri:
Not sure if this exactly the same thing, but would something like Caleb's PolyIte scheme work here?

splhack:
if (
        isinstance(t_type, TupleMeta) and isinstance(f_type, TupleMeta) and
        t_type.field_dict.keys() == f_type.field_dict.keys()
    ):
        for k, v in t_type.field_dict.items():
            if magma_type(v) is not magma_type(f_type[k]):
                raise type_error

works, I don't like this though since it doesn't work nested one.

splhack:
not finished yet, but `iteeq` would work for nested case and Array.

splhack:
ok, this is the minimum change for supporting MagmaProtocol in Array, Tuple, and Product. can be nested.

splhack:
Now it can handle complex nested `Array`/`Tuple`/`Product` `ite`,

    class Test:
        def __init__(self):
            self.a = m.Register(T=m.Bit, init=m.bit(0))()
            self.b = m.Register(T=m.Bits[2], init=m.bits(0, 2))()

        def __call__(self, sel: m.Bit) -> m.AnonProduct[dict(
                                              a=m.Tuple[m.Bits[2]],
                                              b=m.Array[2, m.Bit],
                                          )]:
            self.a = self.a
            self.b = self.b
            if sel:
                return m.namedtuple(
                           a=m.tuple_([self.b.prev()]),
                           b=m.array([self.a.prev(), self.a.prev()]),
                       )
            else:
                return m.namedtuple(
                           a=m.tuple_([m.array([self.a.prev(), self.a.prev()])]),
                           b=m.array(self.b.prev()),
                       )

also complex register definition.

    class T(m.Product):
        a = m.Array[1, m.Bits[2]]
        b = m.Tuple[m.AnonProduct[dict(c=m.Array[2, m.Bit])]]

    @m.sequential2()
    class Test:
        def __init__(self):
            self.a = m.Register(
                T=T,
                init=m.namedtuple(
                    a=m.array([m.bits(0, 2)]),
                    b=m.tuple_([m.namedtuple(c=m.array([m.bit(0), m.bit(0)]))]),
                ),
            )()

        def __call__(self, sel: m.Bit) -> T:
            self.a = self.a
            if sel:
                return m.namedtuple(
                    a=m.array([m.bits(0, 2)]),
                    b=m.tuple_([m.namedtuple(c=m.array([m.bit(0), m.bit(0)]))]),
                )
            else:
                return self.a

splhack:
looks like it depends on ast_tools master (haven't checked which part does)

splhack:
commented out a test that depends on ast_tools trunk.

leonardt:
I just pushed a new release for ast_tools so once that's ready we can retry the test

splhack:
eh, I think the cause was Python 3.7 vs 3.8 in ast_tools.
The ast_tools code supports 3.8 but pip package's immutable_ast.py is for 3.7.

> E       TypeError: __init__() got an unexpected keyword argument 'type_comment'

splhack:
https://github.com/leonardt/ast_tools/pull/59

rsetaluri:
@leonardt @splhack I don't have the full context for this change, but it seems pretty sizable and a change to the core. Would it be worth doing a small design review?

splhack:
@rsetaluri sure,

----
# Goal

To make sequential2 replaceable with sequential

# Context

In sequential2, Register object becomes a _SequentialRegisterWrapper object which is an instance of MagmaProtocol.

    @m.sequential2()
    class Test:
        def __init__(self):
            self.a = m.Register(T=m.Bit, init=m.bit(0))()

        def __call__(self) -> m.Bit:
            print(type(self.a))

This code shows `_SequentialRegisterWrapperOut(Bit)` not `Register`.

So MagmaProtocol can be being as a Bit or a Bits.

    self.a = m.Register(T=m.Bit, init=m.bit(0))()

    self.a = m.Register(T=m.Bits[8], init=m.bits(0, 8))()

Or more complex type like Array, Tuple, or Product. 

    class T(m.Product):
        a = m.Array[1, m.Bits[2]]
        b = m.Tuple[m.AnonProduct[dict(c=m.Array[2, m.Bit])]]

    self.a = m.Register(T=T, init=m.namedtuple(...))

    # print(type(self.a))
    # _SequentialRegisterWrapperTuple(a=Array[1, Out(Bits[2])],b=Tuple(Tuple(c=Array[2, Out(Bit)])))

# Changes

## `magma_type` and `magma_value`

As https://github.com/phanrahan/magma/issues/553 and https://github.com/phanrahan/magma/pull/566, MagmaProtocol can be casted to the underlying magma type via `_to_magma_()` method. But it requires type check first (`if isinstance(value, MagmaProtocol):`). `magma_value` and `magma_type` are kind of convenient cast functions for those conversion.

The code doesn't call those cast as much as possible, because we want to keep MagmaProtocol as much as possible.

## `is_wireable`

MagmaProtocol can be a part of Array, Tuple, and Product. And `__eq__` type check fails for the types since those are different types.

    Tuple[Bit]

    Tuple[_SequentialRegisterWrapperTuple(Bit)]

`is_wireable` method allows the capability to wire each other, on even nested complex class.


## `is_bindable`

Similar with `is_wireable`, but for `issubclass`, on even nested complex class.

## MagmaProtocol properties, methods

Those things allows MagmaProtocol can be being as a Bit/Bits/Array/Tuple/Product.

rsetaluri:
@splhack thanks for the detailed review! taking a look.

magma/array.py
```diff
@@ -228,7 +241,7 @@ def __init__(self, *args, **kwargs):
         else:
             for i in range(self.N):
                 T = self.T
-                t = T(name=ArrayRef(self, i))
+                t = magma_type(T)(name=ArrayRef(self, i))
```
rsetaluri, line 244: is this the desired behavior? this excludes the possibility of `Array[N, MyProtocol]()[0]` returning an instance of `MyProtocol`? (Admittedly, it probably wasn't possible before either.)

magma/array.py
```diff
@@ -325,7 +338,7 @@ def as_list(self):
     def wire(i, o, debug_info):
         # print('Array.wire(', o, ', ', i, ')')
 
-        if not isinstance(o, ArrayType):
+        if not isinstance(magma_value(o), ArrayType):
```
rsetaluri, line 341: should we just do `o = magma_value(o)` at the top? For example, below we do `o.N`...what if `o` is protocol type there?

magma/backend/coreir_utils.py
```diff
@@ -49,8 +49,7 @@ def magma_port_to_coreir_port(port):
 
 
 def check_magma_type(type_, error_msg=""):
-    if issubclass(type_, MagmaProtocol):
-        type_ = type_._to_magma_()
+    type_ = magma_type(type_)
```
rsetaluri, line 52: thanks for replacing these!

magma/conversions.py
```diff
@@ -257,7 +260,8 @@ def _tuple(value, n=None, t=Tuple):
     if isinstance(value, t):
         return value
 
-    if not isinstance(value, (Digital, Array, IntegerTypes, Sequence, Mapping)):
+    if not isinstance(magma_value(value),
+                      (Digital, Array, IntegerTypes, Sequence, Mapping)):
```
rsetaluri, line 264: we should probably also change the error message here, like "bit() can only be used on Bit, Array, int, or wrapped protocol type, not {type(value)}"

magma/digital.py
```diff
@@ -126,6 +127,12 @@ def qualify(cls, direction):
     def __eq__(cls, rhs):
         return cls is rhs
 
+    def is_wireable(cls, rhs):
```
rsetaluri, line 130: @leonardt should we doing these checks on the undirected form of `cls` and `magma_type(rhs)`?

magma/digital.py
```diff
@@ -154,7 +161,7 @@ def wire(self, o, debug_info):
         if isinstance(o, IntegerTypes):
             o = HIGH if o else LOW
 
-        if not isinstance(o, Digital):
+        if not isinstance(magma_value(o), Digital):
```
rsetaluri, line 164: same here as above, should we just do `o = magma_value(o)` so `o._wire` is valid?

magma/protocol_type.py
```diff
@@ -34,6 +34,16 @@ def flip(cls):
     def __len__(cls):
         return len(cls._to_magma_())
 
+    @property
```
rsetaluri, line 37: I feel like I am starting to agree with @cdonovick ... rather than just adding all these functions directly to this type (ala mixins) we should just do `value = magma_value(value)` everywhere, then we don't have to duplicate these functions on the protocol types (ala protocols).

magma/protocol_type.py
```diff
@@ -87,9 +125,20 @@ def wire(self, other):
     def __imatmul__(self, other):
         self.wire(other)
 
+    def __getitem__(self, idx):
```
rsetaluri, line 128: feel pretty strongly about not having these defined on protocols. see above comment. furthermore, it means we can't define these custom functions on the protocols.

magma/array.py
splhack: tried to get MagmaProtocol one and failed. let me try again.

magma/digital.py
splhack: The purpose of `is_wireable` is for `Bit` `ite`, `t_type is not f_type`.
https://github.com/phanrahan/magma/blob/687c1aa1174a52f78fdefaab26896b89e5029e7a/magma/bit.py#L149
The problem was, for example, when comparing `Tuple[Bit]` vs `Tuple[MyProtocolBit]`. So the check is supposed to work for nested Array/Tuple/Product.

magma/protocol_type.py
splhack: ok, let me try to use `magma_value` more to avoid these property/methods.

magma/digital.py
rsetaluri: Got it. In that case we should use the undirected version right? The concern is if `In(Bit)` is wireable to `Out(Bit)` or not.

magma/protocol_type.py
splhack: deleted

magma/protocol_type.py
splhack: deleted

magma/digital.py
splhack: deleted with `o = magma_value(o)`

magma/array.py
splhack: tried like this, 

    t = T._from_magma_value_(T._to_magma_()(name=ArrayRef(self, i)))

but no luck..

        def _get_magma_value_(self):
            # Return output
    >       return self.circuit()
    E       TypeError: __call__() missing 1 required positional argument: 'output'

magma/array.py
splhack: called magma_value before calling wire.

magma/conversions.py
splhack: for now, leave it as is in this PR. wrapped protocol type is not enough, but the underlying type of the wrapped protocol type should be Bit, Array, or int, in this case.

magma/digital.py
splhack: maybe `is_wireable` is not a good name. It's `__eq__` for `ite`, only (so, I named it `ite_eq` before.)

splhack:
hmm, I can't repro the travis-ci tests/test_syntax/test_to_verilog.py::test_simple_alu failure locally.

magma/array.py
rsetaluri: Can you provide the full stacktrace of this error? not sure exactly where this is coming from

magma/digital.py
rsetaluri: Is it only used for `ite` right now? I think we should have a free function implementing this logic locally where it is used (we can move it in the future if it gets used elsewhere). Just concerned about attaching this specific of logic to the types.

rsetaluri: scratch that, because it gets called in bit.py which would then have dependency on Array/Tuple types, which we don't want. Member functions are fine.

leonardt:
I just ran into the  tests/test_syntax/test_to_verilog.py::test_simple_alu, it looks like an upstream tool for verilog generation is producing non-deterministic verilog, which sometimes causes the test to fail when checking the verilog output

for now, you can either re-run the build and see if it passes, or at a `@pytest.skip` decorator on the test for now until we resolve the issue upstream.

leonardt:
Actually, I was wrong, it's not a non-determinism issue, it's if you upgrade the tool, it changed the ordering of the code generation, so you can try upgrading your local copy with `pip install --upgrade kratos` and you should be able to reproduce, we can just update the gold

magma/array.py
splhack: Ok, figured out what's going on.
- patch https://gist.github.com/splhack/e97dc61a886626f956fecfb9258b742a
- full failure log https://gist.github.com/splhack/27bdefdb19f95eca9d7b05da8805e867

I'm using sequential2 test, so all `MagmaProtocol` class is `_SequentialRegisterWrapper`. The class `__init__` expects always circuit (`Register`) as magma value. `_get_magma_value_` calls `__call__()` method of the "circuit".

But, in the failed tests, the magma value is non-circuit, like `Bit`. It doesn't have `__call__()`.

rsetaluri: It seems like `_SequentialRegisterWrapperMeta._to_magma_` needs to be updated then?

splhack: These code paths 

- array.py https://github.com/phanrahan/magma/blob/69c64ce286cb89a6cd10285d4a0c1cb71faf96b5/magma/array.py#L229-L232
- tuple.py https://github.com/phanrahan/magma/blob/2b16f67f687ff7a59860e95960325bb8d85c7476/magma/tuple.py#L134-L138

are only called from interface.py `_make_port` (in my understanding.)

- interface.py https://github.com/phanrahan/magma/blob/a7e30f265137cb1e5e87bff6580a6c2ffd265ca3/magma/interface.py#L63

but

> this excludes the possibility of Array\[N, MyProtocol\]()[0] returning an instance of MyProtocol?

No, it can be done by the other side of the code path. For example, from this PR tests,

    def __init__(self):
        self.v = m.Register(T=m.Bit, init=m.bit(0))()
    def __call__(self):
        a = m.array([self.v.prev(), self.v.prev()])
        print(type(a)) # -> Array[2, _SequentialRegisterWrapperOut(Bit)]
        print(isinstance(a[0], MagmaProtocol)) # -> True

splhack:
@rsetaluri @leonardt any other concerns on this?

leonardt:
@splhack, can you allow edits from contributors? There should be a little button on the top right of this page.  The golds need to be updated because coreir changed how some verilog code is produced, I can take care of it though and push to the branch.

splhack:
@leonardt yes, 'Allow edits by maintainers' is checked.

leonardt:
Oops, forgot I had to point the branch to push to your fork as the remote, working now

rsetaluri:
@leonardt heads up that some recent changes (#829) may affect this change:

https://github.com/phanrahan/magma/commit/849ab54e4bdb789b615e3c6e250764c4dcfdda99
https://github.com/phanrahan/magma/commit/d49637eb85501572c94c1bddacb2f82868c15760

has to do with initializing array/tuple with protocol type

leonardt:
Fixed, they were basically the same change which was nice.  There's an issue popping up with named tuple and product type checking, will look into a fix and then merge this since it's been outstanding for awhile and I think it's good progress towards this support.

splhack:
🎉

PR #785, opened 2020-07-28
Only call set top if main is a definition

leonardt:
This fixes a regression introduced by #775.  In order to support
user_namespace, we changed the logic to save the entire context instead
of just saving the top module.  Before, saving the top module worked
since it just passed a name (str) ref to the fileWriter logic which was
used to emit the top reference (see
https://github.com/rdaly525/coreir/blob/5b4e98355e8f849f2ede137e32f8280d8a28f7d1/src/passes/analysis/coreirjson.cpp#L367).
This unblocks downstream users that have a top that is a declaration
(coreir's Context.setTop will error if trying to call setTop with a
declaration).  Later on we could change setTop to accept a declaration
rather than a definition, but I'm not sure if that's important (because
marking a declaration as Top isn't really useful since there's no
dependency graph to trace from the top anyways).

### DIFF
--- a/magma/backend/coreir_compiler.py
+++ b/magma/backend/coreir_compiler.py
@@ -3,6 +3,7 @@ import subprocess
 from .coreir_ import InsertWrapCasts
 from ..compiler import Compiler
 from ..frontend import coreir_ as coreir_frontend
+from ..is_definition import isdefinition
 from ..passes import InstanceGraphPass
 from ..passes.insert_coreir_wires import InsertCoreIRWires
 from ..logging import root_logger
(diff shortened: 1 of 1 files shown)


leonardt:
This unblocks `canal` which compiles an empty top for a test

leonardt:
Another option is to make compiling a declaration an error? One issue is that it produces an empty verilog file (since declarations are skipped as they're assumed to be externally defined and there's no dependencies to compile).  We could force canal to use "unused" or "undriven" to make the empty module have a definition (if an invariant we want to enforce is that compiled circuits have definitions, at least for the top).  Unless we see a use case for having a coreir json file with just a declaration inside it? Then we may want to add support for a declaration "top", but I don't really see value in this (again, no dependencies to be traced from the top).

rsetaluri:
It seems like it could be desirable to compile a declaration to CoreIR, e.g. to capture a generator/module interface or something for a library header. I would agree that on the CoreIR end, it doesn't make sense to compile a declaration to Verilog. For now, I don't see a "fundamentally" right thing, whatever supports the current use cases is fine.

PR #787, opened 2020-07-31
Support integer arg in mux operator

leonardt:


magma/primitives/mux.py
```diff
@@ -79,7 +79,14 @@ def mux(I, S, **kwargs):
         S = seq2int(S.bits())
     if isinstance(S, int):
         return I[S]
-    T = type(I[0])
+    # get first magma arg for type introspection
```
rsetaluri, line 82: Good change. While we're here, a few questions:

* Should we eagerly check the types of `I` to check that they're somehow all compatible? The main reason is that something like `mux(Bit[3](), Bit[4](), ...)` would error a little downstream (probably wiring inside the `Mux()()()` call) and produce confusing error messages. Although on the flip side, you don't want to be too eager since e.g. protocol types might have custom wiring.
* How are integer types wired to complex types, e.g. Tuples? Imagine it would be the same error as above, like trying to wire `0` to `Tuple[Bit, Bit]`.

Perhaps the cleanest bet would be to just add some documentation like, "Multiplexes input list `I`, using integral constant, or select signal `S`. The type is inferred using the first magma/magma protocol type in `I`, and all other elements of `I` are required to be wireable to this element."

magma/primitives/mux.py
```diff
@@ -79,7 +79,14 @@ def mux(I, S, **kwargs):
...
+    for arg in I:
+        if isinstance(arg, (Type, MagmaProtocol)):
+            T = type(arg)
+            break
+    else:
+        raise TypeError("Cannot use m.mux with non-magma types (need at least "
```
rsetaluri, line 88: nit: width -> type

magma/primitives/mux.py
leonardt: I think the ability to wire an integer relies on the type defining coercion logic, which we usually define by using a constructor (e.g. coercing a `1` to a `Bits[5]` should call `Bits[5](1)`, for a general T, this may or may not work, which would define whether it supports coercing integers).

I think we can possibly combine the two options: state that the type is inferred to be the first magma type, and the non-magma types are coerced using the above pattern (so this should be raise a type error at this point).  In the future, we may investigate a more sophisticated inference scheme (e.g. the first magma type might not be able to coerce a later argument, but the type of the later argument might be able to coerce the earlier type, we could use some greedy algorithm that checks the first, then tries the next, etc...).  In general, the user can use an explicit Mux(T) if the inference algorithm doesn't work for their use case.

leonardt: Okay, it turns out `T(arg)` won't work for some of our cases.  I'll have to dig deeper into why since it works if you wire arg to T()), it may simply be an inconsistent interface issue.  Ideally if you can wire `x` to `T()` then `T(x)` should work as well but that may not be the case for some of our types.  For now, I'll note in the docs "coercion is defined by wiring arg to T(), and if that fails, you'll get a wiring error"

rsetaluri: Yeah the constructors of the various types are ad-hoc, and could easily be buggy. That makes sense.

PR #792, opened 2020-08-07
Define arith ops on Bits for consistency with hwtypes

### LINKED ISSUE #587: Arithmetic operators on Bits
Currently every implementation of `AbstractBitVector` other than `Bits`  supports the use of arithmetic `__operators__` (`__add__`, `__sub__`, etc...).   While not specifically defined in the abc this had lead to a widespread assumption that `AbstractBitVector` supports these operators, further, it leads to PEak code that works in every context except for magma.  

For example the following mantle test assume that `BitVector` defines `__add__`
https://github.com/phanrahan/mantle/blob/master/tests/test_coreir/test_arith.py#L10

Now, I believe that one following must happen: 
  - `Bits` needs to add support for arithmetic ops, 
  -  we need to remove them from `BitVector` so that two are consistent
  -  we need to create a `MagmaBV` distinct from `Bits` behaves like the rest of hwtypes.   

Now removing support for arithmetic in `BitVector` would mean breaking almost every consumer of hwtypes.  Adding them to `Bits`  or creating a `MagmaBV` would break nothing.

leonardt:
Fixes #587

Should we also define the comparison ops?


cdonovick:
Yes all `bv*` should be defined.  I have come around to having ambiguous `__operators__` undefined on `Bits` for behaviour subtyping reasons (i.e. anything with a signed and unsigned version (compares, div, rem, shift, ext)).   However the explicit `bv*` (along with sext/zext) methods are unambiguous and hence should be defined regardless of the type.

PR #796, opened 2020-08-12
Update context reset to free memory

leonardt:
Depends on https://github.com/leonardt/pycoreir/pull/143

### DIFF
--- a/magma/backend/coreir_.py
+++ b/magma/backend/coreir_.py
@@ -24,6 +24,9 @@ _logger = root_logger().getChild("coreir_backend")
 _logger.setLevel(config.coreir_backend_log_level)
 
 
+_CONTEXT_TO_MODULES = {}
+
+
 # Singleton context meant to be used with coreir/magma code
 @singleton
 class CoreIRContextSingleton:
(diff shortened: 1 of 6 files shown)


magma/backend/coreir_.py
```diff
@@ -24,6 +24,9 @@
 _logger.setLevel(config.coreir_backend_log_level)
 
 
+_CONTEXT_TO_MODULES = {}
```
rsetaluri, line 27: nit: prefer lower case even for global variables, unless constant

magma/compile.py
```diff
@@ -45,6 +45,8 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
     opts = kwargs.copy()
     compiler = _make_compiler(output, main, basename, opts)
 
+    WireClockPass(main).run()
+
```
leonardt, line 49: From some reason I needed to add this for the tests to pass with the new context resetting logic? I scanned the coreir backend code and can't seem to find anywhere where we're doing this, nor is it being done in the coreir backend (don't see any reference to the wire clocks pass), CC @rsetaluri remember any changes related to this logic? Can't seem to remember if/why we removed it.

rsetaluri: honestly, looking back and can't find any commit where this was called in this file...that is weird. Did we ever call it?

rsetaluri: or did we call it in a different place? does CoreIR have an optional pass that gets triggered by dependency?

leonardt: We used to do it in the coreir backend, the wireclock and wiredefaultclock functions are imported but unused, maybe we removed it sometime?

leonardt: https://github.com/phanrahan/magma/blob/5d85cc39ce10a7ba68a0b421dc25623d77eab923/magma/backend/coreir_transformer.py#L9

rsetaluri: I see. I did a bisect (albeit half-baked) and couldn't find where we were actually calling it...I'll keep looking

PR #810, opened 2020-08-28
Add preliminary support for BitPattern

leonardt:
This adds support for declaring and comparing bit pattern values that
can include don't cares (?) ala chisel's BitPat:
https://github.com/freechipsproject/chisel3/blob/v3.3.2/src/main/scala/chisel3/util/BitPat.scala

For now, we implement the comparison logic in magma using a mask for the
don't care bits, in the future we should consider whether it matters
that we generate these in the output verilog (for synthesis quality or
readability).  If it does matter, we'll need to add support for
compiling it through CoreIR.  This uses the chisel syntax for bit
pattern strings.

### DIFF
--- a/magma/types/bit_pattern.py
+++ b/magma/types/bit_pattern.py
@@ -40,7 +40,7 @@ class BitPattern:
         self.bits = BitVector[count](bits)
         self.mask = BitVector[count](mask)
         self.width = count
-        self.const = self.mask != ((1 << self.width) - 1)
+        self.const = self.mask == ((1 << self.width) - 1)
 
     def __eq__(self, other):
         if not isinstance(other, Bits):
--- a/magma/types/bit_pattern.py
+++ b/magma/types/bit_pattern.py
@@ -38,6 +38,7 @@ class BitPattern:
                 count += 1
         self.bits = BitVector[count](bits)
         self.mask = BitVector[count](mask)
+        self.width = count
 
     def __eq__(self, other):
         if not isinstance(other, Bits):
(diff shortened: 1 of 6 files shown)


magma/types/bit_pattern.py
```diff
@@ -0,0 +1,55 @@
...
+        count = 0
+        for digit in pattern[1:]:
+            if not (digit == '_' or digit in string.whitespace):
+                if digit not in "01?":
+                    raise ValueError(
+                        f"Literal: {pattern} contains illegal character: {d}"
+                    )
+                mask = (mask << 1) + (0 if digit == "?" else 1)
+                bits = (bits << 1) + (1 if digit == "1" else 0)
+                count += 1
```
rsetaluri, line 38:
```suggestion
        for digit in pattern[1:]:
            if digit == '_' or digit in string.whitespace:
                continue
            if digit not in "01?":
                raise ValueError(
                    f"BitPattern {pattern} contains illegal character: {d}"
                )
            mask = (mask << 1) + (0 if digit == "?" else 1)
            bits = (bits << 1) + (1 if digit == "1" else 0)
            count += 1
```

magma/types/bit_pattern.py
```diff
@@ -0,0 +1,55 @@
...
+        self.bits = BitVector[count](bits)
+        self.mask = BitVector[count](mask)
+        self.width = count
+
+    def __eq__(self, other):
+        if not isinstance(other, Bits):
+            raise TypeError(
+                "BitPattern equality can only be compared with Bits value"
```
rsetaluri, line 46:
```suggestion
                "BitPattern can only be compared to Bits"
```

magma/types/bit_pattern.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+        return self.bits == (other & self.mask)
+
+    def as_bv(self):
+        if self.mask != ((1 << self.width) - 1):
```
rsetaluri, line 51: if this is something that's planned to be done often, might wanna cache this. Or could just wrap this in memoize decorator. If not a common op, doesn't matter.

leonardt: Not sure, but it's simple enough to just compute this once per instance so let's just do it now

PR #813, opened 2020-09-01
Add verilator compat compile option

leonardt:
Exposes https://github.com/rdaly525/coreir/pull/957

### DIFF
--- a/magma/backend/coreir_compiler.py
+++ b/magma/backend/coreir_compiler.py
@@ -35,6 +35,8 @@ def _make_verilog_cmd(deps, basename, opts):
         cmd += " --disable-width-cast"
     if opts.get("disable_ndarray", False):
         cmd += " --disable-ndarray"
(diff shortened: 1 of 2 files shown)


magma/backend/coreir_compiler.py
```diff
@@ -35,6 +35,8 @@ def _make_verilog_cmd(deps, basename, opts):
         cmd += " --disable-width-cast"
     if opts.get("disable_ndarray", False):
         cmd += " --disable-ndarray"
+    if opts.get("verilator_compat", False):
+        cmd += " --verilator_compat"
```
rsetaluri, line 39: Just a note that we would eventually want to add this to fast_compile version; requires plumbing all the way from here -> pycoreir -> CoreIR C API, so don't think we need to do it now since no one is using fast compile. Ideally we should unify these interfaces so we don't have to duplicate that logic (again a problem stemming all the way from CoreIR).

PR #815, opened 2020-09-01
Add support for tuple/bv/bool in mux operator

leonardt:


rsetaluri:
TL;DR: prefer moving out the list logic out of `mux()` and either (in order of ideal-ness) (a) promote using `m.tuple_()` to wrap list values, (b) provide separate function that does exactly that, or (c) ask users to do the list comprehension logic explicitly.

-----------------

Not a big fan of overloading `mux()` to take in a list like that, rather than allowing the user to metaprogram the mux call to do that on the outside. If anything we can provide a *separate* function like `multi_mux()` (or array_mux, or a better name 😄) that has this particular interface.

IMO, even better would be to use `m.tuple_()` function, like

```
        O0, O1, O2 = m.mux([
            m.tuple([True, BitVector[2](3), ht.Bit(0)]),
            m.tuple([False, BitVector[2](0), ht.Bit(1)])
        ], io.S).values()
```

This of course requires (a) `m.tuple_()` to be robust, and (b) the promotion logic from python-POD -> magma types to be the same in `mux()` and `tuple_()`. In fact, it would be good to consolidate that logic since it appears in a bunch of places actually -- but that's a separate point.

leonardt:
Perhaps an option that doesn't incur too much ergonomic overhead on the user (i.e. having to write these basic conversions themselves every time), I would put this in the category of `tuple -> Tuple` implicit conversion in the same way we do `int -> Bits` and `bool -> Bit`, so perhaps we can support using tuple literals (no need for explicit m.tuple_), but internally use the Tuple type logic (rather than list and special casing list support)

leonardt:
Ok, the latest rev now infers an m.Tuple when the user passes a python tuple, and returns an m.Tuple (so no special handling of the return value necessary, e.g. converting back into a python tuple like were doing with lists), so the only added logic is to infer a tuple type and conversion the arguments using m.tuple_, similar to how we handle inference based on BitVector/Bit (convert args to Bits and Bit)

leonardt:
I thought about this some more, and I propose we add a decorator used on all user facing primitive functions that consume magma or python values.  This decorator will capture the implicit coercion semantics for python values (int -> Bits/UInt/SInt, bool -> Bit, tuple -> Tuple) and can be reused to ensure consistency for all the places where this coercion can occur.  They should use the conversion function under the hood (m.bits, m.bit, m.tuple_), to ensure consistency with explicit user conversions as well as ensuring that these functions are robust in terms of Python value conversion.  With this decorator in place, primitive generators can simply operate on magma types, since we can run the implicit coercion logic before doing type inference (therefore avoiding the various cases of int/BV/tuple and instead assume everything will have a magma type)

However, I think we can do this as a separate change as this logic should work for now and is immediately useful for the mux operator.

magma/bit.py
```diff
@@ -143,14 +143,13 @@ def ite(self, t_branch, f_branch):
         f_type = get_type(f_branch)
 
         # Implicit int/bv conversion
-        if (issubclass(t_type, (int, ht.BitVector)) and
-                not issubclass(f_type, (int, ht.BitVector))):
+        if (issubclass(t_type, (int, ht.BitVector, ht.Bit)) and
```
rsetaluri, line 146: might be good to pull out `(int, ht.BitVector, ht.Bit)` as `_SELECT_TYPES` const or something, since it's gonna appear 4x and maybe subject to change. not critical

magma/conversions.py
```diff
@@ -286,11 +286,13 @@ def _tuple(value, n=None, t=Tuple):
             decl[k] = type(v)
     for a, d in zip(args, decl):
         # bool types to Bit
-        if decl[d] is bool:
+        if issubclass(decl[d], (bool, hwtypes.Bit)):
```
rsetaluri, line 289: :+1:

magma/primitives/mux.py
```diff
@@ -80,6 +81,32 @@ def _dynamic_mux_select(this, key):
 Array.dynamic_mux_select = _dynamic_mux_select
 
 
+def _infer_mux_type(args):
```
rsetaluri, line 84: 👍

tests/test_primitives/test_mux.py
```diff
@@ -174,3 +176,33 @@ class test_mux_operator_int(m.Circuit):
     tester.compile_and_run("verilator", skip_compile=True,
                            directory=os.path.join(os.path.dirname(__file__),
                                                   "build"))
+
+
+def test_mux_operator_tuple():
+    class test_mux_operator_tuple(m.Circuit):
+        io = m.IO(S=m.In(m.Bit), O0=m.Out(m.Bit), O1=m.Out(m.Bits[2]),
+                  O2=m.Out(m.Bit))
+        O0, O1, O2 = m.mux([
```
rsetaluri, line 185: does this unpack `m.Tuple` type automatically? or do we need to call `.values()` on the output first?

rsetaluri (approved):
Like this paradigm and definitely the idea of a single decorator. Agree it can be done later.

tests/test_primitives/test_mux.py
leonardt: It does by using the Iterable interface (which tuple values provide)

PR #826, opened 2020-09-08
Add right-hand operators to Bits/UInt/SInt

leonardt:
Fixes https://github.com/phanrahan/magma/issues/770

Needed to change the `const` exception to a TypeError so it would force
BitVector to return NotImplemented rather than raising an exception
(which then causes Python to try the right hand operator).

### DIFF
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -409,6 +409,9 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
         except TypeError:
             return NotImplemented
 
+    def __rand__(self, other):
+        return self & other
+
     def __or__(self, other):
         try:
             return self.bvor(other)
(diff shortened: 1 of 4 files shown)


magma/bits.py
```diff
@@ -542,7 +566,21 @@ def simulate(self, value_store, state_store):
 BitsType = Bits
 
 
-class UInt(Bits):
+class Number(Bits):
```
rsetaluri, line 569: pref `Int` over `Number`.

On that note, some of the other common circuit generator functions can prob be moved here, like `add` if we dispatch on `cls` (that way it will be still be specialized for `SInt` vs `UInt` but the logic is generic). Not critical to do that now.

rsetaluri (approved):
small note, lgtm

Issue #483, opened 2019-11-01
Slicing off the end of an array does not error


leonardt:
I believe this is because we dispatch to the Python standard slicing logic, which allows slicing off the end of a list, for example:
```
>>> a = [1, 2, 3]
>>> a[1:5]
[2, 3]
```

The proposal is to change the slicing behavior to be more strict and raise an error when the user tries to slice off the end of the array, rather than doing "implicit truncation" like Python's lists.

Let me know if you have objections to this change.

phanrahan:
Seems reasonable to me.

PR #827, opened 2020-09-08
Raise exception when slicing off end of array

leonardt:
Fixes https://github.com/phanrahan/magma/issues/483

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -335,8 +335,8 @@ class Array(Type, metaclass=ArrayMeta):
             return self.dynamic_mux_select(key)
         if isinstance(key, slice):
             if not _is_valid_slice(self.N, key):
-                raise IndexError(f"Trying to index array (type={type(self)}) "
-                                 f"with invalid slice: {key}")
+                raise IndexError(f"array index out of range "
+                                 f"(type={type(self)}, key={key})")
             _slice = [self[i] for i in range(*key.indices(len(self)))]
             return type(self)[len(_slice), self.T](_slice)
         else:
--- a/magma/array.py
+++ b/magma/array.py
@@ -303,6 +303,13 @@ class Array(Type, metaclass=ArrayMeta):
                  key[-1] == slice(None, len(self)) or
                  key[-1] == slice(0, len(self))))
 
+    def _is_valid_slice(self, key):
+        start, stop = key.start, key.stop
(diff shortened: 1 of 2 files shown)


magma/array.py
```diff
@@ -303,6 +303,13 @@ def _is_whole_slice(self, key):
                  key[-1] == slice(None, len(self)) or
                  key[-1] == slice(0, len(self))))
 
+    def _is_valid_slice(self, key):
+        start, stop = key.start, key.stop
+        if (((start is not None and (start >= self.N or start <= -self.N))) or
+                (stop is not None and (stop > self.N or stop < -self.N))):
+            return False
+        return True
+
```
rsetaluri, line 312: Prefer free function rather than member function, and pushed negate into condition, and direct return rather than `if`. Can be invoked as `_is_valid_slide(T.N, key)`.
```suggestion
def _is_valid_slice(n, key):
    start, stop = key.start, key.stop
    return ((start is None or start in range(-n + 1, n)) and
            (stop is None or stop not in range(-n, n + 1)))

```

magma/array.py
```diff
@@ -328,6 +335,9 @@ def __getitem__(self, key):
             # indexed using a dynamic magma value, generate mux circuit
             return self.dynamic_mux_select(key)
         if isinstance(key, slice):
+            if not self._is_valid_slice(key):
+                raise IndexError(f"Trying to index array (type={type(self)}) "
+                                 f"with invalid slice: {key}")
```
rsetaluri, line 340:
```suggestion
            if not self._is_valid_slice(key):
                raise IndexError(f"array index out of range "
                                 f"(type={type(self)}, key={key})")
```

rsetaluri (approved):
few code comments, lgtm

magma/array.py
leonardt: Using `in range` seems like a bad choice for performance, I think python will expand to the range to check the condition?

Here's a quick experiment for a small n
```
>>> timeit.timeit('3 in range(-64, 65)')
0.21609285899999975
>>> timeit.timeit('3 < -64 or 3 >= 65')
0.04527043200000236
```

rsetaluri: Interesting, would've presumed range doesn't eagerly expand for bounds checks. Thanks for verifying.

Fine using explicit comparisons over `range` but still prefer the remaining suggested changes.

PR #829, opened 2020-09-09
Smart bits improvements

rsetaluri:


magma/type_utils.py
```diff
@@ -0,0 +1,110 @@
+import dataclasses
```
rsetaluri, line 1: This file and `value_utils.py` are some common useful utilities that might be useful in other places as well. Let me know any feedback you have that might make these more useful.

magma/array.py
```diff
@@ -12,6 +12,7 @@
 from .debug import debug_wire, get_callee_frame_info
 from .logging import root_logger
 from magma.wire_container import WiringLog
+from magma.protocol_type import MagmaProtocol
```
rsetaluri, line 15: changes to this file and similar changes to tuple.py are for allowing protocol-type arguments to `Array` and `Tuple`.

magma/type_utils.py
```diff
@@ -0,0 +1,110 @@
...
+        constructor = lambda T: T
+        return _TypeWrapper(T, [], {}, constructor, "Bits")
+    if isarray(T):
+        constructor = lambda T, subT: T[T.N, subT]
+        return _TypeWrapper(T, [T.T], {"N": T.N}, constructor, "Array")
+    if istuple(T):
+
+        def _constructor(T, *args):
+            field_dict = {list(T.field_dict.keys())[i]: args[i]
+                          for i in range(len(T.field_dict))}
```
leonardt, line 81:
```suggestion
            field_dict = {key: arg for key, arg in zip(T.field_dict.keys(), args)}
```

magma/type_utils.py
```diff
@@ -0,0 +1,110 @@
...
+            field_dict = {k: v for k, v in field_dict.items() if v is not None}
+            unbound = _get_unbound_root(T)
+            return unbound.from_fields(T.__name__, field_dict)
+
+        attrs = {"fields": T.field_dict}
+        children = [value for value in T.field_dict.values()]
```
leonardt, line 87:
```suggestion
        children = list(T.field_dict.values())
```

magma/type_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    def generic_visit(self, T):
+        wrapped = _wrap_type(T)
+        for child in wrapped.children:
+            self.visit(child)
+
+
+class TypeTransformer(TypeVisitor):
+    def generic_visit(self, T):
+        wrapped = _wrap_type(T)
+        args = [self.visit(child) for child in wrapped.children]
```
leonardt, line 109:
```suggestion
        args = (self.visit(child) for child in wrapped.children)
```
Can just leave as a generator (no need to allocate list since it's unpacked in the next line)

magma/type_utils.py
leonardt: Avoids unpacking keys into a list for every item in the dict

magma/value_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    T = type(value)
+
+    if isdigital(T):
+        return _ValueWrapper(value, [], "Digital")
+    if isbits(T):
+        return _ValueWrapper(value, [], "Bits")
+    if isarray(T):
+        return _ValueWrapper(value, [t for t in value], "Array")
+    if istuple(T):
+        return _ValueWrapper(value, [t for t in value], "Tuple")
```
leonardt, line 35: Perhaps we could add a utility like `isdigitalval` or `isinstdigital` or something rather than having to fetch `type(value)` every time? Honestly this isn't so bad though and packing 3 words without underscores kinda hurts my eyes (`is_digital_value` wouldn't be so bad, but then we'd probably want to change everything to be consistent and it's not consistent with isinstance/issubclass)

magma/value_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    def visit(self, value):
+        wrapped = _wrap_value(value)
+        method = 'visit_' + wrapped.name
+        visitor = getattr(self, method, self.generic_visit)
+        return visitor(value)
+
+    def generic_visit(self, value):
+        wrapped = _wrap_value(value)
+        for child in wrapped.children:
+            self.visit(child)
```
leonardt, line 50: Seems like ValueVisitor and TypeVisitor could inherit from a base class with common visit/generic_visit logic and a class defined method `wrap` that dispatches to _wrap_value or _wrap_type.

magma/value_utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+    raise NotImplementedError(T)
+
+
+class ValueVisitor:
+    def visit(self, value):
+        wrapped = _wrap_value(value)
+        method = 'visit_' + wrapped.name
+        visitor = getattr(self, method, self.generic_visit)
+        return visitor(value)
```
leonardt, line 45: Hmm, this seems a little inconsistent with the TypeVisitor pattern, but not sure if this is intended.

Here we visit with wrapped value, while in TypeVisitor visits with the raw value.  The main potential wonkiness I see is that if we use generic_visit here, we'll wrap twice, not sure if that's intended?

magma/value_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        value = self._select(value)
+        if self.child is None:
+            return value
+        return self.child.select(value)
+
+    def _select(self, value):
+        return value
+
+    def _child_str(self):
+        return "" if self.child is None else str(self.child)
```
leonardt, line 67: Should this define __str__ method that prepends the value name before the child names?

leonardt: Actually, never mind it looks like this is only used in the case when there's no children (value is None)

leonardt (commented):
LGTM so far, I'm guessing you're testing with some non-public code?  Would be good to get some generic tests.

My summary (to check my understanding):
* Minor improvements to smart_bits (extension and optional resolution logic)
* make -> from_bits
* eval to force resolution at a certain point (by simply creating a temporary and wiring to it)
* make_smart logic to convert to smart bits
  * type visitor for converting Bits to SmartBits
  * value visitor for collecting the leaf references (used to wire up to the new smart bits value)
    * distinguish between bits and other values so we know when we're wiring up to a smart_bits and need to unwrap
  * selector logic makes it easy to "reselect" the same fields from the new smart_bits value using reference to the non smart value (nice job on this code, seems like a good design pattern)  for wiring non smart value to smart value

magma/smart/smart_bits.py
```diff
@@ -537,4 +556,47 @@ def _eval(lhs: SmartBits, rhs: _SmartExpr) -> (SmartBits, _SmartExpr):
...
+    leaf_collector.visit(value)
+    for bits in leaf_collector.bits_leaves:
+        selector = make_selector(bits)
+        smart_bits = selector.select(smart_value)
+        smart_bits_unwrapped = smart_bits._get_magma_value_()
+        smart_bits_unwrapped @= bits
+    for leaf in leaf_collector.other_leaves:
+        selector = make_selector(leaf)
+        new_leaf = selector.select(smart_value)
+        new_leaf @= leaf
```
leonardt, line 601: Do these need to be separate loops? If wiring handles the fetching of the magma protocol value, perhaps you don't need to handle smart bits separately?

magma/value_utils.py
rsetaluri: an alternative would be for these `is<x>` methods to be general for both values and types, so like:

```python
def _isdigital(T):
    return issubclass(T, Digital)

def isdigital(value_or_T):
    try:
        return _isdigital(value_or_T)
    except TypeError:
        pass
    value_or_T = type(value_or_T)
    try:
        return _isdigital(value_or_T)
    except TypeError:
        pass
    return False
```

then

```python
assert isdigital(Bit)
assert isdigital(Bit())
```

It is a bit overloaded and not clear but logically it just applies the same logic to the value or the type of the value. Let me know how it reads.

magma/value_utils.py
rsetaluri: Good call. I've done that.

magma/value_utils.py
rsetaluri: It's not clear because of the code but notice that here we still call `visitor(value)` (not `visitor(wrapped)`). So the code paths are the same - I should've used `wrapped.raw` here to be more consistent. Actually, I've pulled this all into a common class so it should be more clear now.

magma/smart/smart_bits.py
rsetaluri: Good point. I needed to augment `SmartBits.wire` to handle this case, which it should have.

rsetaluri:
@leonardt summary of PR is correct. Working on bringing in some public tests. Will have soon

rsetaluri:
@leonardt ready for re-review

magma/type_utils.py
```diff
@@ -0,0 +1,125 @@
...
+        def _constructor(T, subT):
+            return T[T.N, subT]
+
+        return _TypeWrapper(T, [T.T], "Array", {"N": T.N}, _constructor)
+    if istuple(T):
+
+        def _constructor(T, *args):
+            field_dict = {key: arg
+                          for key, arg in zip(T.field_dict.keys(), args)}
+            field_dict = {k: v for k, v in field_dict.items() if v is not None}
```
leonardt, line 105: Could move the `v is not None` filter into the above comprehension, but maybe this is split into two steps for readability?

leonardt (approved):
LGTM!

magma/value_utils.py
rsetaluri: Leaving as is for now.

Issue #828, opened 2020-09-09
Implicit coercion decorator for Python values


leonardt:
See https://github.com/phanrahan/magma/pull/815 for discussion

Summary: Provide a decorator that handles the implicit coercion of Python values (bool, int, Bit, BV, tuple) to the magma equivalent values using the conversion operators (m.bit, m.bits, m.tuple_).  This should standardize the implicit coercion logic for all user facing functions such as primitives and operators.  It should also simplify type inference since the primitives can then simply operator on magma types (since all python values will first be coerced to magma values).

PR #830, opened 2020-09-09
Add tuple conversion logic for wire

leonardt:
Follow up to https://github.com/phanrahan/magma/pull/815, adds the same python tuple conversion logic to the wire function (needed for a specific case of the mux primitive shown in the test).  Also had to do some rearranging of code to fix a circular import so I could bring the tuple_ function into wire.py

### DIFF
--- a/magma/backend/blif.py
+++ b/magma/backend/blif.py
@@ -5,7 +5,7 @@ from ..bitutils import seq2int, fun2seq
 from ..bit import Bit, Digital
 from ..array import Array
 from ..compiler import make_compiler
-from ..clock import wiredefaultclock
+from ..clock_utils import wiredefaultclock
 from ..is_definition import isdefinition
 
 __all__ = ['compile']
--- a/magma/backend/blif.py
+++ b/magma/backend/blif.py
@@ -5,7 +5,7 @@ from ..bitutils import seq2int, fun2seq
 from ..bit import Bit, Digital
 from ..array import Array
 from ..compiler import make_compiler
-from ..clock_utils import wiredefaultclock
+from ..clock import wiredefaultclock
 from ..is_definition import isdefinition
 
 __all__ = ['compile']
--- a/magma/backend/coreir_transformer.py
+++ b/magma/backend/coreir_transformer.py
@@ -20,7 +20,8 @@ from ..logging import root_logger
 from ..passes import InstanceGraphPass
 from ..tuple import Tuple
 from .util import get_codegen_debug_info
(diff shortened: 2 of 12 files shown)


leonardt:
Hmm, turns out my circular import fix didn't quite work and the dependencies are quite nasty.  I propose we just live with the import for now and fix it when we have https://github.com/phanrahan/magma/issues/828.  It's a bit tricky because wiring needs to import types to do the coercion, and types need to import wire for their internal recursive wiring logic.

I think what would do is define the common coerce function, then wrap the various wiring functions after the fact (we could monkey patch in the wrapped functions) to avoid the circular import.

rsetaluri:
Yeah this is a bit hairy. I think the fundamental problem is the dependencies that arise from recursive types.

As a simple example, the conversion (/promotion) function from a POD tuple to magma tuple would require the conversion from potentially underlying POD's to other types. So Tuple has to know about Array, and Array has to know about Tuple.

E.g. to convert `(True, BitVector[8](), 10)` to a tuple of type `Tuple[Bit, Bits, Bits]` there is a dependency chain Tuple -> Bits -> Array, but to convert an array of tuples, there is a dependency chain from Array -> Tuple.

Wiring is another example of this. The outermost wire call is `magma.wire` which contains all the conversions, and dispatch to the simple type wire member functions (e.g. `Tuple.wire`, `Bit.wire`). But if those member functions need to do conversions then they need to call back out to `magma.wire`, which may create cycles.

rsetaluri:
In general, I think there's a lot of similar logic sharded across wiring, conversions, and constructors. I think the right thing to do would be to have a proper standardization of this logic and put it in only "one" place (maybe distributed across the constructors, but not duplicate code paths).

rsetaluri:
However, given that that's a longer change I think this is a decent intermediate solution. Not too hacky.

rsetaluri (approved):

leonardt:
I think moving things from conversions to constructors seems like a reasonable move, then conversions are essentially just aliases for constructors.   I think the lowercase conversion syntax was chosen to mimic the python `int` and `bool` functions, but those actually act as types too (e.g. `isinstance(x, bool)`) so in some ways we could just view those as built-in classes with special lower case syntax (so defining m.Bit(1) is like bool(1)).  It's not quite the same because I think those functions do some magic in that they call the special methods __int__ and __bool__, but that could be an interesting option for that might be related to our protocol (e.g. define a _bits_ method to be convertible to Bits)

PR #832, opened 2020-09-09
Add support for user provided inline wire prefix

leonardt:
When magma is generating inline_verilog code, it assigns some values to temporary wires.  This change allows the user to choose a string for the wire names, useful for the case when these wires are used inside an `ifdef macro in the verilog.  

For example, an assertion may be wrapped inside an \`ifdef so it can be turned off for synthesis.  By providing a specific prefix for the wires entering into the \`ifdef block that may be unused (normally marked as an error by the linter), the user can provide a pattern matching string to the lint tool to ignore these dangling wires (i.e. ignore all dangling wires containing "_MAGMA_INLINE_ASSERTION_" in the name)

### DIFF
--- a/magma/inline_verilog.py
+++ b/magma/inline_verilog.py
@@ -41,10 +41,10 @@ def _make_inline_value(cls, inline_value_map, value):
     return f"{{{value_key}}}"
 
 
-def _make_temporary(defn, value, num, parent=None):
+def _make_temporary(defn, value, num, inline_wire_prefix, parent=None):
     with defn.open():
         # Insert a wire so it can't be inlined out
-        temp_name = f"_magma_inline_wire"
+        temp_name = inline_wire_prefix
         temp_name += f"{num}"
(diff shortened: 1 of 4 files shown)


magma/inline_verilog.py
```diff
@@ -41,10 +41,10 @@ def _make_inline_value(cls, inline_value_map, value):
     return f"{{{value_key}}}"
 
 
-def _make_temporary(defn, value, num, parent=None):
+def _make_temporary(defn, value, num, inline_wire_prefix, parent=None):
     with defn.open():
         # Insert a wire so it can't be inlined out
-        temp_name = f"_magma_inline_wire"
+        temp_name = inline_wire_prefix
```
rsetaluri, line 47: nit: can be a single line
`temp_name = f"{inline_wire_prefix}{num}`

rsetaluri (approved):

PR #833, opened 2020-09-09
Add support for user_namespace to bind

leonardt:
Bind logic needs to be aware that the generated module names will change
if the user_namespace option is present.

Also needed to change the generator cacheing to be use cache_definition
so the cache is reset between test invocations (issue revealed in the
bind test)

Also needed the coreir compiler to check if a namespace exists and use it before creating a new one (issue arises when using multiple invocations of compile with the same user_namespace)

### DIFF
--- a/magma/bind.py
+++ b/magma/bind.py
@@ -95,7 +95,6 @@ Bind monitor interface does not match circuit interface
     set_compile_dir("normal")
     # Circular dependency, need coreir backend to compile, backend imports
     # circuit (for wrap casts logic, we might be able to factor that out).
-    print(monitor, id(monitor))
     compile_fn(f".magma/{monitor.name}", monitor, inline=True,
                user_namespace=user_namespace)
     set_compile_dir(curr_compile_dir)
--- a/magma/bind.py
+++ b/magma/bind.py
@@ -40,7 +40,7 @@ def _wire_temp(bind_arg, temp):
     wire(bind_arg, temp)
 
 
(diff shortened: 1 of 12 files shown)


magma/bind.py
```diff
@@ -83,30 +83,32 @@ def _bind(cls, monitor, compile_fn, *args):
         bind_arg = temp
         ports += _gen_bind_port(cls, mon_arg, bind_arg)
     ports_str = ",\n    ".join(ports)
-    bind_str = f"bind {cls.name} {monitor.name} {monitor.name}_inst (\n    {ports_str}\n);"  # noqa
+    cls_name = cls.name
```
rsetaluri, line 86: is there any hope of not special casing this logic here and resolving the verilog name using a "symbol table"?

rsetaluri (approved):
Seems like a good candidate (whether for now or future) to try to use proper name resolution techniques. It seems like the real issue is verilog names could be arbitrarily by downstream and we're just "guessing." User namespace is just one example and maybe other modifications down the line and don't want to duplicate all that logic. Not sure how complicated that would be for this case.

Food for thought - good pattern for now.

magma/bind.py
leonardt: I think in this case the better long term solution is to just push the bind logic down into coreir, so the name resolution occurs there.  This is just a hot fix since FB is using user_namespace now and that broke the verification flow using bind.

PR #834, opened 2020-09-10
Fix bug with automatic wiring of temporary clock

leonardt:
This adds a test and fixes bugs related to automatic wiring of temporary
clocks.  The issue is that the automatic clock wiring logic assumed that
clocks were either driven or undriven at the sink of an instance.
However, in some case, the clock could be driven by a temporary value,
which itself isn't driven.  https://github.com/phanrahan/magma/pull/823
addressed one case where this issue arose, but did not cover the cases
presented in the added test.

First, the InsertWrapCasts logic had to be improved to traverse the
drivers of a port to find any instances where wrapping was necessary.
For example, we could have a temporary value of type Bit driving an
instance port of type Bit (no wrapping needed).  However, the driver of
the temporary Bit might be a clock, so a wrap needs to occur there.

The repr/insert_coreir_wire logic has been improved to handle anon bits
that are no driven (before it would try to call value on these to "skip"
them, but since they are undriven value would actually return the value
that it drives).

This changes the internal `driven` API to return True if a value is
driven (something is wired up to it), rather than True if trace() is
None.  The idea here is that this new version of `driven` is useful for
the above changes, while when we want to check whether a value can be
traced to a source driver, we can simply do trace() is None (whereas
there was no API to check if something is driven without tracing all the
way to a source).

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -665,13 +665,13 @@ class DefineCircuitKind(CircuitKind):
 
     def check_unconnected(self):
         for port in self.interface.inputs():
-            if not port.driven():
+            if port.trace() is None:
                 msg = f"Output port {self.name}.{port.name} not driven"
                 _logger.error(msg, debug_info=self.debug_info)
 
         for inst in self.instances:
             for port in inst.interface.inputs():
-                if not port.driven():
+                if port.trace() is None:
                     msg = f"Input port {inst.name}.{port.name} not driven"
                     _logger.error(msg, debug_info=inst.debug_info)
 
--- a/magma/passes/insert_coreir_wires.py
+++ b/magma/passes/insert_coreir_wires.py
@@ -104,11 +104,13 @@ class InsertCoreIRWires(DefinitionPass):
         if value in self.seen:
             return  # in the case of inouts, we may see more than once
         self.seen.add(value)
+        print(type(value))
         if not value.driven():
             return  # undriven value, skip wire insertion
         driver = value.value()
 
-        while (driver is not None and driver.name.anon() and
+        while (driver is not None and
+               (driver.name.anon() and driver.driven()) and
                not driver.is_output()):
             value, driver = driver, driver.value()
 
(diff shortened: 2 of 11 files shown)


magma/circuit.py
```diff
@@ -665,13 +665,13 @@ def __new__(metacls, name, bases, dct):
 
     def check_unconnected(self):
         for port in self.interface.inputs():
-            if not port.driven():
```
rsetaluri, line 668: why should this be trace vs driven?

magma/passes/insert_coreir_wires.py
```diff
@@ -25,8 +26,9 @@ def __init__(self, T):
         if issubclass(T, Digital):
             coreir_lib = "corebit"
             coreir_genargs = None
-            # Convert to bit so we wrap named types like clock if necessary
-            T = Bit
+            if issubclass(T, (AsyncReset, AsyncResetN, Clock)):
```
rsetaluri, line 29: what is the case where we don't want `T = Bit` ?

magma/passes/insert_coreir_wires.py
```diff
@@ -108,7 +110,8 @@ def _insert_wire(self, value, definition):
             return  # undriven value, skip wire insertion
         driver = value.value()
 
-        while (driver is not None and driver.name.anon() and
+        while (driver is not None and
```
rsetaluri, line 113: just noting that i see this same condition in `interface.py` above, and again below in this file. maybe a common function

rsetaluri (approved):

magma/circuit.py
leonardt: With the change to the driven() interface, a port can be driven by a temporary (driven() is True), but tracing it to a source returns None (it's not driven by an output)

magma/passes/insert_coreir_wires.py
leonardt: ResetN or Reset (or any other subclass of Bit that's not AsyncReset/Clock).  The reason that the above types are special cased are that they correspond to coreir named types.  reset/resetn for example can just be treated as normal coreir bit

PR #835, opened 2020-09-11
Minor bug fixes for riscv_mini CSR

leonardt:
* Support bool/bit in m.concat
* Support BitVector for Register init
* Include "_" separating unique numeric IDs for inline combinational
  (avoids name clobbering when we rollover, such as when there are more
  than 10 variables)

For the third bug, it's unfortunately difficult to create a small example to reproduce the issue (since it inherently only manifests with large programs), so I just factored out the necessary code from the riscv_mini csr as a test.  It's not essential that this stays up to date with the actual code, since all we need is a program complex enough to generate enough IDs to test the unique name generation logic (i.e. what the program does is not important, the fact that it compiles without error is important)

### DIFF
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -14,7 +14,6 @@ from .bfloat import BFloat
 from .digital import Digital
 from .tuple import Tuple, Product
 from .bitutils import int2seq
-import hwtypes
 import hwtypes as ht
 
 __all__ = ['bit']
@@ -104,7 +103,7 @@ def convertbits(value, n, totype, checkbit):
         return value
 
     convertible_types = (Digital, Tuple, Array, IntegerTypes,
-                         Sequence, hwtypes.BitVector)
+                         Sequence, ht.BitVector)
     if not isinstance(value, convertible_types):
         raise ValueError(
             "bits can only be used on a Bit, an Array, a Tuple, an int, or a"
@@ -115,7 +114,7 @@ def convertbits(value, n, totype, checkbit):
         if n is None:
             n = max(value.bit_length(), 1)
         ts = int2seq(value, n)
-    elif isinstance(value, hwtypes.BitVector):
+    elif isinstance(value, ht.BitVector):
         ts = value.bits()
     elif isinstance(value, Sequence):
         ts = list(value)
@@ -192,7 +191,7 @@ def bfloat(value, n=None):
 def concat(*arrays):
     ts = []
     for a in arrays:
-        if isinstance(a, hwtypes.BitVector):
+        if isinstance(a, ht.BitVector):
             ts.extend(a.bits())
         elif isinstance(a, Bit):
             ts.extend([a])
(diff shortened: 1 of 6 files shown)


leonardt:
Also adds logic to infer the type of the register from the init value (this adds some convenience when the user is specifying an init value, avoiding the need to specify both T and init when we can simply infer T from init)

magma/conversions.py
```diff
@@ -15,6 +15,7 @@
 from .tuple import Tuple, Product
 from .bitutils import int2seq
 import hwtypes
+import hwtypes as ht
```
rsetaluri, line 18: nit: stylistically, importing the same thing twice not preferred. think it's ok to refactor `hwtypes` to `ht` in the rest of this file if that's the preferred style going forward

magma/primitives/register.py
```diff
@@ -2,6 +2,7 @@
 
 import coreir
 import hwtypes
+import hwtypes as ht
```
rsetaluri, line 5: same as above

magma/conversions.py
```diff
@@ -195,6 +196,8 @@ def concat(*arrays):
             ts.extend(a.bits())
         elif isinstance(a, Bit):
             ts.extend([a])
+        elif isinstance(a, (bool, ht.Bit)):
+            ts.extend([Bit(a)])
         else:
```
rsetaluri, line 201: also just noting it because I see it now, not sure how robust this else is 😬

tests/test_syntax/test_inline_comb.py
```diff
@@ -136,3 +137,385 @@ def logic():
     tester.compile_and_run("verilator", skip_compile=True,
                            directory=os.path.join(os.path.dirname(__file__),
                                                   "build"))
+
+
+def test_csr_comb():
```
rsetaluri, line 142: Not disputing that it is so, but unclear to me why a simpler test example is difficult to construct? If my understanding is correct the issue was triggered when there were more than 10 new variables -- can this be deterministically hit in the case of 10 or more `elif`s? Or 10 or more separate variables?

If nothing else, would be good to change the name of the test to something more representative of the error, ala `test_many_name_gen` and add a comment at top of this function saying like "tests for issue ____ triggered when ____. actual circuit contents don't matter, just need complex circuit to trigger condition"

rsetaluri (approved):

tests/test_syntax/test_inline_comb.py
leonardt: Actually I think it needs 11 separate variables and 11 separate elifs.  In this situations, we could get a name with 111 which could be case number 11 and variable 1 or case number 1 and variable number 11.

I can try to create a contrived example that exercises the problem, I think I originally just thought that the contrived example would have to be so large/complex that there wasn't much value in creating it versus just using the code that was already written that produced the problem.

magma/conversions.py
leonardt: I think it assumes that `a: Array` and the above cases handle the implicit coercion to an array type.  We can add a type error if it's not an array.

tests/test_syntax/test_inline_comb.py
leonardt: Confirmed it's triggered with 11 variables and 11 elifs (because ssa appends a unique ID for every variable, so we'll get the 111 ambiguous case above without the "_" ending the prefix).  I added a test case that just has 11 variables and 11 cases that assign various values and triggers the type error induced by this problem.

rsetaluri (approved):
Looks good, thanks for addressing comments

PR #836, opened 2020-09-14
Add andR, orR, and xorR methods

leonardt:
These are convenience methods for m.reduce(<op>, <val>).  It's a common
operation and it's significantly less verbose to just call one of these
methods rather than having to do the full m.reduce and operator.<op>
style

To avoid a circular dependency between reduce.py and bits.py, I moved
the reduce logic into bits.  This seems reasonable since the other bits
related primitives are already housed there.


magma/__init__.py
```diff
@@ -100,6 +100,6 @@ def set_mantle_target(t):
 from .syntax.coroutine import coroutine
 
 from magma.primitives import (LUT, Mux, mux, Register, get_slice, set_slice,
-                              slice, reduce, Memory)
+                              slice, Memory)
```
leonardt, line 103: Import is now covered by the existing `from .bits import *`

rsetaluri: 😂

magma/bits.py
```diff
@@ -541,6 +543,15 @@ def __getitem__(self, index):
             return (self >> index)[0]
         return super().__getitem__(index)
 
+    def orR(self):
```
rsetaluri, line 546: unless these names are standard or mirror the names in some other system (say, Chisel), I think doing `<op>_reduce` or `reduce_<op>` looks better :)

rsetaluri (approved):

magma/bits.py
rsetaluri: Ok, I see the corresponding CoreIR primitives are `orr`, `xorr`, `andr`. I still think the function name can be the full word since this is a user-exposed function, and the CoreIR is mostly backend.

leonardt: Chisel uses orR, xoR, ... (that's where I got them), we could provide both.  One one hand, orR is significantly easier to type, on the other hand, maybe we shouldn't sacrifice ease of typing for readability (since autocomplete is prevalent these days), but I don't think orR, xorR, ... is too ambiguous to read easily.

rsetaluri: I think having both would be the least preferred option.

Personally prefer being on the side of the trade-off which is more clear and readable (also follows "code is the documentation" principle). But fine having `orR` etc. especially if it's in Chisel. Your call.

rsetaluri (approved):

PR #839, opened 2020-09-14
Add logic to infer Register type from init

leonardt:
Pulling into https://github.com/phanrahan/magma/pull/835 since it touches the same lines in the register changes.

This adds some convenience when the user is specifying an init value, avoiding the need to specify both T and init when we can simply infer T from init

### DIFF
--- a/magma/primitives/register.py
+++ b/magma/primitives/register.py
@@ -6,7 +6,7 @@ import hwtypes as ht
 
 from magma.array import Array
 from magma.bit import Bit
-from magma.bits import Bits
+from magma.bits import Bits, UInt, SInt
 from magma.circuit import coreir_port_mapping
 from magma.conversions import as_bits, from_bits
 from magma.interface import IO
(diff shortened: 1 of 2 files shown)


magma/primitives/register.py
```diff
@@ -112,14 +129,20 @@ def __init__(self, T: Kind, init: Union[Type, int, ht.BitVector] = None,
...
+        if T is None:
+            if init is None:
+                raise ValueError("User must provide type T or init value (from"
+                                 " which T will be inferred)")
+            T = _get_T_from_init(init)
         if not isinstance(T, Kind):
             raise TypeError(
                 f"Expected instance of Kind for argument T, not {type(T)}")
         if init is not None and not isinstance(init,
-                                               (Type, int, ht.BitVector)):
```
rsetaluri, line 119: another option would be to `try` `_get_T_from_init(init)` and if there isn't a `ValueError` then this is valid. Benefit is not having to update the set of valid types for init in 2 places. Added benefit is getting early type checking on compatibility of init and T (if T is specified). Downside is whatever the cost of calling `get_T_from_init` is (not big but not 0).

rsetaluri (approved):

magma/primitives/register.py
leonardt: I tweaked it slightly so that it only checks the type of init matches T when T and init are both provided.  When only init is provided, we do the _get_T_from_init logic to infer the type (so we know they'll match).  When both are provided, we use _get_T_from_init to infer the type based on init, and make sure it is compatible with the user provided T.

PR #840, opened 2020-09-15
Update bind for coreir ndarray support

leonardt:
Bind logic is now aware that coreir will generate ndarrays

### DIFF
--- a/magma/bind.py
+++ b/magma/bind.py
@@ -16,14 +16,20 @@ from magma.inline_verilog import _get_view_inst_parent
 
 
 def _gen_bind_port(cls, mon_arg, bind_arg):
-    if isinstance(mon_arg, Tuple) or isinstance(mon_arg, Array) and \
-            not is_nd_array(type(mon_arg)):
+    if (isinstance(mon_arg, Tuple) or isinstance(mon_arg, Array) and not
+            is_nd_array(type(mon_arg))):
         result = []
         for child1, child2 in zip(mon_arg, bind_arg):
             result += _gen_bind_port(cls, child1, child2)
         return result
-    port = value_to_verilog_name(mon_arg)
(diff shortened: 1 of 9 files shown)


magma/bind.py
```diff
@@ -16,14 +16,20 @@
...
+            is_nd_array(type(mon_arg))):
         result = []
         for child1, child2 in zip(mon_arg, bind_arg):
             result += _gen_bind_port(cls, child1, child2)
         return result
-    port = value_to_verilog_name(mon_arg)
-    arg = value_to_verilog_name(bind_arg)
+    port = value_to_verilog_name(mon_arg, False)
+    disable_ndarray = False
+    if (isinstance(mon_arg, Array) and bind_arg.name.root() is not None):
```
rsetaluri, line 27: nit: don't need parens

magma/verilog_utils.py
```diff
@@ -8,43 +8,57 @@
 from magma.view import InstView, PortView
 
 
-def value_to_verilog_name(value):
-    if isinstance(value, Array) and not issubclass(value.T, Digital):
-        elems = ", ".join(value_to_verilog_name(t) for t in reversed(value))
+def is_nd_array(T, skip=True):
```
rsetaluri, line 11: what is the meaning of skip? Looks like for the top level call we're allowed to return true if digital? So the top level call for `is_nd_array(Bit)` returns true? Why is this desired?

rsetaluri (approved):
LGTM. Just a nit and a question for my clarification.

magma/verilog_utils.py
leonardt: Actually it's the other way around `is_nd_array(Bit)` will return False (since skip defaults to True), so we require at least one level of array before we a see a Bit and return True.

PR #842, opened 2020-09-15
Add dict_lookup and list_lookup helpers

leonardt:
Convenience functions for easily generating muxes to lookup items in a
dictionary or list


rsetaluri (approved):
Nice functionality and implementation. A few comments:

I think the interface would be better with `default` as the last argument, with a default value of 0 or something (can be encoded as `None`). This mirrors the python dict method `dict.get(key, default=None)`. So I would propose `dict_lookup(dict_, select, default=None)`.

These are *somewhat* similar to CAMs but not quite the same. Will be interesting to implement those if they come up in a use case.

PR #843, opened 2020-09-15
Restrict inline_combinational to use wiring

### LINKED ISSUE #812: Altering locals() not guaranteed to work
Amazingly, we've been able to alter the locals() dictionary without problem, but I recently encountered a problem where it wasn't working.  After some browsing I found that it's actually documented that changing the value of locals() is not guaranteed to work: https://docs.python.org/3/library/functions.html#locals (cc @cdonovick), so even passing it from the user scope won't save us here.

> Note The contents of this dictionary should not be modified; changes may not affect the values of local and free variables used by the interpreter.

We'll need to find a different pattern for supporting our syntaxes that don't involve modifications to locals.

leonardt:
Avoids the problem discussed in #812

Since we cannot guarantee that the assignment values can be used in the
outer scope, we restrict that the user use wiring.  If the user wants a
value to "escape" to the outer scope, they can create a temporary value
and wire up to it.

Also fixes a wiring bug in the inline comb logic (it should be wire
driver to drivee, this matters for using temporary values)

Fixes #812

### DIFF
--- a/docs/inline_combinational.md
+++ b/docs/inline_combinational.md
@@ -15,18 +15,23 @@ Here's a simple example:
 
 ```python
 class Main(m.Circuit):
-    io = m.IO(invert=m.In(m.Bit), O=m.Out(m.Bit))
+    io = m.IO(invert=m.In(m.Bit), O0=m.Out(m.Bit), O1=m.Out(m.Bit))
     io += m.ClockIO()
     reg = m.Register(m.Bit)()
 
-    @m.inline_combinational()
+    O1 = m.Bit()
+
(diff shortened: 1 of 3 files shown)


rsetaluri (approved):
Looks like a decent solution. One question I have is regarding the parity between this syntax and normal comb. syntax? Are they fundamentally different or is this just slightly restricted?

leonardt:
The syntax should be the same except the handling of return values.  Inline comb doesn't use return values, instead you wire up directly to things in the enclosing scope.  Comb you return values and you can use it as a circuit using the function call interface.  But the way if statements are handled are the same.

PR #846, opened 2020-09-16
Make drive_undriven be auto clock wiring aware

leonardt:
Before, using the drive_undriven option would cause clocks to be marked
as undriven, which would then block the automatic clock wiring logic
from being performed later in the backend.  This issue was introduced
when we moved the automatic clock wiring logic to be done in the backend
rather than as a pass (which could be done before the drive_undriven
pass).

To avoid re-introducing a performance hit from having a generic clock
wiring pass, this adds the automatic clock wiring logic to the
drive_undriven pass (since we're already traversing the instances at
this point, we can do the automatic clock wiring logic and then skip the
undriven marker if it's a clock that's being automatically wired).

I think this is a reasonable option for supporting auto clock wiring
with the drive_undriven pass without having to incur the cost of a full
clock wiring pass.  Another option would be to just skip marking clocks
all together, which has the downside of triggering an "undriven error"
downstream which seems like odd behavior given the semantics of this
pass.  With this, we will mark clocks as undriven if they are not wired
automatically.

### DIFF
--- a/magma/backend/coreir_transformer.py
+++ b/magma/backend/coreir_transformer.py
@@ -6,11 +6,10 @@ import os
 from ..digital import Digital
 from ..array import Array
 from ..bits import Bits
-from ..clock import ClockTypes
 from coreir import Wireable
 from .coreir_utils import (attach_debug_info, check_magma_interface,
                            constant_to_value, get_inst_args,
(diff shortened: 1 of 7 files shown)


rsetaluri (approved):
Nice! looks cleaner now

PR #847, opened 2020-09-16
Improve error message for right hand operators

leonardt:
This adds logic to the right operator definitions to return
NotImplemented if they encounter a TypeError that is raised in
implementation.  As a result, you get a nice error message such as:
```
>       io.evec @= mtvec + (PRV.O << 6)
E       TypeError: unsupported operand type(s) for +: 'BitVector[32]' and 'UInt[(2, Out(Bit))]'
```
Note this is the default error message from Python when both __add__ and
__radd__ return NotImplemented.  It uses the type name which has the
redundant Out(Bit) inside UInt which I will look into fixing in a
separate PR.

Before this change, we would instead get a message such as
```
    def test_radd_type_error():
>       class Main(m.Circuit):

tests/test_type/test_bits.py:404:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_type/test_bits.py:407: in Main
    BitVector[32](0xDEADBEEF) + io.I
magma/bits.py:494: in __radd__
    return self + other
magma/bits.py:487: in __add__
    return self.bvadd(other)
magma/bits.py:51: in wrapped
    other = _coerce(type(self), other)
magma/bits.py:41: in _coerce
    return T(val)
magma/bits.py:63: in __call__
    result = super().__call__(*args, **kwargs)
../hwtypes/hwtypes/bit_vector_abc.py:28: in __call__
    return super().__call__(value, *args, **kwargs)
magma/bits.py:111: in __init__
    Array.__init__(self, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = bits(0, 2), args = ([1, 1, 1, 1, 0, 1, ...],), kwargs = {}

    def __init__(self, *args, **kwargs):
        Type.__init__(self, **kwargs)
        self.ts = []
        if args:
            if len(args) == 1 and isinstance(args[0], (list, Array, int,
                                                       BitVector)):
                if isinstance(args[0], list):
                    if len(args[0]) != self.N:
>                       raise ValueError("Array list constructor can only be used "
                                         "with list equal to array length")
E                       ValueError: Array list constructor can only be used with list equal to array length

```

Note that the problematic source line is buried deep inside the stack
trace leading to a confusing error message.

As a part of this change, I improved the Array constructor to raise
TypeError rather than a ValueError (as shown above) when encountering a
BitVector initializer with an inconsistent size (so it would be caught
by the NotImplemented logic).

### DIFF
--- a/magma/bits.py
+++ b/magma/bits.py
@@ -28,13 +28,18 @@ from magma.generator import Generator2
 _logger = root_logger()
 
 
+def _check_size(val, T):
+    if len(val) != len(T):
+        raise InconsistentSizeError('Inconsistent size')
+
+
 def _coerce(T: tp.Type['Bits'], val: tp.Any) -> 'Bits':
     if isinstance(val, ht.BitVector):
+        _check_size(val, T)
         val = val.bits()
     if not isinstance(val, Bits):
         return T(val)
-    if len(val) != len(T):
-        raise InconsistentSizeError('Inconsistent size')
+    _check_size(val, T)
     return val
 
 
@@ -412,7 +417,10 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
             return NotImplemented
 
     def __rand__(self, other):
-        return self & other
+        try:
+            return self & other
+        except TypeError:
+            return NotImplemented
 
     def __or__(self, other):
         try:
@@ -423,7 +431,10 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
             return NotImplemented
 
     def __ror__(self, other):
-        return self | other
+        try:
+            return self | other
+        except TypeError:
+            return NotImplemented
 
     def __xor__(self, other):
         try:
@@ -434,7 +445,10 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
             return NotImplemented
 
     def __rxor__(self, other):
-        return self ^ other
+        try:
+            return self ^ other
+        except TypeError:
+            return NotImplemented
 
     def __lshift__(self, other):
         try:
@@ -445,7 +459,10 @@ class Bits(Array, AbstractBitVector, metaclass=BitsMeta):
             return NotImplemented
 
     def __rlshift__(self, other):
-        return type(self)(other) << self
+        try:
+            return type(self)(other) << self
+        except TypeError:
+            return NotImplemented
 
     def __rshift__(self, other):
         try:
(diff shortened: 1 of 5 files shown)


magma/bits.py
```diff
@@ -412,7 +417,10 @@ def __and__(self, other):
             return NotImplemented
 
     def __rand__(self, other):
-        return self & other
+        try:
```
rsetaluri, line 420: would consider using decorator like

```python
def _not_implemented_on_type_error(fn):

    @functools.wraps(fn)
    def _wrapper(*args, **kwargs):
        try:
            return fn(*args, **kwargs)
        except TypeError:
            return NotImplemented

    return _wrapper
```

Actually, could make it even more general if wanted (and put in `common.py`):

```python
def return_on_exception(func=None, *, exceptions=(Exception,), return_value=None):
    if func is None:
        return functools.partial(return_on_exception, exceptions=exceptions, return_value=return_value)

    @functools.wraps(func)
    def _wrapper(*args, **kwargs):
        try:
            return func(*args, **kwargs)
        except exceptions:
            return return_value

    return _wrapper
```

This could then be used like

```python
@return_on_exception((TypeError,), NotImplemented)
def __rand__(self, other):
    ...
```

(could also do like `roperator_decorator = return_on_exception((TypeError,), NotImplemented)` for shorthand on each op)

rsetaluri (approved):
lgtm

magma/bits.py
leonardt: I think there was some reason we couldn't use a decorator (this is the pattern that hwtypes uses as well), but let me see if I can remember

leonardt: Hmm couldn't seem to remember/find any comments on the old related PRs.

I used the first decorator approach since I also refactored the
```
        except InconsistentSizeError as e:
            raise e from None
```
and this wasn't quite as easy to capture as function arguments (couldn't put it in a lambda)

PR #849, opened 2020-09-17
Improve mux inference

leonardt:
* Ensures that UInt/SInt is inferred from UIntVector/SIntVector inputs
* Adds upcasting logic so that the parent type is used when encountering
  two arguments where the type of one is the subclass of the other
* Improves mux type checking to use wireability in the case that the
  types don't match exactly (avoids problem with structurally equal
  types but different nominal types, e.g. when namedtuple is used)

As a side effect of this change, some gold files are updated since the 
inferred type doesn't include directionality for clarity.


magma/primitives/mux.py
```diff
@@ -84,18 +85,43 @@ def _infer_mux_type(args):
...
+        elif isinstance(arg, int):
+            # Cannot infer type without width, use wiring implicit coercion to
+            # handle (or raise type error there)
+            continue
+
+        if T is not None:
+            if issubclass(T, next_T):
+                # upcast
+                T = next_T
+            elif not wireable(next_T, T):
```
rsetaluri, line 111: nice!

magma/wireable.py
```diff
@@ -0,0 +1,19 @@
+from .array import Array
+from .tuple import Tuple
+from .protocol_type import magma_type
```
rsetaluri, line 3: nit (since new file :) )

```suggestion
from magma.array import Array
from magma.tuple import Tuple
from magma.protocol_type import magma_type
```

rsetaluri (approved):
Looks like this doesn't functionally change the verilog but changes the output verilog nominally (generator instance names different cuz of different types). Given the garnet PD flow **may** depend on some names, I think it would be a good idea to run aha-flow tests on this change. Can definitely help with that.

leonardt:
@rsetaluri have you had a chance to check if there's any issues with the PD flow?

rsetaluri:
@leonardt I've kicked of aha flow (compiler flow tests): https://github.com/StanfordAHA/aha/tree/magma-mux-infer-intv

For PD I'm pinging relevant folks. Will have an answer soon.

rsetaluri:
@leonardt software flow works. Haven't heard from PD. Let's merge this and see. Shouldn't be too bad to figure it out post if need be.

PR #853, opened 2020-09-24
Restrict Clock wiring

leonardt:
Fixes https://github.com/phanrahan/magma/issues/651

Enforces nominal typing for clock types. Clocks (and
resets/enables/etc...) can only be wired to values that are instances of
the same type (i.e. not instances of Digital).

Includes updates to portions of the code base that were relying on the
ability to wire Digital/Bit to clock types.  Mainly the change requires
using an explicit `m.clock` or `convertbit(value, T)` (for a generic
conversion of value to T) to connect a bit to a clock type.

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -5,6 +5,7 @@ from .digital import DigitalMeta, Digital
 from .wire import wire
 from magma.bit import Bit
 from magma.array import Array
+from magma.debug import debug_wire
 from magma.tuple import Tuple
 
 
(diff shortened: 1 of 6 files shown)


magma/clock.py
```diff
@@ -19,6 +20,16 @@ def unused(self):
     def undriven(self):
         Bit.undriven(self)
 
+    @debug_wire
```
rsetaluri, line 23: this kind of an interesting pattern. if we think of the inheritance as a tree, the default is for any nodes of the tree (types) to be wireable if they share a common ancestor, but sometimes we want to break that and say two types have to have a specific common ancestor, and we basically do that by overriding `def wire()` to do this check. The interesting thing is that if we wanted apply the same thing to other types the code would be identical, but we can't get reuse through inheritance b/c that's the problem in the first place!

Just an observation.

rsetaluri (approved):

Issue #820, opened 2020-09-04
getitem/setitem syntax with memory+combinational


leonardt:
Chisel provides a getitem/setitem syntax for their memory primitive that's quite useful when combined with their when statement.  In this example: https://github.com/ucb-bar/chisel-tutorial/blob/release/src/main/scala/examples/Router.scala#L65-L83

the `tbl` memory only needs one read port and one write port (compatible with the coreir mem primitive), but the read port is shared between two different cases in the when statement (mux on the raddr).  In magma, we have to do this mux explicitly: https://github.com/leonardt/magma_examples/blob/be5826641c30ebe5279b00e205919ed61475b925/magma_examples/router.py#L62-L67

It would be nice we could support this pattern in combinational.

Also the write enable logic is inferred from the when statement (https://github.com/ucb-bar/chisel-tutorial/blob/release/src/main/scala/examples/Router.scala#L72), again it would be useful if we could use setitem inside a combinational block and have it infer the enable logic.

When using the current memory primitive, we could error if the user logic would require more than one read/write port.  It would be nice if we could also support the RegFileBuilder which would allow arbitrary numbers of read/write ports, but then this would probably motivate this being a primitive rather than as part of mantle (since we'll have special logic in the syntax to support it).  Well, maybe not, since we could just translate this into read/write port wiring and raise an exception depending on what we're trying to wire up to (the Memory primitive would error, the RegFileBuilder wouldn't, so instead of having it as another primitive, we could instead define an abstract interface based on getitem/setitem and the memory primitive would enforce the port limit when it is being used improperly).

PR #854, opened 2020-09-24
Add __setitem__/__getitem__ support for memory primitive

leonardt:
Fixes https://github.com/phanrahan/magma/issues/820

For now, we provide a warning when the read/write ports are already
driven.  When we add support for the conditional wiring syntax, we can
improve this check to handle the conditional wiring cases.

### DIFF
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -9,8 +9,12 @@ from magma.clock_io import ClockIO
 from magma.conversions import from_bits, as_bits
 from magma.generator import Generator2
 from magma.interface import IO
+from magma.logging import root_logger
 from magma.primitives.register import Register
 from magma.t import In, Out, Kind
+
+
+_logger = root_logger()
 
 
 class CoreIRMemory(Generator2):
@@ -106,3 +110,25 @@ class Memory(Generator2):
         for _ in range(read_latency - 1):
             rdata = Register(T)()(rdata)
         self.io.RDATA @= rdata
+
+        def __setitem__(self, addr, data):
+            if self.WADDR.driven() or self.WDATA.driven():
+                _logger.warning(
+                    "Calling __setitem__ on a Memory instance with WADDR or"
+                    " WDATA already driven, will overwrite previous values"
+                )
+            self.WADDR @= addr
+            self.WDATA @= data
+
+        self.__setitem__ = __setitem__
+
+        def __getitem__(self, addr):
+            if self.RADDR.driven():
+                _logger.warning(
+                    "Calling __getitem__ on a Memory instance with RADDR"
+                    " already driven, will overwrite previous values"
+                )
+            self.RADDR @= addr
+            return self.RDATA
+
+        self.__getitem__ = __getitem__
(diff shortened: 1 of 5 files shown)


tests/test_primitives/test_memory.py
```diff
@@ -19,10 +19,8 @@ class test_memory_basic(m.Circuit):
 
         )
         Mem4x5 = m.Memory(4, m.Bits[5])()
-        Mem4x5.RADDR @= io.raddr
-        io.rdata @= Mem4x5.RDATA
-        Mem4x5.WADDR @= io.waddr
-        Mem4x5.WDATA @= io.wdata
+        io.rdata @= Mem4x5[io.raddr]
+        Mem4x5[io.waddr] = io.wdata
```
leonardt, line 23: One thought, here we use `=` since it dispatched to __setitem__, alternatively we could look into `__getitem__` returning something smart so that we could `Mem4x5[io.waddr] @= io.wdata` as the syntax, but this is a bit tricky since we'll need to distinguish between cases where we're doing reads and writes.  The benefit would be we would not need to introduce `=` in this special context (I think ideally we could just state that `@=` is alway used for working with magma values)

magma/primitives/memory.py
```diff
@@ -106,3 +110,25 @@ def __init__(self, height, T: Kind,
...
+        def __getitem__(self, addr):
+            if self.RADDR.driven():
+                _logger.warning(
+                    "Calling __getitem__ on a Memory instance with RADDR"
+                    " already driven, will overwrite previous values"
+                )
+            self.RADDR @= addr
+            return self.RDATA
+
+        self.__getitem__ = __getitem__
```
leonardt, line 134: Wasn't too happy with how this had to be done, I wonder if we can come up with something better for defining methods on generated Circuits.  Perhaps we could define the methods inside the Generator2 subclass body, and mark them as `@circuitmethod` or something like that?

tests/test_primitives/test_memory.py
rsetaluri: That sounds nice. One idea would be to return a special object which stages the decision of read or write. This special object could look like a normal value when something wires to it (and then do the read op.). If it is wired to, then it would be treated as read (and we could ensure it isn't written to again). Probably some hiccups to what I just proposed but just a thought.

magma/primitives/memory.py
rsetaluri: If I understand correctly the issue is having member functions on the the instances? And right now a `def foo(self)` would be a member function on the module (/circuit). So the only way around it is to do `self.foo = fn`. Just trying to make sure I understand.

I think a decorator would work well. Let me think about it. It's kind of a weird problem and probably where the "elegance" of the solution breaks down.

leonardt: Understanding is correct (defining a method here defines it on the class instead of the instance)

rsetaluri: Ok. I have an idea of how to do it similar to your idea. If you want to merge this, I can make a PR shortly that adds this feature and updates this memory generator to use it.

rsetaluri (approved):

PR #857, opened 2020-10-01
Improve error message when calling == or != on input values

leonardt:
Before, in some cases, the operator would return True or False rather
than raising an error.  We probably want to make this consistent for all
operators, but I think this is an important hotfix since the current
return value (bool rather than a magma value) can cause downstream
errors that are confusing and making tracing the root cause harder
(might get some typeerror related to the bool returned, which might lead
you to start debugging at the wrong place)

### DIFF
--- a/magma/operator_utils.py
+++ b/magma/operator_utils.py
@@ -4,7 +4,7 @@ import functools
 def output_only(msg):
 
     def _wrapper(fn):
-    
+
         @functools.wraps(fn)
         def _wrapped(self, *args, **kwargs):
             if self.is_input():
--- a/magma/operator_utils.py
+++ b/magma/operator_utils.py
@@ -1,0 +1,12 @@
+import functools
+
+
+def output_only(err_msg):
+    def wrapper(fn):
+        @functools.wraps(fn)
+        def wrapped(self, *args, **kwargs):
+            if self.is_input():
+                raise TypeError(err_msg)
+            return fn(self, *args, **kwargs)
+        return wrapped
+    return wrapper
--- a/magma/operator_utils.py
+++ b/magma/operator_utils.py
@@ -1,12 +1,16 @@
 import functools
 
 
-def output_only(err_msg):
-    def wrapper(fn):
+def output_only(msg):
+
+    def _wrapper(fn):
+    
         @functools.wraps(fn)
-        def wrapped(self, *args, **kwargs):
+        def _wrapped(self, *args, **kwargs):
             if self.is_input():
-                raise TypeError(err_msg)
+                raise TypeError(msg)
             return fn(self, *args, **kwargs)
-        return wrapped
-    return wrapper
+
+        return _wrapped
+
+    return _wrapper
--- a/magma/array.py
+++ b/magma/array.py
@@ -13,6 +13,7 @@ from .debug import debug_wire, get_callee_frame_info
 from .logging import root_logger
 from .protocol_type import magma_type, magma_value
 
+from magma.operator_utils import output_only
 from magma.wire_container import WiringLog
 from magma.protocol_type import MagmaProtocol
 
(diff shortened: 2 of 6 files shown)


leonardt:
One thing to consider is this raises a TypeError that stops execution.  If we want to preserve the model of collecting multiple errors before moving on, then perhaps we should make this use the staged logger instead.  The issue with the current error message is you'll get something like "Error: Using inst.O an Input as an Output" when trying to wire to the underlying operator created instances.  A better error message would be something like "Error: Using == with inst.O (operators cannot be used on inputs, only outputs)"

magma/operator_utils.py
```diff
@@ -0,0 +1,12 @@
...
+
+def output_only(err_msg):
+    def wrapper(fn):
+        @functools.wraps(fn)
+        def wrapped(self, *args, **kwargs):
+            if self.is_input():
+                raise TypeError(err_msg)
+            return fn(self, *args, **kwargs)
+        return wrapped
+    return wrapper
```
rsetaluri, line 12:
```suggestion
def output_only(msg):

    def _wrapper(fn):
    
        @functools.wraps(fn)
        def _wrapped(self, *args, **kwargs):
            if self.is_input():
                raise TypeError(msg)
            return fn(self, *args, **kwargs)

        return _wrapped

    return _wrapper
```

rsetaluri (commented):
This is a good change to improve debuggability of the relevant errors.

While we're here, should we think about standardizing the interfaces of these methods? For example some of them return `False` on a wrong type. Some return `NotImplemented`. Sometimes you get a `TypeError`?

If you ask me, I think these should never return a python `bool`, always a `magma.Bit` (that might be const). And I think all the *type* checks should return `NotImplementedError`, and the direction checks (`output_only`) can raise a `TypeError`.

Also unclear why a staged error is necessary. It seems we have all the information we need here to provide a readable message (`"Cannot use == on x (an input)"`)? Yes, staging would be good for collecting all the errors, but the bigger reason for staging IMO is the whole lazy circuit business.

leonardt:
Sorry staging was probably not the right word, all though staging is probably useful for getting the Circuit name when it's involved (e.g. using a circuit output incorrectly should give `Foo.O` instead of `.O` or `LazyCircuit.O`).  However, in my original comment I was actually thinking about error messages versus exceptions (i.e. our current pattern of logging errors rather than raising exceptions and stopping compilation, so we can collect multiple errors at once)

leonardt:
Here's a thought:
If we want to allow multiple error messages to be logged, we do want to avoid cascading error messages triggered from an earlier one (as noted originally, the issue was that while we got an error message at the site of error, the return of a `bool` causes misleading error messages to be produced later one).

Here's a simple example of where this might come up:
```
class Foo(m.Circuit):
    io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit))
    x = io.O ^ 1
    y = x ^ x
```

In this case, we should get the "can't use `^` with input" error for the first operation.  However, perhaps `x ^ x` is actually what the user wanted and they just mistakenly used `io.O` instead of `io.I`.  One idea would be to return an output value from the invalid code (e.g. still create the `^` instance and return the output), which would then any error popping up with `x ^ x`.  Otherwise if we do something like return None, then we'll get a Python exception that would stop our compilation.

rsetaluri:
I see now. I actually think eagerly raising a `TypeError` upon an invalid operator is fine. Imagine if you did `x ^ ""`, it would definitely raise a TypeError eagerly, so doing the same for output values seems reasonable to me. The specific case of an isolated error stopping execution doesn't seem necessarily like a bad pattern -- e.g. that's how normal python execution works.

My bigger concern is the inconsistency of return values and errors of the various operators. From above:

> While we're here, should we think about standardizing the interfaces of these methods? For example some of them return False on a wrong type. Some return NotImplemented. Sometimes you get a TypeError?

Doesn't seem to be critical, but an opportunity to clean stuff up while we're here. Up to you.

rsetaluri (approved):

rsetaluri:
@leonardt this looks good to merge?

leonardt:
Let me review the comments to see if there was anything left outstanding that we wanted to include or if it just got dropped, but I think adding the improved error message is something we should merge in, just want to make sure there's no low hanging improvements that we can make before doing the merge

PR #858, opened 2020-10-02
Add support for coreir_module metadata

leonardt:
Adds support for a user defined class variable coreir_metadata that
contains a dictionary to be passed through to coreir.  For each
dictionary item the key must be a string and the value must be
convertible to json (with `json.dumps`).

Depends on https://github.com/rdaly525/coreir/pull/965

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -235,6 +235,7 @@ class CircuitKind(type):
         dct.setdefault('primitive', False)
         dct.setdefault('coreir_lib', 'global')
         dct.setdefault("inline_verilog_strs", [])
+        dct.setdefault("coreir_metadata", {})
         dct["inline_verilog_generated"] = False
         dct["bind_modules"] = {}
         dct["compiled_bind_modules"] = {}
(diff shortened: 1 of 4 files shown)


magma/circuit.py
```diff
@@ -235,6 +235,7 @@ def __new__(metacls, name, bases, dct):
         dct.setdefault('primitive', False)
         dct.setdefault('coreir_lib', 'global')
         dct.setdefault("inline_verilog_strs", [])
+        dct.setdefault("coreir_metadata", {})
```
rsetaluri, line 238: it's a bit cheeky, but we could consider adding a set of predefined dicts in `__prepare__` so that the code could look like:

```python
class Foo(m.Circuit):
    coreir_metadata["inline_single_instances"] = False
    ...
```

it just kinda looks a little nicer but not sure it's worth it :)

rsetaluri (approved):

magma/circuit.py
leonardt: Clever, I wonder if this has any problematic interactions with subclassing?
I think in subclassing you would need to do something like
```
class Bar(Foo):
    coreir_metadata = Foo.coreir_metadata
```
to inherit class variables so doesn't seem like it would be an issue since you need to explicitly do it in Python (and this would override the prepare introduced dictionary)

leonardt: I guess from that perspective it introduces an implicit class variable which is not a typical pattern in Python so the downside is losing Pythonic style.

leonardt: Merging this as a first pass since if we decide to add the __prepare__ logic, it should be backwards compatible (this syntax will just overwrite the prepare dict which is totally fine).

PR #859, opened 2020-10-03
conversion operators should extend to n

leonardt:
This adds logic to the conversion operators to extend to the user
specified `n`.  Before, it would simply just convert to the type without
extension.

### DIFF
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -172,10 +172,7 @@ def convertbits(value, n, totype, checkbit):
 
     value = totype[len(Ts), T](ts)
     if n is not None and len(value) < n:
-        if isinstance(value, SInt):
-            value = sext_to(value, n)
-        else:
-            value = zext_to(value, n)
+        value = value.ext_to(n)
     return value
 
 
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -170,7 +170,13 @@ def convertbits(value, n, totype, checkbit):
             Direction.InOut: InOut
         }[T.direction](Bit)
 
-    return totype[len(Ts), T](ts)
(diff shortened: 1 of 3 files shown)


magma/conversions.py
```diff
@@ -170,7 +170,13 @@ def convertbits(value, n, totype, checkbit):
             Direction.InOut: InOut
         }[T.direction](Bit)
 
-    return totype[len(Ts), T](ts)
+    value = totype[len(Ts), T](ts)
+    if n is not None and len(value) < n:
+        if isinstance(value, SInt):
```
rsetaluri, line 175: do these types have `ext_to`? then we could just use that and it'll dispatch to sext/zext? not sure.

tests/test_conversion.py
```diff
@@ -32,3 +32,15 @@ def test_ext():
 def test_concat_type_error():
     with pytest.raises(TypeError):
         m.concat(object(), object())
+
+
+@pytest.mark.parametrize('op', [m.uint, m.sint])
+def test_convert_extend(op):
```
rsetaluri, line 38: did this not come up anywhere else? i guess every usage had `len(T) == n`?

rsetaluri (approved):

magma/conversions.py
leonardt: We could add that, but I originally only added the function style interface

tests/test_conversion.py
leonardt: Yea, I ran into this bug when doing some code of the riscv_mini, I had to use ext first, then convert.  I think a lot of code might be using `m.UInt[N](val)` rather than using the conversion operator which I think works.

PR #861, opened 2020-10-08
Fix clock undriven bug

leonardt:
See comment about unused and why it doesn't need the same change.  I
think we need to add a wireable API to handle the clock wiring
restriction cleanly, or reorganize the type hierarchy if we want to
allow the current pattern.

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -15,10 +15,27 @@ class _ClockType(Digital):
         return True
 
     def unused(self):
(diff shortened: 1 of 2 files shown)


rsetaluri (commented):
@leonardt is this indirectly addressed by #892?

leonardt:
Possibly, I will add it a test to that branch to check

leonardt:
Looks like it's been resolved

PR #862, opened 2020-10-10
Add read enable logic with coreir.sync_read_mem

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/968, see discussion
for context on why this needs a new primitive for a quick solution.

### DIFF
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -33,9 +33,10 @@ class CoreIRMemory(Generator2):
         if sync_read:
             self.io += IO(ren=In(Enable))
             self.coreir_name = "sync_read_mem"
+            self.coreir_lib = "memory"
         else:
             self.coreir_name = "mem"
-        self.coreir_lib = "coreir" if not sync_read else "memory"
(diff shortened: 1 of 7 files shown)


magma/primitives/memory.py
```diff
@@ -119,16 +122,21 @@ def _get_magma_value_(self):
 class Memory(Generator2):
     def __init__(self, height, T: Kind,
                  read_latency: int = 0, read_only: bool = False,
-                 init: Optional[tuple] = None):
+                 init: Optional[tuple] = None, has_read_enable: bool = False):
         if read_latency < 0:
             raise ValueError("read_latency cannot be negative")
+        if has_read_enable and has_read_enable == 0:
```
rsetaluri, line 128: based on the value error should this be `read_latency == 0`? seems like either this if condition is wrong, or the error message is wrong.

rsetaluri (approved):
minor comment. lgtm

magma/primitives/memory.py
leonardt: good catch

PR #863, opened 2020-10-12
Fix deprecation warning from bfloat code

leonardt:
Also add deprecation decorator to DeclareCoreIRCircuit.  @deprecation
uses stacklevel=2 which means that the DeclareCircuit deprecation will
just show up in DeclareCoreIRCircuit.  Adding the decorator to
DeclareCoreIRCircuit will also show the warning where it occurs for the
user of DeclareCoreIRCircuit (in this case, was in the bfloat code).

### DIFF
--- a/magma/bfloat.py
+++ b/magma/bfloat.py
@@ -1,15 +1,8 @@
 from functools import lru_cache
 from magma.bit import Bit
 from magma.bits import UInt
-from magma.circuit import Circuit, coreir_port_mapping
-from magma.interface import IO
+from magma.circuit import declare_coreir_circuit
 from magma.t import In, Out
-
-
-class CoreIRFloatOp(Circuit):
-    renamed_ports = coreir_port_mapping
-    coreir_genargs = {"exp_bits": 8, "frac_bits": 7}
-    coreir_lib = "float"
 
 
 class BFloat(UInt):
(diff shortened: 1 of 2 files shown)


magma/bfloat.py
```diff
@@ -1,54 +1,74 @@
...
             raise NotImplementedError("Only BFloat16 supported")
-        coreir_genargs = {"exp_bits": 8, "frac_bits": 7}
-        return DeclareCoreirCircuit(f"magma_BFloat_{N}_{op}",
-                                    "I", In(cls),
-                                    "O", Out(cls),
-                                    coreir_name=op,
-                                    coreir_genargs=coreir_genargs,
-                                    coreir_lib="float")
+
+        class _BFloatOp(CoreIRFloatOp):
```
rsetaluri, line 23: Could also pull out the generation of these modules into a function local to this file. though that almost replicates the original `DeclareCoreirCircuit` ... so perhaps that's a useful function to have. I think I just didn't want the CamelCase naming and list-style IO (as opposed to dict) to live on.

Maybe we can deprecate `DeclareCoreirCircuit` and define a new function

```python
def declare_coreir_circuit(name: str, ports: dict, coreir_name, coreir_genargs, coreir_lib):
    ...
```

But it's not critical so can do it later.

rsetaluri (approved):

rsetaluri (approved):

PR #865, opened 2020-10-15
Add set_bit function

leonardt:


rsetaluri (commented):
There's actually many ways to implement dynamically setting a bit. They might all be "isomorphic" in some way (of course in the obvious way that they compute the same function, but perhaps also in the total number of 1-bit AND's and OR's required). But I imagine they have different implications on area and wire length. In fact my guess is that if this is a directly supported operator in verilog (i.e. `assign thiswire[thatwire] = otherwire`), backend tools do well at creating the the best structures. It's also possible that this operation is infrequent enough and small enough that it doesn't matter at all :). 

Here are the ones I came up with. Method 1 is the one from this PR.

```python
import hwtypes


def method1(number, index, newbit):
    return type(number)([
        (index == i).ite(newbit, bit) for i, bit in enumerate(number)
    ])


def method2(number, index, newbit):
    T = type(number)
    index = index.zext(len(T) - len(index))
    newbit = T(newbit)
    return (number & ~(T(1) << index)) | (newbit << index);


def method3(number, index, newbit):
    T = type(number)
    index = index.zext(len(T) - len(index))
    newbit = T([newbit for _ in range(len(T))])
    return number ^ ((newbit ^ number) & (T(1) << index))


def sample():
    return (hwtypes.BitVector.random(32),
            hwtypes.BitVector.random(5),
            hwtypes.Bit.random())


number, index, newbit = sample()
# Gold computation.
gold_bits = number.bits()
gold_bits[int(index)] = int(newbit)
gold = hwtypes.BitVector[32](gold_bits)

assert method1(number, index, newbit) == gold
assert method2(number, index, newbit) == gold
assert method3(number, index, newbit) == gold
```

I wrote them in hwtypes to verify, but I think the functions actually are abstract enough to work like peak by passing in magma values.

The most ideal option would be to have a CoreIR primitive that lowers in to the bit-set verilog syntax (assuming it exists), but that's a bigger question. The next best option would be to run an experiment to figure out which code synthesizes the best, but that might be a complex question that is very context-dependent. Without doing either, my *guess* is that Method 3 is best. Method 1 has `N` equality operators, and Method 2 has 2 shifters.

I might run the 3 designs through Yosys just for fun to see what happens :)

rsetaluri:
After some *very* rough experiments, I'm unconvinced we can get a good handle on the QoR implications of the various designs. I'll walk back my claims a bit. Perhaps the best thing to do is to get in the functionally correct thing, get people to use it, and keep a note in our minds that this could be a source of high area/fan out. And if they get to the QoR part of this then we can revisit the designs.

Separately, maybe another case for linking ;) (@rdaly525)

magma/primitives/set_bit.py
```diff
@@ -0,0 +1,13 @@
+from magma.bit import Bit
+from magma.conversions import bits
+from magma.bits import Bits, UInt
+from magma.primitives import mux
```
rsetaluri, line 4:
```suggestion
from magma.bit import Bit
from magma.bits import Bits, UInt
from magma.conversions import bits
from magma.primitives import mux
```

rsetaluri (approved):

leonardt:
Just ran into another use case where I'd like to do this pattern (set an index dynamically) for a more general type (in this case an array of bits), so this general pattern should work for more complex types (although we can still specialize the bit implementation later on if we want)

leonardt:
Perhaps I'll change the name to `set_index`

rsetaluri:
Good point about the logic for more generic types. Also like `set_index` as an analog to `set_slice`. On that note, should `set_slice` also support `Array[T]` instead of just `Bits[n]`?

leonardt:
Hmm, yes that would probably be generally useful (although I haven't run into that yet in an example)

rsetaluri (approved):
looks good to me

PR #866, opened 2020-10-15
Add `when` parameter to Memory.write method

leonardt:
Before, the `WE` port was being automatically wired by the clock wiring
logic (since Enable is considered a clock type).  Since the `WE` port is
an explicit part of the memory interface, we should expose the enable
parameter to the `write` method.

### DIFF
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -192,7 +192,8 @@ class Memory(Generator2):
         self.read = read
 
         if not read_only:
-            def write(self, data, addr):
+            def write(self, data, addr, when):
                 self.WDATA @= data
                 self.WADDR @= addr
+                self.WE @= when
             self.write = write
--- a/tests/test_primitives/test_memory.py
+++ b/tests/test_primitives/test_memory.py
@@ -157,7 +157,7 @@ def test_memory_read_latency(en):
         else:
             rdata = Mem4x5.read(io.raddr)
         io.rdata @= Mem4x5.RDATA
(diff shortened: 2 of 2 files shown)


rsetaluri (approved):
Looks good. Should it be optional, in the case that the auto wiring makes sense to the user?

leonardt:
Hmm, not sure about that.  We can make it optional, but I imagine most users want to control when enable occurs rather than letting is pass through, so the common case would not be automatic.  Another option could be to support passing `None` as a parameter which then skips the `WE` wiring.  In my experience, auto clock wiring is really useful for clocks and reset, since typically these signals are just passed through the hierarchy.  Enables are typically wired explicitly (e.g. when to update a register).  The uncommon case is writing something like a register (or in this case memory) wrapper which just passes the enable through, so I'd almost vote for requiring the enable to be explicitly chosen to be implicitly wired (otherwise it's probably more often than not a bug?).  This makes me consider whether we should include the Enable type in the auto clock wiring logic by default, I'm not sure if there's many people relying on this behavior (in this case, it was implicitly fixing a bug since I didn't actually want it to be done implicitly, so if people are using it, it may be by accident).

rsetaluri:
That makes sense. On that note, I would agree with the idea that enables should *probably* not be auto-wired. But that's for later.

leonardt:
I"ll leave it explicit and required for now then, since in the uncommon wrapper case, having to explicitly pass the enable port isn't too bad.

PR #867, opened 2020-10-15
Return coreir module from coreir compiler

leonardt:
Useful if we want to compile a circuit then interact with it using another library via the coreir API (e.g. Pono), ideally we'd also skip the save to file logic, but this adds the baseline functionality.  For now, other compile options will return None

### DIFF
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -60,6 +60,7 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
         TerminateUnusedPass(main).run()
 
     result = compiler.compile()
+    result = {} if result is None else result
 
     if hasattr(main, "fpga"):
         main.fpga.constraints(basename)
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -59,7 +59,9 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
     if opts.get("terminate_unused", False):
         TerminateUnusedPass(main).run()
 
-    compiler.compile()
+    result = compiler.compile()
 
     if hasattr(main, "fpga"):
         main.fpga.constraints(basename)
+
+    return result
--- a/magma/backend/coreir_compiler.py
+++ b/magma/backend/coreir_compiler.py
@@ -63,6 +63,7 @@ class CoreIRCompiler(Compiler):
         self.backend = coreir_frontend.GetCoreIRBackend()
 
     def compile(self):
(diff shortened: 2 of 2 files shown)


magma/compile.py
```diff
@@ -59,7 +59,9 @@ def compile(basename, main, output="coreir-verilog", **kwargs):
     if opts.get("terminate_unused", False):
         TerminateUnusedPass(main).run()
 
-    compiler.compile()
+    result = compiler.compile()
 
     if hasattr(main, "fpga"):
         main.fpga.constraints(basename)
+
+    return result
```
rsetaluri, line 67: Not sure I like the top level compile function returning an arbitrary object that we're right now interpreting as either `None` or a coreir module. Might be pedantic/picky since we pretty much only ever call compile with "coreir" anyway. If we're designing this generally I think a good option would be to return the `compiler` object. Which can store whatever stuff. In this case the coreir compiler can cache the top compiled module. Or keep a log of them.

leonardt: Hmm, but wouldn't returning the result of compilation be a reasonable expectation for a function called `compile`? In most cases, the result is a file (so we could return a file name as a "reference").  Perhaps it makes more sense if we change this to only be behavior when we compile to coreir without producing a file, which would then require this as the only means to retrieve a reference to the compilation result.  (perhaps we can make the output filename an optional argument for this, but raise an error unless it's the coreir backend).

returning the `Compiler` object would be fine as well, but seems a bit odd for a compile function to return a Compiler object to me (especially since it dispatches to Compiler.compile, so I would more expect it to return whatever the reuslt of `Compiler.compile` is)

rsetaluri: Returning the result of `compile` is definitely a reasonable expectation for a function called `compile` -- but my issue is that depending on the arguments to compile the returned object maybe of completely different type, flavor, and meaning (today it is just `None` or the module, but in general this feels weird). I also agree returning the `Compiler` object is weird.

I think I would feel better if we *always* returned an attribute dict, which depending on the arguments to `compile` will have known keys. It can definitely be empty. In the CoreIR case it should say return `{"coreir_module": ...}`.

Let me propose the change I'm thinking.

leonardt: This seems reasonable, and I could see adding other things like "file_name" or other useful information like a file list

rsetaluri (approved):

PR #873, opened 2020-10-22
Set inline=True default for coreir-verilog

leonardt:


leonardt:
Sent a note to the AHA mailing list about this, with the step to turn it off.  We can give them a few days to get ready, also some tests may break (e.g. gold checks) if they assume inline=False, so we'll need to update those

rsetaluri:
Ok was going to comment regarding that.

My main concern is the number and distribution of `m.compile()` call sites. Even owners of various pieces may not even know where all the call sites are (which is a separate issue).

One fairly safe thing would be to have an environment variable which changes the default inline behavior. Since magma @ industry is used in a *very* controlled environment, it seems safer to set an env var in magma @ industry flow.

So it could look like:

config.py:
```MAGMA_INLINE_DEFAULT = EnvVar("MAGMA_INLINE_DEFAULT", default=False)```

coreir_compiler.py:
```opts.get("inline", default=config.config["MAGMA_INLINE_DEFAULT"])```

Then in magma @ industry flow setup shell script:
```export MAGMA_INLINE_DEFAULT=1```

It adds an extra level of indirection and is kind of confusing (there are 2 levels of defaults), but seems like the safest option. After tapeout we could change to

config.py:
```MAGMA_INLINE_DEFAULT = EnvVar("MAGMA_INLINE_DEFAULT", default=True)```

rsetaluri:
See https://github.com/StanfordAHA/garnet/pull/724

rsetaluri:
@leonardt since StanfordAHA/garnet#724 is merged we can merge

leonardt:
Sounds good, still have to fixup these tests

rsetaluri:
@leonardt do we still want to do this? if so, can take a pass on tests

leonardt:
Yes I think we should, and this seems like a good time to do it.  New year, new default compiler option xD

rsetaluri:
> Yes I think we should, and this seems like a good time to do it. New year, new default compiler option xD

Sounds good, I'll take this on.

leonardt:
Deprecated by MLIR compiler

PR #877, opened 2020-10-25
Support init value for non-Bits Memory

splhack:
To init non-`Bits[N]` Memory, use `int(uint(as_bits(x)))` to convert the value.

By the way, a bit confused about the order of coreir memory `.init`

    .init({16'd127,16'd65535,16'd32640,16'd32768}),

it's reverse order?

    test_data = [
        {"X": 0, "Y": -128},   # 32768
        {"X": -128, "Y": 127}, # 32640
        {"X": -1, "Y": -1},    # 65535
        {"X": 127, "Y": 0},    # 127
    ]

### DIFF
--- a/magma/primitives/memory.py
+++ b/magma/primitives/memory.py
@@ -5,13 +5,13 @@ from magma.bits import Bits
 from magma.bitutils import clog2
 from magma.clock import Clock, Enable
 from magma.clock_io import ClockIO
-from magma.conversions import from_bits, as_bits, enable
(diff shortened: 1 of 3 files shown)


leonardt:
Verilog uses MSB -> LSB ordering in the concat statement (so `{MSB, ..., LSB}`), so for index `0` of the Python list (first element) to map to index `0` of the verilog value, it will be the last element.

leonardt (approved):
LGTM, thanks!

PR #878, opened 2020-10-26
Remove old, untested/unused coroutine testing code

leonardt:
This old code was originally used to generate test vectors using
coroutines.  It really should live in fault, but we've also migrated to
using other patterns (e.g. sequence tester works with a generic
iterable, like a coroutine/generator) so I'm not sure we need this
anymore.  I've tagged this removal with a decently searchable name so if
we need to revive the code we should be able to find it in the git
history.


rsetaluri (approved):
thanks for cleanup

PR #879, opened 2020-10-26
Remove old, untested/unused compile_and_regress

leonardt:
Similar to https://github.com/phanrahan/magma/pull/878, this code was
originally used to manage compilation tests for the various backends.
This probably belongs in mantle now, but again, should be easy to find
in the history if we need it.


rsetaluri (approved):
thanks for cleanup

PR #883, opened 2020-11-03
Add register utility fn

rsetaluri:
Simple function to register a value, similar to mantle (https://github.com/phanrahan/mantle/blob/7fee3a401f1ba9197c3e5ac125445c35bb44452e/mantle/common/register.py#L75-L90). On that note should probably converge mantle reg and magma primitive reg?

### DIFF
--- a/magma/primitives/register.py
+++ b/magma/primitives/register.py
@@ -10,7 +10,7 @@ from magma.circuit import coreir_port_mapping
 from magma.conversions import as_bits, from_bits, bit
 from magma.interface import IO
 from magma.generator import Generator2
-from magma.t import Type, Kind, In, Out
+from magma.t import Type, Kind, In, Out, Direction
 from magma.tuple import Tuple
 from magma.clock import (AbstractReset,
                          AsyncReset, AsyncResetN, Clock, get_reset_args)
@@ -209,7 +209,7 @@ class Register(Generator2):
 
 
 def register(value, **kwargs):
-    T = type(value)
+    T = type(value).qualify(Direction.Undirected)
     inst_kwargs = {}
     if "name" in kwargs:
         inst_kwargs["name"] = kwargs["name"]
--- a/magma/primitives/register.py
+++ b/magma/primitives/register.py
@@ -206,3 +206,9 @@ class Register(Generator2):
         elif (has_reset or has_resetn):
             I = Mux(2, T)()(I, init, reset_select)
         reg.I @= as_bits(I)
+
+
+def register(value, **kwargs):
+    T = type(value)
+    ckt = Register(T, **kwargs)
+    return ckt()(value)
--- a/magma/primitives/register.py
+++ b/magma/primitives/register.py
@@ -211,8 +211,10 @@ class Register(Generator2):
 def register(value, **kwargs):
     T = type(value).qualify(Direction.Undirected)
     inst_kwargs = {}
-    if "name" in kwargs:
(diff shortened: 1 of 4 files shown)


leonardt:
Yes, I think we should deprecate mantle reg and add any features to the primitive reg that are missing, unless we think it makes more sense to keep some aspects of the mantle reg in mantle as a wrapper around the primitive, but I can't think of any reason to do that

magma/primitives/register.py
```diff
@@ -206,3 +206,9 @@ def __init__(self, T: Kind = None,
         elif (has_reset or has_resetn):
             I = Mux(2, T)()(I, init, reset_select)
         reg.I @= as_bits(I)
+
+
+def register(value, **kwargs):
+    T = type(value)
+    ckt = Register(T, **kwargs)
+    return ckt()(value)
```
leonardt, line 214: Let's add a simple test.

Also, we may want to provide the ability to pass through the `name` parameter to the instance instead of the generator (to manually override the instance name)

rsetaluri: Done. Lmk if the interface is good

magma/primitives/register.py
```diff
@@ -206,3 +206,13 @@ def __init__(self, T: Kind = None,
         elif (has_reset or has_resetn):
             I = Mux(2, T)()(I, init, reset_select)
         reg.I @= as_bits(I)
+
+
+def register(value, **kwargs):
+    T = type(value).qualify(Direction.Undirected)
```
rsetaluri, line 212: @leonardt one question is if the un-qualification logic should be in the base Register generator or not. not a big deal either way

leonardt: Hmm, I'm not sure.  I think normally a user would say the type explicitly, e.g. `Register(Array[N, Bits[M])`, so it's undirected.  In this case, when you call `register` it's often directed (unless it's an intermediate value) so it makes sense to call it here.  I think the one case I can imagine is if a user says something like `Register(type(v))` where `v` is a directed value.  Often though, in this case, they could just call `register(v)`, but maybe for somer reason they want to just create a register of the same type without using the value as the input, so maybe it's useful to have in the base generator even if it's an obscure case.  We could use a decorator that handles this conversion since it's probably a similar issue in other generators.

leonardt (approved):
LGTM

tests/test_primitives/test_register.py
```diff
@@ -295,3 +295,53 @@ class test_reg_init_uniq(m.Circuit):
...
+        name = "test_basic_reg"
+        io = m.IO(I=m.In(m.Bits[8]), O=m.Out(m.Bits[8]))
+        io += m.ClockIO(has_reset=True)
+        io.O @= register(io.I, init=ht.BitVector[8](0xDE), reset_type=m.Reset)
+
+    m.compile("build/test_basic_reg_function", test_basic_reg)
+
+    assert check_files_equal(__file__,
+                             f"build/test_basic_reg_function.v",
+                             f"gold/test_basic_reg.v")
```
leonardt, line 312: Need to check in this gold file

leonardt: Oh wait, nvm, this is reusing the other gold file.

magma/primitives/register.py
rsetaluri: Makes sense -- agree that its more specific to `register` than `Register` because the common case is an explicit type passed in rather than `type(value)`

PR #885, opened 2020-11-07
Add API to reset generator cache

leonardt:
Needed for testing where coreir context is reset (it seems that
generators are keeping around circuits in the cache that break when
trying to compile, clearing the cache alongside the coreir context fixes
this)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,3 +9,4 @@ def magma_test():
     magma.config.set_compile_dir('callee_file_dir')
     clear_cachedFunctions()
     magma.frontend.coreir_.ResetCoreIR()
+    magma.generator.reset_generator_cache()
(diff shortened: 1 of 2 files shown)


leonardt:
This should be equivalent to this logic: https://github.com/phanrahan/magma/blob/master/magma/__init__.py#L9-L11 for the old `@cache_definition` style

rsetaluri:
How is this related to the coreir context clearing? I see clearing the metaclass cache being useful, but not understanding how it relates to the backend stuff (coreir context).

leonardt:
The issue is the current backend flow modifies the circuit to attach some state corresponding to a backend (see https://github.com/phanrahan/magma/blob/master/magma/backend/coreir_transformer.py#L145).  When a context is reset, this pointer hangs around and will get used (to avoid recompiling a circuit twice), but in this case it should be invalidated.  The simplest solution is just invalidate the generator cache, so all circuits are regenerated fresh.  We should probably investigate a way to cleanup this stateful backend logic so it doesn't attach to the circuit (instead it should probably maintain a mapping in the backend so it's contained in the same place)

rsetaluri:
Got it, makes sense. Made an issue (#886) to figure this out later.

Seems like a reasonable workaround for now.

Issue #870, opened 2020-10-16
[RFC] Logic on enable signals


leonardt:
Right now we consider Enable signals to be similar to clocks, where logic is not allowed (this issue is related https://github.com/phanrahan/magma/issues/869).  In order to have enable logic, we need to cast to/from the enable type to the bit type.  

In my experience, we often do require logic for enable signals (e.g. only write to a register when a counter is at a specific value, or only write to memory when a certain condition is met).  Since requiring logic for enables happens often, and we simply cast to and from Bit to do this, I wonder what having an explicit Enable type that disables logic affords us? I often encounter myself running into an Enable type error related to performing logic, inserting the casts, then moving on (so it's not like it's really catching any real errors for me).  I'm open to keeping this type strict about logic, but I'm just wondering what we gain by forcing the user to manage this casting for performing logic?

leonardt:
Closed by https://github.com/phanrahan/magma/pull/889

PR #889, opened 2020-12-02
Update enable type for PEak compatability

leonardt:
* Enable subclasses Bit to make logic on the enable signal easier (see
  https://github.com/phanrahan/magma/issues/870 for some discussion)
* Add a parameter to the register generator for the name of the enable
  signal (PEak uses `en` instead of the default magma `CE`).  A more
  general solution to this problem might be to add the ability to alias
  port names (rather than having to parametrize generators with the port
  names).  Another option is to provide some library routine that takes
  as circuit and mapping from old to new port names and generates a
  wrapper.  However since PEak just needs this for the register, this is
  a simple stopgap solution.
* A side effect of making Enable a Bit instead of a Clock is now it is
  explicitly required when using the `__call__` syntax.  We think this
  is good because in the common case you'll want to provide an enable
  signal when calling a register to update it.  The only time that you
  wouldn't want to do this is when you want the enable to be implicitly
  wired by the automatic clock wiring logic, which I would say is the
  uncommon case.  Similar to the motivation described in
  https://github.com/phanrahan/magma/issues/870, I think we should just
  requires the user explicitly pass through the enable for the uncommon
  case rather than rely on the automatic wiring.  This has the benefit
  of catching bugs where the user forgets to pass an enable signal when
  calling a register to update it.

For now, we leave some of the clock type related logic for Enables
around for backwards compatability, but we should plan to deprecate it
moving forward (e.g. automatic Enable wiring logic).

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -81,7 +81,7 @@ AsyncResetNOut = AsyncResetN[Direction.Out]
 
 # Preset
 # Clear
-class Enable(_ClockType, metaclass=DigitalMeta):
+class Enable(Bit):
     pass
 
 
--- a/magma/primitives/register.py
+++ b/magma/primitives/register.py
@@ -5,6 +5,7 @@ import hwtypes as ht
 
 from magma.array import Array
 from magma.bit import Bit
+from magma.clock import Enable
 from magma.bits import Bits, UInt, SInt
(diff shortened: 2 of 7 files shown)


cdonovick (approved):
LGTM

PR #890, opened 2020-12-04
Support renaming register input/output ports

leonardt:
Needed for PEak since they use an alternative naming convention

Note we need to use ParamDict (rather than a standard dict) so the generator arguments are hashable

### DIFF
--- a/magma/primitives/register.py
+++ b/magma/primitives/register.py
@@ -6,11 +6,12 @@ import hwtypes as ht
 from magma.array import Array
 from magma.bit import Bit
 from magma.clock import Enable
+from magma.common import ParamDict
 from magma.bits import Bits, UInt, SInt
 from magma.circuit import coreir_port_mapping
 from magma.conversions import as_bits, from_bits, bit
 from magma.interface import IO
-from magma.generator import Generator2, ParamDict
+from magma.generator import Generator2
 from magma.t import Type, Kind, In, Out, Direction
 from magma.tuple import Tuple
 from magma.clock import (AbstractReset,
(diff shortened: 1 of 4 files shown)


tests/test_circuit/test_reg_enable_call.py
```diff
@@ -1,11 +1,13 @@
...
+from magma.generator import ParamDict
 
 
 def test_reg_enable_call():
     class test_reg_enable_call(m.Circuit):
         io = m.IO(I=m.In(m.Bits[5]), O=m.Out(m.Bits[5]), nen=m.In(m.Bit))
         io += m.ClockIO()
-        reg = m.Register(m.Bits[5], has_enable=True, enable_name="en")()
+        reg = m.Register(m.Bits[5], has_enable=True,
+                         name_map=ParamDict(CE="en"))()
```
cdonovick, line 10: Does the `name_map` need to be a `ParamDict` or will any mapping work?

cdonovick (approved):

tests/test_circuit/test_reg_enable_call.py
leonardt: Needs to be hashable (ParamDict is just a dict that defines a __hash__ method)

magma/primitives/register.py
```diff
@@ -10,13 +10,14 @@
 from magma.circuit import coreir_port_mapping
 from magma.conversions import as_bits, from_bits, bit
 from magma.interface import IO
-from magma.generator import Generator2
+from magma.generator import Generator2, ParamDict
```
rsetaluri, line 13: can we move `ParamDict` to `magma/common.py`?

rsetaluri (approved):

Issue #571, opened 2020-02-04
Type polymorphism weirdness


rdaly525:
I would suspect the following to error in magma but it is falsely working. lassen:master is doing something equivalent to the following: 
```
def test_adc():
    Data = m.Bits[16]
    UData = m.UInt[16]
    SData = m.SInt[16]
    @m.circuit.sequential(async_reset=True)
    class ADC:
        def __call__(self, s:m.Bit, a:Data, b:Data) -> (Data, m.Bit):
            if s:
                a = SData(a)
                b = SData(b)
            else:
                a = UData(a)
                b = UData(b)
            return a.adc(b,s)
```

It is unclear which type's adc should be used, so this should error.

from the ast_tools SSA pass which is more conservative, this fails complaining about 'a' being different types on each branch (inputs to the phi)

I have this test in the branch magmag:adc

cdonovick:
Untested but my guess is that magma allows `condition.ite(signed, unsigned)`.  Which hwtypes would reject,  while `signed` and `unsigned` share a common base I do not think it appropriate to operate in such a manner.  I don't want: 
```
class Foo: pass
foo = Foo()
bar = BitVector[8](...)
con = Bit(...)
assert type(con.ite(foo, bar)) is object
```
I want it to error (technically it would error as you can't construct `object` from other of the arguments, but my point should be clear).

Related to https://github.com/leonardt/hwtypes/issues/122

leonardt:
Yes this should be an error and it's probably an artifact of the subclass relationship as Caleb pointed out.

leonardt:
Should be fixed by #892, let me know if it isn't

PR #892, opened 2020-12-07
Fix ite type checking bug (#571)

leonardt:
Before, ite was allowing SInt and UInt to be used together, which should
not be allowed.  This fixes it by first updating the ite logic to use
the mux operator (which has improved type checking logic) and updating
the wireable logic to prevent UInt and SInt from being wireable (they
match structurally but we should error when trying to wire them
together).  I think we should eventually update the wireable logic to
allow the types to define their wireability themselves (rather than
housing the logic internal to magma since this won't extend to user
defined types).  However, since UInt/SInt are built-in to magma, this
should be fine for now.


tests/test_primitives/test_ite.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+import magma as m
+
+
+def test_adc_signed_unsigned():
+    Data = m.Bits[16]
+    UData = m.UInt[16]
+    SData = m.SInt[16]
+
+    with pytest.raises(TypeError) as e:
```
rsetaluri, line 10: Can we distill the test(s) for what we want to check? In particular it might be easier to read and maintain if the test just does the minimal thing, i.e. not wrapped in `combinational`. Perhaps one test to check (a) basic sint-uint wiring (see comment above), and one test for (b) mux'ing those.

magma/wireable.py
```diff
@@ -8,6 +9,9 @@ def wireable(T1, T2):
     Returns true if T1 can be wired to T2
     """
     T1, T2 = magma_type(T1), magma_type(T2)
+    if ((issubclass(T1, UInt) and issubclass(T2, SInt)) or
```
rsetaluri, line 12: what's the difference between `wireable(x, y)` and `x.is_wireable(y)`? It seems there's some fragmentation of this logic. Ok if it's the case, but ideally should document when to use one vs the other (or update names appropriately).

In this case, I tried wiring a uint to an sint and it worked so that seems to be inconsistent with the logic here? Is this only for muxing?

cdonovick:
> think we should eventually update the wireable logic to allow the types to define their wireability themselves

I think this a good idea.  Add a `_wire_` method to the magma interfce.

And make the wire function do something basically like:
```Python
def wire(lhs, rhs):
    # ... Logic around the magma protocol ...
    try:
        res = lhs._wire_(rhs)
    except AttributeError:
        res = NotImplemented # or some magma specific singleton
    if res is NotImplemented:
        try:
            res = rhs._wire_(lhs)
        except AttributeError:
            res = NotImplemented # or some magma specific singleton
    if res is NotImplemented:
        raise WiringError()
```

magma/wireable.py
leonardt: They should be consistent, I can change it to call is_wireable on the types (related to Caleb's feedback)

tests/test_primitives/test_ite.py
leonardt: Sure, the sint-uint wiring is the underlying problem, so we need to fix that.  then there's the fact that ite (used by combinational/sequential) isn't using the mux operator which defines the type checking rules for inference (so we can change it from using combinational to just calling ite directly)

magma/primitives/mux.py
```diff
@@ -108,7 +107,7 @@ def _infer_mux_type(args):
             if issubclass(T, next_T):
                 # upcast
                 T = next_T
-            elif not wireable(next_T, T):
```
rsetaluri, line 111: can we get rid of `magma/wireable.py` then?

magma/bits.py
```diff
@@ -627,6 +646,16 @@ def __lt__(self, other):
 class SInt(Int):
     hwtypes_T = ht.SIntVector
 
+    @debug_wire
+    def wire(self, other, debug_info=None):
+        if isinstance(other, UInt):
```
rsetaluri, line 651: it seems redundant to have logic `is_wireable` and then do this check again in `wire()`. Seems like `wire()` for any type should by default call `is_wireable()` and the error message of `"Cannot wire..."` is generic.

TL;DR: should we just call `is_wireable()` by default before any logic in `wire()`?

rsetaluri (commented):
Thanks for addressing comments. Two last questions in comments.

magma/primitives/mux.py
leonardt: Already removed by https://github.com/phanrahan/magma/pull/892/commits/69a2783b0110e230261c2d6d0af4005b9f9d762a

magma/bits.py
leonardt: Ok, moved the wire error checking logic to wire() and updated to use is_wireable.  Overall I think it's an improvement, it consolidates the logic into a clear API, whereas before the error checking was done in various parts of the wiring pipeline.  As a side effect, it seems some error messages are improved, whereas other are less precise (since we early error before wiring things by checking is_wireable, we don't descend into the wiring pipeline where old errors were being raise).  I'll work on moving the old error checking and message logic up into the is_wireable flow (I think we can augment it to return an object that indicates and error with a message, this way we can report the generic error "Cannot wire X to Y" and then add a more specific error below it (e.g. array sizes do not match).

magma/bits.py
```diff
@@ -23,6 +23,8 @@
 from magma.language_utils import primitive_to_python
 from magma.logging import root_logger
 from magma.generator import Generator2
+from magma.wire_container import WiringLog
+from magma.debug import debug_wire
```
rsetaluri, line 27: still need these 2? don't see them used

magma/wire.py
```diff
@@ -52,5 +52,15 @@ def wire(o, i, debug_info=None):
...
 
+    i_T, o_T = type(i), type(o)
+    if not i_T.is_wireable(o_T):
+        _logger.error(
+            WiringLog(f"Cannot wire {{}} ({o_T}) to {{}} ({i_T})",
+                      o, i),
+            debug_info=debug_info
+        )
+        return
+
```
rsetaluri, line 63: nit: extra newline

rsetaluri (approved):
Looks good modulo final comments.

magma/bits.py
leonardt: debug_wire is being used on line 159, don't need WiringLog anymore though

PR #893, opened 2021-01-12
Fix CoreIR name error for inline verilog temporary

leonardt:
Before, inserting a temporary wire for the inline verilog logic could
trigger a CoreIR name error by using a directed type for the wire name
(e.g. `Out(Clock)`) since the Wire genertor logic generates a name
`"Wire{T}"`.  It doesn't really make sense to generate a Wire of type
`Out(T)`, rather it should be some undirected type `T`.  This updates
the Wire generator raise an error if you try to construct it with a
directed type, and updates the inline verilog code to use the undirected
type for the intermediate wire.

### DIFF
--- a/magma/t.py
+++ b/magma/t.py
@@ -164,6 +164,10 @@ class Kind(type):
     def undirected_t(cls):
         return cls.qualify(Direction.Undirected)
 
+    @property
+    def is_directed(cls):
+        return cls is not cls.qualify(Direction.Undirected)
+
 
 def In(T):
     return T.qualify(Direction.In)
--- a/tests/test_verilog/test_inline.py
+++ b/tests/test_verilog/test_inline.py
@@ -1,7 +1,6 @@
 import magma as m
 import magma.testing
 import pytest
-import os
 
 
 def test_inline_verilog():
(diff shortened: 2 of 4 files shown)


magma/t.py
```diff
@@ -164,6 +164,10 @@ def flip(cls):
     def undirected_t(cls):
         return cls.qualify(Direction.Undirected)
 
+    @property
```
rsetaluri, line 167: think we can just do `return cls.direction is not None`?

tests/test_verilog/test_inline.py
```diff
@@ -1,6 +1,7 @@
 import magma as m
 import magma.testing
 import pytest
+import os
```
rsetaluri, line 4: spurious?

rsetaluri (approved):
LGTM, 2 comments

magma/t.py
leonardt: Not quite, I think only leaf (Digital) types define the `direction` attribute.  The default implementation of flip is misleading, since arrays/tuples define their own implementations of flip (so it's not really inherited and probably should be removed).

For example, a tuple doesn't necessarily have a direction (it could be mixed based on the children), so I think the most general thing to check is whether the undirected version of the type is equal to the current object.  Alternatively we could have each class implement it specifically (probably more efficient, since you can check if any of the children are directed for tuple and short circuit if you find one), but this avoids having to define an implementation for each type (by using the qualify API).

PR #895, opened 2021-01-19
Guard update_register pass for user defined init

### LINKED ISSUE #894: Sequential2 breaks if reset type is specified but class does not have an init.
```Python3
import magma as m

Data = m.UInt[8]

@m.sequential2(reset_type=m.AsyncReset)
class Inc:
    def __call__(self, i: Data) -> Data:
        return i + 1
```
`TypeError: cannot create weak reference to 'wrapper_descriptor' object`

I believe what is happening is that magma is trying to `get_ast` on `object.__init__` which is not supported.

leonardt:
Fixes #894

Only run the update_register pass (for the automatic reset logic) when
the user defines an __init__ method.

Note: all objects will inherit the default object.__init__, so we check
whether the __init__ defined is different.  This hasn't been tested with
inherited __init__ methods from user defined classes.

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -409,9 +409,11 @@ def sequential2(pre_passes=[], post_passes=[],
         if reset_type:
             update_register = _UpdateRegister(reset_type,
                                               has_enable, reset_priority)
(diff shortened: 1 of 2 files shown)


cdonovick (approved):
lgtm

PR #897, opened 2021-01-20
[seq2] add magma reference if not available in env

### LINKED ISSUE #896: Sequential2 breaks with custom env if magma not in scope
```Python3
import magma as m
from ast_tools import SymbolTable

Data = m.UInt[8]
env = SymbolTable({}, {'Data': Data})

@m.sequential2(env=env, reset_type=m.AsyncReset)
class Inc:
    def __init__(self): pass
    def __call__(self, i: Data) -> Data:
        return i + 1
 ```
 `UnboundLocalError: local variable 'magma_id' referenced before assignment`
 https://github.com/phanrahan/magma/blob/master/magma/syntax/sequential2.py#L286
 
 The prefered design pattern in peak is have to users explicitly pass `locals()` and `globals()` to avoid the error prone process of trying to capture them implicitly.  This means magma will generally not be in the env of peak programs.

### PR DESCRIPTION
**Pull request by leonardt**
Before, we searched for a reference to magma in the sequential2
environment (we need this to add references to types when generating the
reset code).  If we didn't find one, we raised an error.  This changes
the logic to insert a reference to magma with a free name when we can't
find an existing one.

Fixes #896

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -5,6 +5,7 @@ from typing import Optional, MutableMapping
 from ast_tools.stack import SymbolTable
 from ast_tools.passes import (PASS_ARGS_T, Pass, apply_ast_passes, ssa,
                               if_to_phi, bool_to_bit)
+from ast_tools.common import gen_free_name
 
 from ..circuit import Circuit, IO
 from ..clock import AbstractReset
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by cdonovick — COMMENTED** [review]
I don't think this will work `gen_free_name` expects a tree and an env.
https://github.com/leonardt/ast_tools/blob/master/ast_tools/common.py#L161

File: magma/syntax/sequential2.py [hunk 1]
```diff
@@ -333,11 +327,25 @@ def __init__(self, reset_type, has_enable, reset_priority):
...
         self.reset_priority = reset_priority
 
+    def _get_or_add_magma_module(self, env: SymbolTable):
+        magma_id = None
+        for k, v in env.items():
+            if v is magma_module:
+                magma_id = k
+                break
+        if magma_id is None:
+            magma_id = gen_free_name(env)
```
cdonovick, line 337: meant to attach my comment here

cdonovick: modulo that lgtm

cdonovick (approved):

magma/syntax/sequential2.py
leonardt: Yea i had a typo in the condition that actually made the code path not run, fixed up the api usage

PR #903, opened 2021-01-28
Cleanup coreir backend prefix

rsetaluri:
Bunch of cleaning up. Branch `cleanup-coreir-backend` has some continued work that might require more thought, so I wanted to get in the beginning of this sooner than later, and keep working on that.


magma/frontend/coreir_.py
```diff
@@ -1,21 +1,28 @@
...
-    return CoreIRContextSingleton().get_instance()
+    return coreir_context()
+
 
 def ResetCoreIR():
-    CoreIRContextSingleton().reset_instance()
+    reset_coreir_context()
     backend = GetCoreIRBackend()
+    # NOTE(rsetaluri): Is this next line necessary? Don't think it is.
     backend.reset()
```
leonardt, line 23: This calls the backend._init function which resetts all the dictionaries/sets which I think is necessary.  For example, the constant cache holds references to modules for constants.  If we reset the context, we need to clear the backend context.

magma/ref.py
```diff
@@ -207,8 +208,17 @@ def qualifiedname(self, sep="."):
...
         return str(self.view.port.name)
+
+
+def _get_root(ref: Ref) -> typing.Optional[Ref]:
+    parent = ref.parent()
+    if parent is None:
+        return None
+    if parent is ref:
+        return ref
+    return _get_root(parent)
```
leonardt, line 224: Hmm, since this is called with self, is there a reason it's not just a private instance method?

leonardt (approved):

magma/frontend/coreir_.py
rsetaluri: Updated comment.

magma/ref.py
rsetaluri: Good point; refactored

PR #904, opened 2021-01-28
Migrate to ast_tools@cst

leonardt:
Depends on https://github.com/leonardt/ast_tools/pull/62

Drops support for sequential since it doesn't seem worth it to have to migrate that code to the new libcst infrastructure.  We can do a major version bump to indicate the breaking change, I don't think there's any major users besides Peak which is now moved to sequential2


magma/syntax/sequential2.py
```diff
@@ -275,49 +276,48 @@ def _seq_phi(s, t, f):
...
 
-    def visit_Call(self, node: ast.Call):
+    def leave_Call(self, original_node, updated_node):
         # find m.Register
         if (
-            isinstance(node.func, ast.Attribute)
-            and isinstance(node.func.value, ast.Name)
-            and node.func.attr == "Register"
-            and eval(node.func.value.id, {}, self.env) is magma_module
+            isinstance(updated_node.func, cst.Attribute)
```
cdonovick, line 300: Could probably use a matcher https://libcst.readthedocs.io/en/latest/matchers_tutorial.html

magma/syntax/sequential2.py
```diff
@@ -346,26 +346,28 @@ def rewrite(
...
 
-    def visit_Call(self, node: ast.Call):
+    def leave_Call(self, original_node, updated_node):
         # find self.attr.prev()
-        if (isinstance(node.func, ast.Attribute) and
-                isinstance(node.func.value, ast.Attribute) and
-                isinstance(node.func.value.value, ast.Name) and
-                node.func.value.value.id == "self" and
-                node.func.attr == "prev"):
+        if (isinstance(updated_node.func, cst.Attribute) and
```
cdonovick, line 360: also could use a matcher

cdonovick (approved):
LGTM

magma/syntax/sequential2.py
leonardt: very useful pattern

cdonovick:
I am going to set up an integration build on the aha flow

cdonovick:
https://github.com/StanfordAHA/aha/pull/1099

PR #906, opened 2021-02-09
Factor array init into separate functions

leonardt:
Making the array init code a little easier to read


rsetaluri:
@leonardt sorry, reviewing this slipped. Wanted to make a few suggestions re: the init functions becoming free functions rather than members. #907 has those changes.

PR #909, opened 2021-02-13
Add fix for inline veriloq unique

leonardt:
Ensures inline verilog is considered for uniquification by adding the
string to the repr when computing the circuit hash.  Also includes the
str of the kwargs, symbol table, and inline_verilog_string to catch the
(unlikely) case where there's identical inline verilog strings but
different kwargs or prefix.

### DIFF
--- a/magma/uniquification.py
+++ b/magma/uniquification.py
@@ -31,10 +31,10 @@ def _make_hash_struct(definition):
     inline_verilog = ((), (), (), ())
     for s, args, st, prefix in definition._context_._inline_verilog:
         inline_verilog += (
-            (s, ),
-            (", ".join(str(arg) for arg in args), ),
-            (str(st), ),
-            (prefix, )
+            (s,),
+            (", ".join(str(arg) for arg in args),),
+            (str(st),),
+            (prefix,)
         )
     if hasattr(definition, "verilogFile") and definition.verilogFile:
         return _HashStruct(repr_, True, definition.verilogFile, inline_verilog)
--- a/magma/uniquification.py
+++ b/magma/uniquification.py
@@ -23,15 +23,22 @@ class _HashStruct:
     defn_repr: str
     is_verilog: bool
     verilog_str: bool
+    inline_verilog: tuple
 
 
 def _make_hash_struct(definition):
     repr_ = repr(definition)
+    inline_verilog = ((), (), (), ())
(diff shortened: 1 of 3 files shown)


tests/test_verilog/test_inline.py
```diff
@@ -148,3 +148,25 @@ class Foo(m.Circuit):
...
+
+    class Top(m.Circuit):
+        io = m.IO(I=m.In(m.Bit))
+        Bar()(io.I)
+        Foo()(io.I)
+
+    m.compile("build/test_inline_verilog_unique", Top)
+    assert m.testing.check_files_equal(__file__,
+                                       f"build/test_inline_verilog_unique.v",
+                                       f"gold/test_inline_verilog_unique.v")
```
rsetaluri, line 172: need to add this file?

rsetaluri (approved):

magma/uniquification.py
```diff
@@ -27,6 +27,8 @@ class _HashStruct:
 
 def _make_hash_struct(definition):
     repr_ = repr(definition)
+    for s, args, st, prefix in definition._context_._inline_verilog:
```
rsetaluri, line 30: another option is to make these explicit fields of the HashStruct. might be cleaner and more transparent/maintainable

rsetaluri: Looks good.

magma/uniquification.py
```diff
@@ -23,13 +23,22 @@ class _HashStruct:
...
 def _make_hash_struct(definition):
     repr_ = repr(definition)
+    inline_verilog = ((), (), (), ())
+    for s, args, st, prefix in definition._context_._inline_verilog:
+        inline_verilog += (
+            (s, ),
+            (", ".join(str(arg) for arg in args), ),
+            (str(st), ),
+            (prefix, )
+        )
```
rsetaluri, line 38:
```suggestion
        inline_verilog += (
            (s,),
            (", ".join(str(arg) for arg in args),),
            (str(st),),
            (prefix,)
        )
```

see https://www.python.org/dev/peps/pep-0008/#pet-peeves

tests/test_verilog/test_inline.py
rsetaluri: Resolved

rsetaluri (approved):
Good, modulo formatting nit

PR #910, opened 2021-02-17
Add namespace logic to coreir backend module dict

leonardt:
Fixes an error such as:
```
../magma/magma/compile.py:64: in compile
    result = compiler.compile()
../magma/magma/backend/coreir/coreir_compiler.py:71: in compile
    backend.compile(self.main, opts)
../magma/magma/backend/coreir/coreir_backend.py:75: in compile
    transformer.run()
../magma/magma/backend/coreir/coreir_transformer.py:97: in run
    child.run()
../magma/magma/backend/coreir/coreir_transformer.py:98: in run
    self.run_self()
../magma/magma/backend/coreir/coreir_transformer.py:240: in run_self
    self.coreir_module.definition = self.get_coreir_defn()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <magma.backend.coreir.coreir_transformer.DefinitionTransformer object at 0x7f727dba8880>

    def get_coreir_defn(self):
>       coreir_defn = self.coreir_module.new_definition()
E       AttributeError: 'Generator' object has no attribute 'new_definition'

../magma/magma/backend/coreir/coreir_transformer.py:243: AttributeError
```

that is encountered when using a circuit named `add` since it is
clobbered by a module of the same name (the coreir.add primitive).  Does
this by storing module references separated by namespaces (using
separate dictionaries).  Updates the compile API to return the namespace
-> module map dictionary rather than just a single module map.

### DIFF
--- a/magma/backend/coreir/coreir_backend.py
+++ b/magma/backend/coreir/coreir_backend.py
@@ -25,18 +25,20 @@ class CoreIRBackend:
         if context is not singleton:
             _logger.warning("Creating CoreIRBackend with non-singleton CoreIR "
                             "context.")
-        self._modules = module_map().setdefault(context, {})
+        self._namespaces = module_map().setdefault(context, {})
         self._context = context
         self._lib_cache = {}
         self._included_libs = set()
         self._bound_modules = {}
 
     def add_module(self, magma_module, coreir_module):
-        self._modules[magma_module.coreir_name] = coreir_module
+        modules = self._namespaces.setdefault(magma_module.coreir_lib, {})
+        modules[magma_module.coreir_name] = coreir_module
 
     def get_module(self, magma_module):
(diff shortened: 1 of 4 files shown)


rsetaluri (approved):

rsetaluri (changes requested):
Actually, can we keep the name `module_map` while changing the semantics to be the new thing? It still is meaningful and not imprecise, and more concise. I.e. Change the names back, keep the new semantics.

PR #912, opened 2021-02-17
Fix uniquification bug in inline verilog old style

leonardt:


rsetaluri (approved):
Looks good. Ideally, we migrate folks away from old style stuff as soon as possible.

PR #914, opened 2021-02-18
Prototype Compile Guard

leonardt:
This is a minimum working example of the compile guard logic on the magma side.  There's still some details to workout for more general support and code quality improvements, but looking for feedback on the architecture before we add more code.

* Adds a frontend context manager for compile guards.  For now we will support only one level of guards (i.e. no nesting).  This context manager injects itself into the definition context placer logic, so instances being placed know if they're being placed in the context of a compile guard.  
* Adds a pass that processes compile guards.  This searches for instances that were placed inside a compile guard and "moves" them to a separate circuit definition.  To do this, we must collect any drivers that are coming from outside the compile guard context, and have them be inputs to the circuit.  The pass removes these instances and adds a new instance of the compile guard wrapper circuit and wires up the external drivers.  The instance of the wrapper circuit has coreir metadata that indicates the compile guard string, with the idea that coreir will simply wrap the instance statement inside an ifdef when generating verilog.

Issues:
* Handling drivers in the general case will be tricky. For example, we could have a temporary value driving a port that is comprised of a subset of values from internal to the compile guard, and another that are external.  To handle this, we'll need to determine whether a driver is whole, otherwise recurse and manage the individual drivers.
* The connection between the compile guard context manager and placer/instance logic feels like monkey patching, we could probably clean this up by inverting the relationship and having the placer query context for the active compile guard.
* We should add a proper API for handling the instance migration to a new circuit.  The approach of having a pass move instances after the fact seemed simpler than opening and managing a new circuit definition on the fly (since we don't know all the external drivers yet so we'd have to incrementally build the IO).  However, right now we just replace the instance list in the circuit object (unsure if this is safe in general, but seems to work).  I think as a first pass, we simply need an API to remove instances.
* We should probably restrict it such that any logic inside a compile guard cannot be used outside of it.  In theory you might want to have different drivers produced in different compile guards, but I don't think this is a use case we need to support until someone demands the feature.  Until then, we should add a check to ensure that no drivers are being used outside of the context (technically this will eventually error when we try to compile a circuit and that instance has been moved to a different definition).


rsetaluri:
Checkout #918. This pattern uses the `CircuitBuilder` class (a good example of when this is useful) to stage and collect the instances. Once everything has been collected, we finalize the class by wrapping all of the instances and adding a circuit interface in the middle of the existing connections.

Overall, the high-level idea is the same as your prototype: stage a context to collect all the instances inside the compile guard, and then do a "pass" (using this work informally) to wrap those instances in a circuit def.

However, I think it is a good idea to isolate the idea of a "compile guard" and make it a isolated/separate feature, rather than a first-class member of the circuit creation flow. Notice that this change doesn't require any changes to `DefinitionContext` or `Placer`. We're able to do this by using the `CircuitBuilder` pattern. This is more maintainable IMO. Also, this makes it so that we don't have to copy instances or rewire internal connections between internal instances. All we need to do is find all external drivers and add ports, and then `unwire(value, port); wire(value, new_port); wire(new_port, port)`.

Since my prototype requires the same logic for wrapping instances, some of your points still remain such as:
* Handling drivers in the general case will be tricky.
* We should probably restrict it such that any logic inside a compile guard cannot be used outside of it.
* Common-ized logic for wrapping instances (this is probably a generic kind of thing)

Since these are separated, I figure we can iterate on that separately, and give the general feature to try (as in this PR as well). I'll clean up a bunch of stuff, just wanted to get your thoughts on this pattern.

PR #916, opened 2021-02-19
Fix always True assertion

leonardt:
In a refactor, this assertion got changed into a multiline using
parentheses, but the parens wrap both the condition and error message.
This fixes the assertion to only wrap the condition in the multiline
parentheses to avoid an always True assertion


rsetaluri (approved):
good catch

PR #917, opened 2021-02-20
[Fix #898] Raise exception when circuit has errors

### LINKED ISSUE #898: Magma should raise an exception when a design contains wiring errors
Here's an example of a circuit that has a wiring error, but will compile
```python
class T(m.Product):
    a = m.In(m.Clock)
    b = m.In(m.Clock)

class Foo(m.Circuit):
    io = m.IO(I=In(T), O=m.Out(m.Clock))
    m.wire(io.I, io.O)

m.compile("build/Foo", Foo)
```
This will produce the following error output:
```
ERROR:magma:Cannot wire Foo.I (Tuple(a=Out(Clock),b=Out(Clock))) to Foo.O (In(Clock))
```

However, it also generates a verilog file since the backend automatically wires up the undriven clock (Foo.O) to a default driver and compiles fine through coreir.

We know our users are not always the best at reading error messages, especially when their code contains a bunch of warnings that might hide the errors.

I propose that for each circuit definitions, we track the number of logged errors.  Then, when compiling a design we raise an exception at the Python level if the number of errors is non zero for all the circuits in a design.  We could then report all the errors that have been collected.

Another useful mode might be to have errors scoped per circuit.  So, at the end of each circuit definition, check if any errors have been logged and raise an error if so.  The idea here is that while we may not want to stop on every error (so we can collect multiple errors and report them at once, similar to how many software compilers will report more than one error if possible), but it may be useful to limit multiple errors on a per-circuit basis (since if there are multiple errors in a circuit, we probably want to fix those first before looking at other errors in the design).

leonardt:
This adds logic to the staged logger flow to store whether any error
messages were recorded during the circuit definition.

If an error was encountered, the backend will raise an exception rather
than proceeding.

This should avoid difficult to parse errors in the backend when
operating on a malformed circuit.  This should also hopefully surface
the error message to the user without less visual noise (e.g. a
traceback from an exception in the backend, leading them to believe it
could be an internal compiler error or something).

The hope is that if a user doesn't want something to be an error, then
it should be explicitly marked by the user.  E.g. if they're leaving a
value undriven on purpose (so a subsequent pass can drive it), then they
should mark it with `undriven()` and the downstream pass should be
responsible for handling the undriven value.  This is the less common
case, so I think it's better for these users to have to explicitly do
things to avoid errors (so the common case when something is an error is
caught by default).


rsetaluri:
@leonardt updated to use a separate pass instead of doing it in the backend directly. let me know what you think of this design.

rsetaluri (approved):

leonardt:
My one concern with doing it in a separate pass is performance.  Since we're iterating over the circuits during the compilation step anyways, it seemed reasonable to do it there.  Doing this as a separate pass may be better for code organization, but comes at the cost of having to traverse the entire hierarchy another time.  This cost is particularly troublesome in the case where the circuit has no errors (is the common case?).  It may be worth profiling the two options on the garnet compile to see how much cost it incurs (if it's trivial, then I'm fine with the separate pass).

rsetaluri:
Good point about performance. But I think the same can be said of the other passes, for example, we call `insert_coreir_wires(); insert_wrap_casts();` (maybe others upstream) which each traverses over the hierarchy separately. The actual meat of the pass is very simple, so my suspicion is that the cost (if any) would be in the instance graph traversal to get the list of definitions on which the pass needs to be run. Even if there are 100-1000s of module defns, the actual iteration over them I would guess would be trivial. I can check this for the CGRA.

Performance w.r.t. passes is a good question in general. Ideally, passes can avoid re-doing work/traversal when necessary but still provide the abstraction of independence (the main conceptual benefit of passes to begin with). Here are two ideas to address this:

- Simple caching of the instance hierarchy, i.e. the list of circuits on which the pass needs to be run. _If_ the hypothesis is true that most of the **overhead** of separate passes is in the extraction of the list of definitions, then this can be overcome easily by sharing that list across different passes. Of course, this needs to be done carefully to ensure that the list is still correct (i.e. a pass doesn't itself change that list), but conceptually this would be pretty easy. Might even be a good time to add a pass manager to manage this stuff. As a first try, this could even be done manually, e.g. `MyPass(main, cache).run()`.
- Doing a "loop fusion" kind of thing on passes. This requires knowing that there are no "RAW" type dependencies between passes, but for simple passes this might be true. We could assume it to be true and let the designer ensure it themselves, to avoid complications. So this would look like `fuse_passes(main, [insert_coreir_wires, insert_wrap_casts, find_errors])`.

leonardt:
Both ideas make sense.  I think having a class of passes that don't introduce dependencies (e.g. only look/modify a single circuit) would allow us to "fuse" by simply calling a series of "run" functions during traversals.  However an example like uniqufiication can't be fused since it needs to see multiple circuits before doing changes. 

Adding cacheing of the hierarchy also seems like a good idea, and we can update the Pass API to require some way of telling whether the hierarchy cache has been invalidated (e.g. return true when modifying the hierarchy)

rsetaluri:
@leonardt ok, I tested this pass on the 32x16 CGRA, and it took about ~1.3s (averaged over a bunch of runs), so it's non-zero but still pretty small in the entirety of the full compilation. (The whole compilation takes 10s of minutes.) I think this is good to merge.

Also #923 tracks the new desired features for passes.

PR #918, opened 2021-02-21
Compile guard updates

rsetaluri:


rsetaluri:
@rsetaluri TODOs

-- Test that inline verilog works here
-- Rename compile_guard to ifdef. I think that ifdef is more clear that it is a verilog-level thing, not a mamga level thing, as compile guard might imply. So `with m.ifdef("ASSERT_ON"):`.

rsetaluri:
Another TODO:
-- Figure out naming for new ports. Right now instance names are not known since they are in a staged placer. Another option is to just make it `port_0`, `port_1`, etc. If we have weird non-whole drivers and stuff it might be hard to name it well anyway, so a linear indexing may not be that bad after all.

rsetaluri:
Also note: I think this will work out of the box for nested ifdef's.

leonardt:
re: naming, I think supporting this beyond verilog as a general feature in magma/coreir could be useful.  For example, in simulation or formal verification we could use some notion of compile guards (like ifdefs) to exclude certain aspects of the design without having to re-run compilation (i.e. use python level if statements).

magma/compile_guard.py
```diff
@@ -0,0 +1,116 @@
...
+        if ref is None:
+            if value is None:
+                self._process_undriven(port)
+                return
+            ref = value.name
+        if isinstance(ref, AnonRef):
+            if value.const():
+                return
+            # TODO(rsetaluri): Implement valid anon. values.
+            raise NotImplementedError()
```
leonardt, line 56: One challenging case to think about here is if the anon value is wired up after the compile guard block.  It may come from an external driver but that driver might not be created yet.  I think because we assume that values only flow into the compile guard, if we encounter an undriven anon, we can assume the external driver case, but we should be sure to test this.

magma/compile_guard.py
```diff
@@ -0,0 +1,116 @@
...
+    def _process_undriven(self, port):
+        # TODO(rsetaluri): Add other system types
+        if not isinstance(port, Clock):
+            return
+        T = type(port).undirected_t
+        if T in self._system_types_added:
+            return
+        name = str(port.name)
+        self._add_port(name, In(T))
+        self._system_types_added[T] = name
```
leonardt, line 82: Can this just be a set? Doesn't look like name is used later on.

leonardt:
Overall the pattern looks good to me and cleaner than the first implementation

leonardt:
> Another TODO:
> -- Figure out naming for new ports. Right now instance names are not known since they are in a staged placer. Another option is to just make it `port_0`, `port_1`, etc. If we have weird non-whole drivers and stuff it might be hard to name it well anyway, so a linear indexing may not be that bad after all.

I think for now we shouldn't worry about names.  E.g. for magma inline verilog we generate wires as `_magma_inline_wire_{i}` and users don't seem to mind.

magma/compile_guard.py
rsetaluri: Done.

rsetaluri:
> re: naming, I think supporting this beyond verilog as a general feature in magma/coreir could be useful. For example, in simulation or formal verification we could use some notion of compile guards (like ifdefs) to exclude certain aspects of the design without having to re-run compilation (i.e. use python level if statements).
Makes sense. That's fine with me.

rsetaluri:
> > Another TODO:
> > -- Figure out naming for new ports. Right now instance names are not known since they are in a staged placer. Another option is to just make it `port_0`, `port_1`, etc. If we have weird non-whole drivers and stuff it might be hard to name it well anyway, so a linear indexing may not be that bad after all.
> 
> I think for now we shouldn't worry about names. E.g. for magma inline verilog we generate wires as `_magma_inline_wire_{i}` and users don't seem to mind.

Fair point. I'll go with convenient unique names.

magma/compile_guard.py
rsetaluri: Yeah that's a good point. So if I understand correctly you're talking about the case like

```python
class Top:
    ...
    value = ...
    tmp = T()
    with compile_guard(...):
        f.assert_immedate(tmp == 0)
    tmp @= value
```

Good point. This could be part of the `compile_guard` spec. If it really turns out that people want to do this frequently, then I think we could make all of this logic a late(r) binding pass as you had it. That would of course be more robust to such things.

rsetaluri:
> Also note: I think this will work out of the box for nested ifdef's.

Ok, I was wrong. It does not work for nested contexts:

```python
with compile_guard("a"):
    with compile_guard("b"):
        ... # will fail if refers to values from outside of "a"
```

If what's desired is in Verilog
```verilog
`ifdef OUTER
`ifdef INNER

<logic>

`endif
`endif
```

then we can do this much simpler by expanding the interface to `m.compile_guard(["OUTER", "INNER"])` which means nested conditions. should be pretty trivial in that case.

If what's desired is actual nesting:

```verilog
`ifdef OUTER

<some logic>

`ifdef INNER

<some other logic>

`endif
`endif
```

then this willbe harder.

leonardt:
I don't think we need to support general nesting for now.  I think the simple case for multiple guards might be useful.  We could also achieve it with an `&` operator (and maybe add `|` too), but also don't think this is necessary for a first release of the feature.

rsetaluri:
Ok. In that case, I think the PR is good to go for a first release. Agreed more features can come in the future.

By the way, I think if we move the logic to `_CompileGuardBuilder._finalize()` (instead of `pre_finalize()` it would actually solve *both* the nested problem and the external-driver problem.

PR #920, opened 2021-02-22
Symbol table

leonardt:


magma/backend/coreir/coreir_compiler.py
```diff
@@ -64,12 +65,16 @@ def __init__(self, main, basename, opts):
 
     def compile(self):
         result = {}
+        result["symbol_table"] = symbol_table = SymbolTable()
         insert_coreir_wires(self.main)
         insert_wrap_casts(self.main)
         backend = self.backend
         opts = _make_opts(backend, self.opts)
+        opts["symbol_table"] = symbol_table
```
rsetaluri, line 73: Note: there might be a more elegant/proper way to pass this around than in `opts`. (the general mechanism of just attaching it somewhere is fine, but `opts` might be a misleading place to put it, although it is very convenient. not sure on this one.)

magma/backend/coreir/coreir_transformer.py
```diff
@@ -361,9 +370,21 @@ def __init__(self, backend, opts, decl):
         self.coreir_module = None
 
     def run_self(self):
-        self.coreir_module = self.run_self_impl()
-
-    def run_self_impl(self):
+        self.coreir_module = self._run_self_impl()
+        # Set symbol table data.
+        # NOTE(rsetaluri): For some reason, fetching the CoreIR names, re-orders
```
rsetaluri, line 375: Note: read note here. something is broken downstream making this not right for me.

leonardt: Can you try this change: https://github.com/rdaly525/coreir/pull/985

rsetaluri:
@leonardt going to merge unless you have any concerns

PR #922, opened 2021-02-23
Miscellaneous updates

rsetaluri:
-- Updates to CircuitBuilder interface + more tests + update tests
-- Fixes to interface (for repr purporses)
-- Fixes zext

### DIFF
--- a/tests/test_ext.py
+++ b/tests/test_ext.py
@@ -52,3 +52,41 @@ def test_types(T):
     if T is m.SInt:
         c = a.sext(ext)
         assert isinstance(c, T)
+
+
+def test_ext_n():
+    value = m.SInt[32]()
+
+    # zext TypeError.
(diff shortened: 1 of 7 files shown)


tests/test_ext.py
```diff
@@ -52,3 +52,41 @@ def test_types(T):
...
         assert isinstance(c, T)
+
+
+def test_ext_n():
+    value = m.SInt[32]()
+
+    # zext TypeError.
+    with pytest.raises(TypeError) as pytest_e:
+        m.zext(value, None)
+        assert False
```
leonardt, line 63: The assert False is unnecessary in this pattern since `pytest.raises` will raise an exception if `TypeError` isn't raised

leonardt (approved):

tests/test_ext.py
rsetaluri: Done.

PR #926, opened 2021-02-25
[Fixes #925] Add test/support for old style circuits with compile_guard

### LINKED ISSUE #925: compile guard doesn't work with oldstyle circuits
Example:
```python
    class _Top(m.Circuit):
        IO = ["I", m.In(m.Bit), "O", m.Out(m.Bit)] + m.ClockInterface()

        @classmethod
        def definition(io):
            with m.compile_guard("COND", defn_name="COND_compile_guard"):
                out = m.Register(m.Bit)()(io.I)

            io.O @= io.I
````

### PR DESCRIPTION
**Pull request by leonardt**

### CODE REVIEW

File: magma/placer.py [hunk 1]
```diff
@@ -129,6 +129,9 @@ def place(self, inst):
             inst.stack = inspect.stack()
         _setup_view(inst)
 
+    def finalize(self, defn):
```
rsetaluri, line 132: ik functionally it doesn't matter, but let's add the same `if self._finalized: raise Exception()` logic

magma/circuit.py
```diff
@@ -652,6 +652,7 @@ def __new__(metacls, name, bases, dct):
                                 "instead", debug_info=self.debug_info)
                 with _DefinitionContextManager(self._context_):
                     self.definition()
+                self._context_.finalize(self)
```
rsetaluri, line 655: why didn't this cause problems before with inline verilog?

rsetaluri (approved):

magma/circuit.py
leonardt: Inline verilog is processed in a pass, not during the finalize logic. The file open/display logic is finalized there, but perhaps that wasn't being used (I think they ended up building their own logging library that generates inline verilog)

PR #927, opened 2021-02-26
Add support for verilog field in coreir backend

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/987

This was a feature we used to support in magma (see first example
https://github.com/phanrahan/magmathon/blob/master/notebooks/advanced/verilog.ipynb)
that used the verilog backend. This adds support for the same feature in
the coreir backend.

This simply allows the user to define circuit interface in magma,
but the body in verilog. This is slightly different than inline verilog
in that we allow the user to drive signals from the verilog body string.
Otherwise, if we drove a signal from inline_verilog, we'd get an
undriven error in magma/coreir since we aren't aware of the driver from
the string.

### DIFF
--- a/magma/backend/coreir/coreir_transformer.py
+++ b/magma/backend/coreir/coreir_transformer.py
@@ -249,6 +249,10 @@ class DefinitionTransformer(TransformerBase):
             metadata = json.dumps({"verilog_string": self.defn.verilogFile})
             self.coreir_module.add_metadata("verilog", metadata)
             return coreir_defn
+        if hasattr(self.defn, "verilog") and self.defn.verilog:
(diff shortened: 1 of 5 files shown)


rsetaluri (approved):

rsetaluri:
From https://travis-ci.org/github/phanrahan/magma/builds/760524255 it looks like a CoreIR issue?

leonardt:
Depends on a coreir PR which adds the feature

leonardt:
https://github.com/rdaly525/coreir/pull/987

rsetaluri:
Oh didn't see that. Looks good

rsetaluri (approved):

PR #930, opened 2021-02-26
Comb set name

leonardt:
In preparation for the combinational symbol table, this adds a pass that inserts a `set_name` function call in all the assignments generated by SSA.  In the normal case, we'll get a temporary wire with the name (I chose this versus renaming the value since this seemed much easier to to do, for example we can't realy rename an output port on a circuit instance).  This includes a safeguard hook that allows set_name to be called with an unexpected type of value, in which case it will just return the value unnamed.  This is actually required for the use case of calling a combinational circuit with hwtypes values (e.g. for a functional model).  In this case, we can't call set_name on the hwtypes type (perhaps we should update combinational to use the non-rewritten version of the function? rather than the post-ssa version?)


magma/syntax/combinational2.py
```diff
@@ -1,14 +1,45 @@
...
 
 from .util import build_io_args, build_call_args, wire_call_result
+from magma.set_name import set_name
+
+
+class NameSetter(cst.CSTTransformer):
+    def __init__(self, set_name_var):
+        self.set_name_var = set_name_var
+
+    def leave_Assign(self, original_node, updated_node):
```
cdonovick, line 23: You might be interested in the match decorators.  I have no problem with this though. 

https://github.com/leonardt/ast_tools/blob/master/ast_tools/transformers/normalizers.py#L9
https://libcst.readthedocs.io/en/latest/matchers.html

cdonovick (approved):
looks good to me.  I think post ssa is probably best?  You could probably query the table  I will generate for the original names if you want to have `x_0 = v` rewritten as `x_0 = set_name("x", v)`

leonardt:
We want the name to be the SSA name, since we need it to do be distinct in the generated hardware (we can't have multiple wires with the same name).  We'll use the mapping from line+variable to SSA name to figure out which hardware wire it corresponds to

magma/syntax/combinational2.py
leonardt: Hmm, nice seems convenient

PR #932, opened 2021-02-27
Add inverted compile guard support

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/988

Support for the `ifndef` pattern has been requested.

### DIFF
--- a/docs/compile_guard.md
+++ b/docs/compile_guard.md
@@ -75,5 +75,5 @@ by using the `ASSERT_ON` define.
 Note that logic inside a compile guard cannot be used outside of a compile
 guard (otherwise a signal may not have a driver when the guard is disabled).
 
-To generate an `ifndef` instead of `ifdef`, pass the argument `invert=True` to
-`m.compile_guard`.
+To generate an `ifndef` instead of `ifdef`, pass the argument
+`type="undefined"` to `m.compile_guard` (the default is `type="defined"`).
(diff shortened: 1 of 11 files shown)


docs/compile_guard.md
```diff
@@ -74,3 +74,6 @@ by using the `ASSERT_ON` define.
 
 Note that logic inside a compile guard cannot be used outside of a compile
 guard (otherwise a signal may not have a driver when the guard is disabled).
+
+To generate an `ifndef` instead of `ifdef`, pass the argument `invert=True` to
```
rsetaluri, line 78: I think `invert` might not be the best phrasing (e.g. confusion between `ifdef !X` vs `ifndef X`). Perhaps `undefined=True` might be better? Same for the related PRs.

Alternatively,  we could have a `type` argument, default to `type="defined"` and allow `type="undefined"`. TBH, I like this more for the explicitness (and flexibility, although I don't think Verilog conditional compilation allows for much more...); but, the single boolean flag is prob more ergonomic.

magma/compile_guard.py
```diff
@@ -120,7 +127,8 @@ def __enter__(self):
         if self._defn_name is None:
             self._defn_name = _CompileGuard._new_name()
         if self._state is None:
-            ckt = _CompileGuardBuilder(self._defn_name, self._cond)
+            ckt = _CompileGuardBuilder(self._defn_name, self._cond,
```
rsetaluri, line 130: nit: looks like this can fit in 80 chars?

rsetaluri (approved):

PR #933, opened 2021-03-03
Fix error with VCC

leonardt:
Before this change, the added test would produce this error:
```
>       raise NotImplementedError(ref, type(ref))
E       NotImplementedError: (VCC, <class 'magma.ref.NamedRef'>)
```

The issue is that ref is not an AnonRef (where const values are
handled), rather it's a NamedRef.  This changes it so that if a value is
const (regardless of the type of the ref), it should work.


rsetaluri (approved):

tests/test_compile_guard.py
```diff
@@ -145,3 +145,19 @@ class _Top(m.Circuit):
     assert m.testing.check_files_equal(
         __file__, f"build/test_compile_guard_basic_undefined.json",
         f"gold/test_compile_guard_basic_undefined.json")
+
+
+def test_basic_vcc():
```
rsetaluri, line 150: nit: `test_vcc`

PR #934, opened 2021-03-03
Add support for verilog prefix

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/991

### DIFF
--- a/magma/backend/coreir/coreir_compiler.py
+++ b/magma/backend/coreir/coreir_compiler.py
@@ -38,7 +38,7 @@ def _make_verilog_cmd(deps, basename, opts):
         cmd += " --disable-ndarray"
     if opts.get("verilator_compat", False):
         cmd += " --verilator_compat"
-    if opts.get("verilog_prefix", False):
+    if opts.get("verilog_prefix", ""):
         cmd += f" --verilog-prefix {opts['verilog_prefix']}"
     return cmd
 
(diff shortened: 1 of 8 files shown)


tests/test_verilog/test_bind.py
```diff
@@ -64,12 +65,13 @@ class Main(m.Circuit):
...
+                             ' --top-module bar_foo_Main')
         # TODO: For now, we ignore duplicate modules since each monitor will
         # regenerate coreir primitives.  # since these have the same definition
         # it's okay, but in general this is a bad flag to have on because if
         # you do have duplicate modules with different definitions, you'd want
         # to catch that
-        # When we merge the inline verilog module branch, this will be resolved since
-        # bind modules can be represented explicitly in the graph and compiled
-        # by coreir, so only one definition per primitive will be emitted
+        # When we merge the inline verilog module branch, this will be resolved
```
rsetaluri, line 74: is this an outdated comment? "When we merge the inline..." seems like it would be outdated?

magma/backend/coreir/coreir_compiler.py
```diff
@@ -38,6 +38,8 @@ def _make_verilog_cmd(deps, basename, opts):
         cmd += " --disable-ndarray"
     if opts.get("verilator_compat", False):
         cmd += " --verilator_compat"
+    if opts.get("verilog_prefix", False):
```
rsetaluri, line 41: nit
```suggestion
    if opts.get("verilog_prefix", ""):
```

rsetaluri (approved):

tests/test_verilog/test_bind.py
leonardt: I've updated the comment, this isn't resolved by inline verilog because bind can't use it (the bind statement appears outside of a module, whereas inline verilog only supports verilog inside a module definition).  I think when the comment was written I was thinking that bind would use inline verilog but for now we can't do it (we need to add an explicit bind construct to coreir)

PR #935, opened 2021-03-06
Initial support for compile guard outputs

leonardt:
Depends on https://github.com/rdaly525/coreir/pull/992

Needs more testing for various cases (e.g. tuples/arrays, only partial
values being used)


magma/compile_guard.py
```diff
@@ -55,9 +56,51 @@ def pre_finalize(self):
...
+        if isinstance(ref, ArrayRef):
+            return self._get_top_ref(ref.array.name)
+        return ref
+
+    def _is_external(self, value):
+        top_ref = self._get_top_ref(value.name)
+        if isinstance(top_ref, DefnRef):
+            return True
+        if isinstance(top_ref, InstRef):
+            return top_ref.inst not in self._instances
```
rsetaluri, line 82: These should all be free functions outside of the class

rsetaluri: Except for `_is_external`

magma/compile_guard.py
```diff
@@ -70,21 +113,21 @@ def _process_port(self, port, ref=None, kwargs=None):
             # TODO(rsetaluri): Implement valid anon. values.
             raise NotImplementedError()
         if isinstance(ref, DefnRef):
-            self._rewire(port, value)
```
rsetaluri, line 73: maybe we can re-use `_is_external` here?

magma/compile_guard.py
```diff
@@ -55,9 +56,51 @@ def pre_finalize(self):
...
+            return self._get_top_ref(ref.array.name)
+        return ref
+
+    def _is_external(self, value):
+        top_ref = self._get_top_ref(value.name)
+        if isinstance(top_ref, DefnRef):
+            return True
+        if isinstance(top_ref, InstRef):
+            return top_ref.inst not in self._instances
+        return False
```
rsetaluri, line 83: This doesn't handle anonymous values created outside the compile-guarded context right? Fine if so, should just add a comment here.

(e.g.
```python
x = Bit()
with compile_guard(...):
    inst = Ckt()
    x @= inst.O  # _is_external(x) == False, even though it is external.
```
)

But perhaps in that case it doesn't really matter?

rsetaluri: I might feel better if we `raise NotImplemented()` instead of returning `False` incorrectly. I imagine we can say that's a case we explicitly don't handle, rather than just handling it wrong.

magma/compile_guard.py
```diff
@@ -55,9 +56,51 @@ def pre_finalize(self):
...
+
+    def _process_output(self, port):
+        drivees = self._get_flat_drivees(port)
```
rsetaluri, line 86: I think this can actually be simplified as `as_bits(port).driving()`?

rsetaluri (commented):
The logic to extract the necessary outputs is good in general. Some feedback/iteration on the implementation though.

magma/compile_guard.py
leonardt: It works with anonymous values created inside the compile guard context that only refer to internal instances, so if we raise an error then we remove that use case.

leonardt: Actually it seems that the input anons are raising an error, so might as well be consistent, I can have it raise an error

PR #939, opened 2021-03-17
Fix error with reset cast in python sim flatten

leonardt:
Before, we'd get this error:
```
ERROR:magma:Cannot wire Foo_flattened.RESET (Out(Reset)) to coreir_commonlib_mux2x4_inst0.in.sel[0] (In(Bit))
```

This is because inside the register code, we cast a register to bit (as
an input to a mux).

This fix changes the flattening code to cast a new source bit before
wiring it to a sink bit since at this point we can assume the user code
was valid (i.e. they performed a cast).

### DIFF
--- a/magma/transforms.py
+++ b/magma/transforms.py
@@ -191,7 +191,9 @@ def wire_new_bit(origbit, newbit, cur_scope, primitive_map, bit_map, old_circuit
 
     newsource = get_new_source(source_qual, primitive_map, old_circuit, new_circuit)
     # Convert newsource because it might have been casted in the user code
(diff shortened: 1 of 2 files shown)


magma/transforms.py
```diff
@@ -1,5 +1,6 @@
 from collections import namedtuple, OrderedDict
 from .circuit import CopyInstance, Circuit, IO
+from .conversions import convertbit
```
rsetaluri, line 3: nit: `from magma.conversions`

rsetaluri (approved):

PR #944, opened 2021-04-01
Support symbol table inlining

rsetaluri:


magma/symbol_table_utils.py
```diff
@@ -127,16 +151,46 @@ def tails(self):
...
+def _finalize_instance_names(module, instance, table, top=True):
+    tail = instance.tail()
+    if top:
+        if not tail.inlined:
+            table.set_instance_name(
+                module.name, instance.name, (SYMBOL_TABLE_EMPTY, tail.name))
+            return
+        else:
+            table.set_instance_name(
+                module.name, instance.name, (SYMBOL_TABLE_INLINED_INSTANCE, ""))
```
leonardt, line 163: Could drop the else since there's a return

magma/symbol_table_utils.py
```diff
@@ -147,29 +201,58 @@ def finalize(self, table: SymbolTableInterface):
...
+                             for m in self._modules.values()
+                             if m.scope == self._scope}
         for in_module_name, out_module_name in module_names.items():
             if self._scope == 0:
                 src = _Module(in_module_name, self._scope)
                 root = src
                 self._modules[src.key()] = src
             else:
                 src = self._modules[(self._scope, in_module_name)]
+                unmapped_srcs.pop(src.name)
```
leonardt, line 239: Could use del instead of pop since you're not using the return value

leonardt (approved):

magma/symbol_table.py
```diff
@@ -31,21 +31,24 @@ def __str__(self):
...
-def _unwrap_value(value):
-    if not isinstance(value, _Sentinel):
-        return value
-    return value.string
-
-
-def _is_union(T):
+def _is_tuple_annotation(T):
     try:
         origin = T.__origin__
```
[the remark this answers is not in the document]
rsetaluri: Actually realized this check is wrong...I needed to check that `origin is tuple` after the try-catch block.

PR #945, opened 2021-04-05
Migrate bind to Generator2

leonardt:
Adds support for bind generators, updates tests

### DIFF
--- a/magma/generator.py
+++ b/magma/generator.py
@@ -67,7 +67,8 @@ def _make_type(cls, *args, **kwargs):
     dct = DefineCircuitKind.__prepare__(name, bases)
     cls.__init__(dummy, *args, **kwargs)
     dct.update(dict(dummy.__dict__))
-    # Use circuit bind for the instance
+    # Use circuit bind for the instance,
+    # rather than inherited Generator2 bind
     dct["bind"] = classmethod(DefineCircuitKind.bind)
     circuit = DefineCircuitKind.__new__(cls, name, bases, dct)
     for gen in cls.bind_generators:
--- a/magma/generator.py
+++ b/magma/generator.py
@@ -67,7 +67,12 @@ def _make_type(cls, *args, **kwargs):
     dct = DefineCircuitKind.__prepare__(name, bases)
     cls.__init__(dummy, *args, **kwargs)
     dct.update(dict(dummy.__dict__))
-    return DefineCircuitKind.__new__(cls, name, bases, dct)
+    # Use circuit bind for the instance
+    dct["bind"] = classmethod(DefineCircuitKind.bind)
+    circuit = DefineCircuitKind.__new__(cls, name, bases, dct)
+    for gen in cls.bind_generators:
+        gen.generate_bind(circuit, *args, **kwargs)
+    return circuit
 
 
 class _Generator2Meta(type):
@@ -75,6 +80,8 @@ class _Generator2Meta(type):
 
     def __new__(metacls, name, bases, dct):
         bases = bases + (DefineCircuitKind,)
+        assert "bind_generators" not in dct
+        dct["bind_generators"] = []
         return type.__new__(metacls, name, bases, dct)
 
     def __call__(cls, *args, **kwargs):
@@ -103,6 +110,10 @@ class Generator2(metaclass=_Generator2Meta):
 
     def __call__(cls, *args, **kwargs):
         return DefineCircuitKind.__call__(cls, *args, **kwargs)
+
+    @classmethod
+    def bind(cls, monitor):
+        cls.bind_generators.append(monitor)
 
 
 def reset_generator_cache():
--- a/magma/generator.py
+++ b/magma/generator.py
@@ -67,13 +67,13 @@ def _make_type(cls, *args, **kwargs):
     dct = DefineCircuitKind.__prepare__(name, bases)
     cls.__init__(dummy, *args, **kwargs)
     dct.update(dict(dummy.__dict__))
-    # Use circuit bind for the instance,
-    # rather than inherited Generator2 bind
+    # NOTE(leonardt): We need to override the Generator2 classmethod bind with
+    DefineCircuitKind.bind for generator instances (circuits).
     dct["bind"] = classmethod(DefineCircuitKind.bind)
-    circuit = DefineCircuitKind.__new__(cls, name, bases, dct)
+    ckt = DefineCircuitKind.__new__(cls, name, bases, dct)
     for gen in cls.bind_generators:
-        gen.generate_bind(circuit, *args, **kwargs)
-    return circuit
+        gen.generate_bind(ckt, *args, **kwargs)
+    return ckt
 
 
 class _Generator2Meta(type):
--- a/magma/generator.py
+++ b/magma/generator.py
@@ -81,8 +81,7 @@ class _Generator2Meta(type):
 
     def __new__(metacls, name, bases, dct):
         bases = bases + (DefineCircuitKind,)
-        assert "bind_generators" not in dct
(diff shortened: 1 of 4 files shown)


magma/generator.py
```diff
@@ -104,6 +111,10 @@ def __new__(metacls, name, bases, dct):
     def __call__(cls, *args, **kwargs):
         return DefineCircuitKind.__call__(cls, *args, **kwargs)
 
+    @classmethod
+    def bind(cls, monitor):
```
rsetaluri, line 115: im not sure im understanding how this works. is there a simpler example test you can write (can be dumb simple). I'm just not seeing where this classmethod gets called.

leonardt: The reason you don't see it in the changes is because this simply updates old code using m.Generator to use m.Generator2, so the test code that calls it didn't change.  The bind call can be found here: https://github.com/phanrahan/magma/blob/master/tests/test_verilog/rtl_monitor.py#L46

We bind a monitor generator to a magma generator, so every time a generator is invoked to create a circuit, we call the monitor generator to create a monitor for that instance.

leonardt: The old m.Generator code can be found here:
* https://github.com/phanrahan/magma/blob/master/magma/generator.py#L20-L21
* https://github.com/phanrahan/magma/blob/master/magma/generator.py#L51-L53

tests/test_verilog/test_bind.py
```diff
@@ -7,7 +7,7 @@
 
 
 def test_bind():
-    RTL4 = RTL.generate(4)
```
rsetaluri, line 10: we don't use `import rtl_monitor` anywhere in this file right?

magma/generator.py
```diff
@@ -67,14 +67,22 @@ def _make_type(cls, *args, **kwargs):
...
+    return circuit
 
 
 class _Generator2Meta(type):
     _cache = weakref.WeakValueDictionary()
 
     def __new__(metacls, name, bases, dct):
         bases = bases + (DefineCircuitKind,)
+        assert "bind_generators" not in dct
+        dct["bind_generators"] = []
```
rsetaluri, line 85:
```suggestion
        assert dct.setdefault("bind_generators", []) == []
```

rsetaluri: Actually, can't just move `bind_generators = []` into the class namespace of `classs Generator2`, i.e.

```python
class Generator2(metaclass=_Generator2Meta):
    bind_generators = []
    ...
```

magma/generator.py
```diff
@@ -67,14 +67,22 @@ def _make_type(cls, *args, **kwargs):
     dct = DefineCircuitKind.__prepare__(name, bases)
     cls.__init__(dummy, *args, **kwargs)
     dct.update(dict(dummy.__dict__))
-    return DefineCircuitKind.__new__(cls, name, bases, dct)
+    # Use circuit bind for the instance,
+    # rather than inherited Generator2 bind
+    dct["bind"] = classmethod(DefineCircuitKind.bind)
+    circuit = DefineCircuitKind.__new__(cls, name, bases, dct)
```
rsetaluri, line 73: This we can do by making `bind` a member function on `Generator2` right?

rsetaluri (commented):
Some more comments. My main thing for this is trying to make this stuff more readable/maintainable/understandable by moving stuff that can be represented as standard class-construction rather than putting it into the metaclass flow, e.g. defining things normally on the base class rather than inserting things into the dict of metacls.__new__.

I think the two places I commented, it makes sense to do that. For this loop:
```python
    for gen in cls.bind_generators:
        gen.generate_bind(circuit, *args, **kwargs)
```

I'm not sure we can do that anywhere else.

tests/test_verilog/test_bind.py
leonardt: We need to import it so that the bind call is done (the pattern is import triggers the bind, so by including a monitor file, the bind logic is added, this reflects how the pattern is done in verilog, since the bind is added by including addition files).

leonardt:
We can't set bind_generators in the Generator2 class since we need a different bind_generators for each sub class rather than a shared one inherited from the parent class

magma/generator.py
leonardt: Can you clarify? The issue is that Generator2 defines a bind method, so that is used instead of looking up the Circuit base class method.

rsetaluri: Like what if we had

```python
class Generator2:
    def bind(self, ...):  # now self is the circuit
        DefineCircuitKind.bind(self, ...)
```

Actually why does the MRO use the `Generator2.bind` instead of `DefineCircuitKind.bind` anyway?

rsetaluri:
> We can't set bind_generators in the Generator2 class since we need a different bind_generators for each sub class rather than a shared one inherited from the parent class

Ah yeah, that's right. I guess this is the only way to do this then.

magma/generator.py
leonardt: I think because we use `cls` and have Circuit as a base (see here https://github.com/phanrahan/magma/pull/945/files#diff-8194325e1fce50138ceca2e91aaadbd8eac22e64573df83388bc066fdbcee985R66), so the default is to lookup in `cls` first, before looking at the bases.

leonardt: That doesn't work because `bind` is already a classmethod in Generator2 (used for binding a monitor generator to the generator), so we can't have both a classmethod and a member method with the same name.  I tried moving the classmethod to the metaclass but that also doesn't work, since it will use the member method.

rsetaluri: Ah yeah ok. Seems like a pretty interesting use case.

magma/generator.py
```diff
@@ -67,14 +67,21 @@ def _make_type(cls, *args, **kwargs):
...
+    for gen in cls.bind_generators:
+        gen.generate_bind(circuit, *args, **kwargs)
+    return circuit
```
rsetaluri, line 76:
```suggestion
    # NOTE(leonardt): We need to override the Generator2 classmethod bind with
    DefineCircuitKind.bind for generator instances (circuits).
    dct["bind"] = classmethod(DefineCircuitKind.bind)
    ckt = DefineCircuitKind.__new__(cls, name, bases, dct)
    for gen in cls.bind_generators:
        gen.generate_bind(ckt, *args, **kwargs)
    return ckt
```

rsetaluri: Last question: So the `generate_bind` method needs to have the same parameters as the bound generator? I ask because we pass `*args, **kwargs` down to `generate_bind`.

rsetaluri (approved):

magma/generator.py
```diff
@@ -67,14 +67,21 @@ def _make_type(cls, *args, **kwargs):
     dct = DefineCircuitKind.__prepare__(name, bases)
     cls.__init__(dummy, *args, **kwargs)
     dct.update(dict(dummy.__dict__))
-    return DefineCircuitKind.__new__(cls, name, bases, dct)
+    # NOTE(leonardt): We need to override the Generator2 classmethod bind with
```
rsetaluri, line 70: oops something got lost here

magma/generator.py
leonardt: The interface is "generated circuit, *<generator args>, **<generator kwargs>", so you the first argument is the result from the generator (so you have a handle to it for calling bind, as well as accessing internal signals to bind to).  Then, yes, the rest of the arguments should be the same as generator arguments (this is so the bind generator has a handle to them in case they want to use them in the code gen).

leonardt: This was added before we had Generator2 (hence the migration), I wonder if we could drum up something a little better that fits into the Generator2 style (e.g. this uses the staticmethod style that Generator had)

PR #947, opened 2021-04-07
Allow seq outputs to be named

### LINKED ISSUE #942: [Feature Request] user-named outputs for circuit.sequential2
Currently output names are generated as "O" for a single output and  "O0" .. "ON" for multiple outputs. It would be great if we could add an ability to name the outputs ports similar to naming the input ports.

Possible APIs (in order of my personal preference)

1)
```
    @m.sequential2()
    class A:
        ...
        def __call__(self, i0: m.Bit, i1: m.Bit) -> {"myport0" : m.Bit, "myport1": m.Bit}:
            ...
```

2) (Current one that Peak supports)
```
    @m.sequential2()
    class A:
        ...
        @name_outputs(myport0=m.Bit, myport1=m.Bit)
        def __call__(self, i0: m.Bit, i1: m.Bit):
            ...
```

3)
```
    class Outputs(m.Product):
        myport0=m.Bit
        myport1=m.Bit

    @m.sequential2()
    class A:
        ...
        def __call__(self, i0: m.Bit, i1: m.Bit) -> Outputs
            ...
```
This one could cause some ambiguity if you actually want a single port which is itself a Product type.

I am happy to look into implementing. @leonardt, thoughts?

cdonovick:
closes #942

### DIFF
--- a/magma/syntax/util.py
+++ b/magma/syntax/util.py
@@ -5,32 +5,49 @@ from ..circuit import Circuit
 from ..wire import wire
 
 
+def build_output_port_names(
+        annotations: Mapping[str, Any],
+        output_port_names: Optional[Sequence[str]] = None,
+    ) -> Sequence[str]:
+
+    try:
+        r_type = annotations['return']
+    except KeyError:
+        raise ValueError('No return annotation') from None
+
+    if output_port_names is None:
+        output_port_names = []
+        if isinstance(r_type, tuple):
+            for i, _ in enumerate(r_type):
+                output_port_names.append(f'O{i}')
+        else:
+            output_port_names.append('O')
+    elif isinstance(r_type, tuple):
+        if len(r_type) != len(output_port_names):
+            raise ValueError(f'recieved {len(output_port_names)} names expected {len(annotation)}')
+    elif len(output_port_names) != 1:
+        raise ValueError(f'recieved {len(output_port_names)} names expected 1')
+
+    return output_port_names
+
+
 def build_io_args(
         annotations: Mapping[str, Any],
         output_port_names: Optional[Sequence[str]] = None,
         ) -> Mapping[str, Any]:
     io_args = {}
 
+    output_port_names = build_output_port_names(annotations, output_port_names)
+
     for param, annotation in annotations.items():
         if param == "return":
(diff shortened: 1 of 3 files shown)


magma/syntax/util.py
```diff
@@ -1,18 +1,37 @@
...
+                    raise ValueError(f'recieved {len(output_port_names)} names expected {len(annotation)}')
             else:
-                io_args["O"] = Out(annotation)
+                if output_port_names is None:
+                    io_args["O"] = Out(annotation)
+                elif len(output_port_names) == 1:
+                    io_args[output_port_names[0]] = Out(annotation)
+                else:
+                    raise ValueError(f'recieved {len(output_port_names)} names expected 1')
+
```
leonardt, line 34: This might be simpler if we filled output_port_names with the default name(s) (i.e. "O"` or "O{i}") if the sequence isn't provided, then the logic would be the same for both cases? Probably not that much of an improvement if any though.

magma/syntax/util.py
```diff
@@ -23,14 +42,23 @@ def build_call_args(io, annotations):
...
+            for i, name in enumerate(output_port_names):
+                wire(getattr(io, name), call_result[i])
+
+
     else:
-        wire(io.O, call_result)
+        if output_port_names is None:
+            wire(io.O, call_result)
+        else:
+            wire(getattr(io, output_port_names[0]), call_result)
```
leonardt, line 64: Hmm, I guess if we fill output_port_names then we avoid the if statements here too (just use output_port_names, which will have the default values for None)

leonardt (approved):
LGTM, minor comment

rdaly525 (approved):
Test looks good!

magma/syntax/util.py
cdonovick: Yeah I actually thought about doing that but there are multiple call sites and I was feeling lazy.  I can move the building of the out_port_names to the top though.

magma/syntax/util.py
cdonovick: same issue as above.

cdonovick:
I will add another util to build default `output_port_names`

PR #948, opened 2021-04-08
Symbol table updates

rsetaluri:

PR #951, opened 2021-04-13
Add clog2safe function

leonardt:
Requested by George, here's a reference example implementation: https://github.com/KastnerRG/riffa/blob/master/fpga/riffa_hdl/functions.vh#L59

The idea is that the 1 parameter comes up often in highly parametrized designs, so sometimes it's useful to have a clog2safe function where you get a minimum size of 1 (e.g. instead of Bits[0] or having to call max(..., 1))


rsetaluri (approved):

rsetaluri:
@leonardt tests are failing bc #948 needs to be merged

rsetaluri:
@leonardt rebased, and now passing

Issue #929, opened 2021-02-26
Handle modified circuits in `isdefinition()`


rsetaluri:
Right now the method `isdefinition()` is static, i.e. it is computed at circuit creation time and not updated. This is problem if we do `ckt.open()` and insert new logic where there wasn't previously logic.

Additionally, the logic for isdefinition is fragmented in many places, ideally we should unify these.

Two ways to fix this:
-- Have `isdefinition` always be a dynamic method that implements the logic
-- Have `Circuit.open()` mark the circuit as modified, and lazily recompute the logic upon the next `isdefinition` call.

Either way that logic needs to be unified and encapsulated.

PR #954, opened 2021-05-13
Raise exception when trying to compile empty defn

leonardt:
This came up with an old-style class definition that had a typo in the
`definition` method (spelled it `definiiton`).  This resulted in a
segfault in coreir since it was trying to compile an empty top.  We
should improve the coreir error handling, but I think at this level it's
fine to raise an Exception when the user tries to do this, because
there's not really any reasonable use case for trying to compile an
empty circuit without a definition (we could just generate a verilog
declaration? but seems like an odd use case for magma that we don't need
to support)

### DIFF
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -42,7 +42,15 @@ def _get_basename(basename):
     return basename
 
 
+class MagmaCompileException(Exception):
+    pass
+
+
 def compile(basename, main, output="coreir-verilog", **kwargs):
+    if not getattr(main, "is_definition", False):
+        raise MagmaCompileException(
+            f"Trying to compile empty definition {main}"
+        )
     if hasattr(main, "circuit_definition"):
         main = main.circuit_definition
     basename = _get_basename(basename)
--- a/magma/compile.py
+++ b/magma/compile.py
@@ -47,10 +47,9 @@ class MagmaCompileException(Exception):
 
 
 def compile(basename, main, output="coreir-verilog", **kwargs):
-    if not getattr(main, "is_definition", False):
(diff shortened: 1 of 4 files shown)


magma/compile.py
```diff
@@ -42,7 +42,15 @@ def _get_basename(basename):
...
 
+class MagmaCompileException(Exception):
+    pass
+
+
 def compile(basename, main, output="coreir-verilog", **kwargs):
+    if not getattr(main, "is_definition", False):
+        raise MagmaCompileException(
+            f"Trying to compile empty definition {main}"
+        )
```
rsetaluri, line 53:
```suggestion
    if not isdefinition(main):
        raise MagmaCompileException(
            f"Trying to compile empty definition {main}")
```

need to add `from magma.is_definition import isdefinition` after L9.

rsetaluri (approved):

tests/test_wire/test_errors.py
```diff
@@ -198,8 +198,10 @@ class T(m.Product):
         b = m.In(m.Clock)
 
     class Foo(m.Circuit):
-        io = m.IO(I=In(T), O=m.Out(m.Clock))
+        io = m.IO(I=In(T), O=m.Out(m.Clock), x=m.Out(m.Bit))
```
rsetaluri, line 201: actually, this is weird. we should be checking driven outputs for isdefinition too.

rsetaluri: oh wait, nevermind. with the removal of `def definition()` method, we actually don't check for this. this should really be fixed. I'll look into it since there's #929 also.

PR #955, opened 2021-05-20
Disable automatic clock wiring logic when multiple clock drivers are …

leonardt:
…present

Fixes #889
Handles the case when the clock driver could be coming from an instance


magma/clock.py
```diff
@@ -134,29 +134,51 @@ def first(iterable):
...
+
+
+def get_clocks_from_defn(defn, clocktype):
+    clocks = []
+    for port in defn.interface.ports.values():
+        clocks += get_clocks(port, clocktype)
+    for inst in defn.instances:
+        # circular import because Wire depends on clock.py
+        # We could move default clock wiring logic to a different file?
+        from magma.primitives.wire import Wire
```
leonardt, line 166: I have a commit that does this refactor (moves clock wiring logic to a separate file to avoid the circular import), but I think I can save it for after the review because otherwise there's a lot of unrelated refactor changes from moving stuff around that will make this more complicated to review.

leonardt: https://github.com/phanrahan/magma/pull/956

magma/clock.py
```diff
@@ -134,29 +134,51 @@ def first(iterable):
     return next((item for item in iterable if item is not None), None)
 
 
-def get_first_clock(port, clocktype):
+def get_clocks(port, clocktype):
+    if port.is_input():
+        # We are looking for default drivers (outputs)
```
rsetaluri, line 139:
```suggestion
        # We are looking for default drivers (only need to consider outputs).
```

magma/clock.py
```diff
@@ -134,29 +134,51 @@ def first(iterable):
...
+        clks = []
+        for elem in port:
+            clks += get_clocks(elem, clocktype)
+        return clks
     if isinstance(port, Array):
-        return get_first_clock(port[0], clocktype)
-    return None
+        clks = get_clocks(port[0], clocktype)
+        if not clks:
+            # early exit to avoid traversing children when unnecessary
```
rsetaluri, line 151:
```suggestion
            # Exit early to avoid traversing children when unnecessary.
```

magma/clock.py
```diff
@@ -134,29 +134,51 @@ def first(iterable):
...
+            return clks
```
rsetaluri, line 152: should this just be `return []` (for readability)

magma/clock.py
```diff
@@ -134,29 +134,51 @@ def first(iterable):
...
+        for elem in port[1:]:
+            clks += get_clocks(elem, clocktype)
+        return clks
+    return []
+
+
+def get_clocks_from_defn(defn, clocktype):
```
rsetaluri, line 159: I think we should make this return a generator rather than a list for performance, since we only every get the first element. I think we can use `yield` + `itertools.chain` to do this concisely. Happy to help make this change.

magma/clock.py
```diff
@@ -134,29 +134,51 @@ def first(iterable):
...
-            for port in defn.interface.ports.values())
-    defnclk = first(clks)
-    if defnclk is None:
-        # Check recursive types
-        clks = (get_first_clock(port, clocktype)
-                for port in defn.interface.ports.values())
-        defnclk = first(clks)
-    if defnclk is None:
+    clks = get_clocks_from_defn(defn, clocktype)
+    if len(clks) > 1 or len(clks) == 0:
```
rsetaluri, line 179: if we use a generator above, then we should define a function `only`:

```python
class IterableOnlyException(Exception):
    pass

def only(lst: Iterable):
    it = iter(lst)
    try:
        value = next(it)
    except StopIteration:
        raise IterableOnlyException("Expected one element, got []") from None
    try:
        new_value = next(it)
    except StopIteration:
        return value
    else:
        msg = f"Expected one element got {[value, new_value] + list(it)}"
        raise IterableOnlyException(msg)
```

(feel free to rename `IterableOnlyException` -- it's pretty a terrible name) 

and then we can use it as follows:

```python
clks = get_clocks_from_defn(defn, clocktype)
try:
    defnclk = only(clks)
except IterableOnlyException:
    return
```

magma/clock.py
```diff
@@ -188,13 +210,14 @@ def get_reset_args(reset_type: Optional[AbstractReset]):
...
-        clocks[clock_type] = first(
-            get_first_clock(port, clock_type)
-            for port in defn.interface.ports.values()
-        )
-    return clocks
+        clocks = get_clocks_from_defn(defn, clock_type)
+        if len(clocks) == 1:
+            default_clocks[clock_type] = clocks[0]
+        else:
+            default_clocks[clock_type] = None
```
rsetaluri, line 219: From above, again, if we use a generator and this `only` method, this code can become
```suggestion
        clocks = get_clocks_from_defn(defn, clock_type)
        try:
            clk = only(clocks)
        except IterableOnlyException:
            clk = None
        default_clocks[clock_type] = clk
```

magma/backend/coreir/coreir_transformer.py
```diff
@@ -434,16 +415,14 @@ def connect(self, module_defn, port, value):
...
             value = port.trace()
         if value is None:
             if port.is_inout():
                 return  # skip inouts because they might be conn. as an input.
             if getattr(self.defn, "_ignore_undriven_", False):
                 return
-            error_str = f"Found unconnected port: {port.debug_name}\n"
-            error_str += _make_unconnected_error_str(port)
-            raise Exception(error_str)
+            raise UnconnectedPortException(port)
```
rsetaluri, line 425: great change!

magma/compile_exception.py
```diff
@@ -0,0 +1,33 @@
+from magma.array import Array
+from magma.tuple import Tuple
+
+
+class MagmaCompileException(Exception):
```
rsetaluri, line 5: nit: can we move the function to the top, above both classes?

tests/test_syntax/test_sequential2.py
```diff
@@ -170,25 +170,6 @@ def __call__(self) -> m.Bits[8]:
                              f"build/TestSequential2DefaultClock.v",
                              f"gold/TestSequential2DefaultClock.v")
 
-    @m.sequential2()
```
rsetaluri, line 173: why is this removed?

rsetaluri (changes requested):
A few comments. The biggest thing is converting the clock-getter functions to return generators into lists for performance. Even though it's probably marginal in the total runtime, it is purely wasted since we often only grab the first or only one.

magma/clock.py
rsetaluri: btw, if possible lets put this `only` function in some common place, perhaps `magma.common`

tests/test_syntax/test_sequential2.py
leonardt: This test relied on implicitly wiring the first clock when there were multiple possible drivers.  This is the behavior that this change set is seeking to avoid, so we can remove this test.  I think when this was added we were considering hte possibility of just automatically wiring up the first driver as default, but I think the changes added here where we don't wire anything up is preferred.  I don't think anyone is using this feature.

leonardt:
I merged the refactor to avoid the circular dependency before doing the requested changes so I didn't have to deal with managing the changes in both branches, but I think the code should be easier to review now since the location of the important changes has already been reviewed.

rsetaluri (approved):
looks good!

PR #957, opened 2021-05-20
Share inline wires for undriven clocks

leonardt:
The current pattern for using the default clocks in inline verilog is to
refer to an undriven anonymous clock (e.g. `clk = m.Clock()`).  Then,
the compiler will use the default clock driver logic to wire this up.
This is useful for the pattern where you have a helper function for
generating properties, with an optional clock argument (so the user can
pass a specific clock, or by default it uses the default clock).  Since
this helper function doesn't know the context in which it's being
invoked (i.e. the default clock of the definition it's being used in),
it simply uses an undriven anonymous clock that will be automatically
wired up later.

The current inline_verilog logic will generate multiple wires for each
undriven clock because they're different anonymous values, which is
annoying (since many properties use the default clock, we get many wires
for the default clock, but they are all the same clock).

This changes the logic to handle undriven clocks and share a wire for
each undriven clock type, with the intention that they will all use the
same default driver wired up later.  Note that we don't handle general
(non clock types) since this logic happens during the process inline
verilog pass during the compilation phase.  At this point, we assume
that all values have been wired up, so if there's no driver for a
non-clock value, then it's an error (won't be handled by the automatic)
clock wiring logic, so we only need to consider the case where clock
types are undriven.

### DIFF
--- a/tests/test_verilog/gold/test_inline_verilog_share_default_clocks.v
+++ b/tests/test_verilog/gold/test_inline_verilog_share_default_clocks.v
@@ -1,0 +1,51 @@
+module coreir_wrap (
+    input in,
+    output out
+);
+  assign out = in;
+endmodule
+
+module corebit_term (
+    input in
+);
+
+endmodule
+
+module WireClock (
+    input I,
+    output O
+);
+wire Wire_inst0;
+wire coreir_wrapInClock_inst0_out;
+assign Wire_inst0 = coreir_wrapInClock_inst0_out;
+coreir_wrap coreir_wrapInClock_inst0 (
+    .in(I),
+    .out(coreir_wrapInClock_inst0_out)
+);
+coreir_wrap coreir_wrapOutClock_inst0 (
+    .in(Wire_inst0),
+    .out(O)
+);
+endmodule
+
+module Foo (
+    input x,
+    input y,
+    input CLK,
+    input RESET
+);
+wire _magma_inline_wire0_O;
+wire _magma_inline_wire1;
+WireClock _magma_inline_wire0 (
+    .I(CLK),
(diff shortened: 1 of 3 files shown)


tests/test_verilog/gold/test_inline_verilog_share_default_clocks.v
```diff
@@ -0,0 +1,51 @@
...
+WireClock _magma_inline_wire0 (
+    .I(CLK),
+    .O(_magma_inline_wire0_O)
+);
+assign _magma_inline_wire1 = RESET;
+
+assert property (@(posedge _magma_inline_wire0_O) disable iff (! _magma_inline_wire1) x |-> ##1 y);
+
+
+assert property (@(posedge _magma_inline_wire0_O) disable iff (! _magma_inline_wire1) x |-> ##1 y);
```
leonardt, line 48: Notice here that the properties share the inline wires for clk/reset

rsetaluri:
@leonardt is this a performance problem? Or is it a back-end readability problem? I ask because the anonymous clock pattern seems like a nice abstraction; basically we don't have to track anything and rely on the default clock-wiring logic (which also works as is).

rsetaluri (approved):
Seems like a good way to implement this change.

If I understand correctly, previously, for all wires we stored the necessary inline wire by associating a wire for each value. Now, *just for clock types*, we store the wire based on just the type of the value, so every value of that type will have the **same wire**. Is that correct?

leonardt:
> If I understand correctly, previously, for all wires we stored the necessary inline wire by associating a wire for each value. Now, just for clock types, we store the wire based on just the type of the value, so every value of that type will have the same wire. Is that correct?

Yes

PR #959, opened 2021-05-25
Add support for ReadyValid[None]

leonardt:


magma/types/ready_valid.py
```diff
@@ -29,14 +29,25 @@ def _check_T(cls, T):
             )
         return undirected_T
 
+    def _add_data_if_not_none(cls, fields, T, dir):
+        if T is not None:
+            T = cls._check_T(T)
+            fields["data"] = dir(T)
```
rsetaluri, line 35:
```suggestion
    def _maybe_add_data(cls, fields, T, qualifier):
        if T is None:
            return
        T = cls._check_T(T)
        fields["data"] = qualifier(T)
```

and let's make it a free function and then invoke it as `_maybe_add_data(cls, ...)`

rsetaluri: Actually on second thought we can probably have a single free function:

```python
_make_fields(cls, T, qualifier):
    fields = {"valid": Bit, "ready": Bit}
    if T is None:
        return fields
    T = cls._check_T
    fields["data"] = qualifier(T)
    return fields
```

on that note `_check_T` can also be free

(my general principle for free functions is if we don't want them to be public *and* they're "stateless" [i.e. do not modify state of the class/instance] then they can be free.)

magma/types/ready_valid.py
```diff
@@ -47,7 +58,10 @@ class ReadyValid(Product, metaclass=ReadyValidKind):
     * m.Consumer(m.ReadyValid[T])  # data/valid are inputs, ready is output
     * m.Producer(m.ReadyValid[T])  # data/valid are outputs, ready is input
 
-    T is the type of data to be wrapped in Ready/Valid
+    T is the type of data to be wrapped in Ready/Valid.
+
+    To create a ReadyValid type with no data, use T=None.  In this case, there
+    will be no data port
```
rsetaluri, line 64:
```suggestion
    will be no data port.
```

magma/types/ready_valid.py
```diff
@@ -59,19 +73,32 @@ def fired(self):
...
 class ReadyValidProducerKind(ReadyValidKind):
     def __getitem__(cls, T: Union[Type, MagmaProtocol]):
-        T = cls._check_T(T)
-        fields = {"valid": Out(Bit), "data": Out(T), "ready": In(Bit)}
-        return type(f"{cls}[{T}]", (cls, ), fields)
+        fields = {"valid": Out(Bit), "ready": In(Bit)}
+        cls._add_data_if_not_none(fields, T, Out)
+        t = type(f"{cls}[{T}]", (cls, ), fields)
+        if T is None:
+            # remove deq methods if no data
```
rsetaluri, line 90:
```suggestion
            # Remove deq/no_deq methods if no data.
```

magma/types/ready_valid.py
```diff
@@ -122,19 +149,31 @@ def no_deq(self, when=True):
             self.ready @= False
 
 
+def _enq_error(self, value, when=True):
+    raise Exception("Cannot enq with T=None")
```
rsetaluri, line 153: same comment as above re: error msg

magma/types/ready_valid.py
```diff
@@ -29,14 +29,25 @@ def _check_T(cls, T):
...
         return type(f"{cls}[{T}]", (cls, ), fields)
 
     def flip(cls):
         raise TypeError("Cannot flip an undirected ReadyValid type")
 
+    @property
+    def undirected_data_t(cls):
+        if "data" in cls.field_dict.keys():
+            return cls.data.undirected_t
+        return None
```
rsetaluri, line 49: more pythonic
```suggestion
        try:
            data = cls.field_dict["data"]
        except KeyError:
             return None
        return data.undirected_t
```

tests/test_type/test_ready_valid.py
```diff
@@ -179,3 +179,23 @@ class StubQueue(m.Circuit):
         do_deq = io.deq.fired()
 
         io.deq.data @= io.enq.data
+
+
+def test_ready_valid_none():
+    class Foo(m.Circuit):
+        io = m.IO(producer_handshake=m.Producer(m.ReadyValid[None]),
```
rsetaluri, line 186: can check that `undirected_t is None` as well

magma/types/ready_valid.py
leonardt: We can make it a free function, but we can't reuse _make_fields the way you wrote it because the producer/consumer kind use different directions for valid/ready, so we either need to add arguments for those qualifiers, but at that point I think it's about the same amount of code as just passing in the initialized dictionary

PR #971, opened 2021-06-21
Add truncation check for Array constructor

### LINKED ISSUE #969: Magma should raise an error when an operator with an integer requires truncation
Here's an example:
```python
import magma as m


class Foo(m.Circuit):
    io = m.IO(x=m.In(m.Bits[2]), y=m.Out(m.Bits[2]))
    io.y @= io.x & 4


m.compile("build/Foo", Foo, inline=True)
```

This produces

```verilog
module Foo (
    input [1:0] x,
    output [1:0] y
);
assign y = x & 2'h0;
endmodule
```

Rather than implicitly truncating the integer value, we should raise and error and force the user to acknowledge that the value will be truncated by doing the truncation themselves.

leonardt:
Fixes #969


magma/array.py
```diff
@@ -239,6 +239,11 @@ def _make_array_from_bv(N, T, arg):
         raise TypeError(
             f"Cannot construct Array[{N}, {T}] with BitVector of length "
             f"{len(arg)} (sizes must match)")
+    if isinstance(arg, int) and arg.bit_length() > N:
+        raise ValueError(
+            f"Cannot construct Array[{N}, {T}] with integer {arg}"
+            f" (requires truncation)"
+        )
```
rsetaluri, line 246: nit

```suggestion
        raise ValueError(
            f"Cannot construct Array[{N}, {T}] with integer {arg} "
            f"(requires truncation)")
```

magma/conversions.py
```diff
@@ -120,6 +120,11 @@ def convertbits(value, n, totype, checkbit):
     if isinstance(value, IntegerTypes):
         if n is None:
             n = max(value.bit_length(), 1)
+        else:
+            if value.bit_length() > n:
+                raise ValueError(
+                    f"Cannot convert {value} to a {totype} of length {n}"
+                )
```
rsetaluri, line 127: nit

```suggestion
                raise ValueError(
                    f"Cannot convert {value} to a {totype} of length {n}")
```

tests/test_type/test_truncation.py
```diff
@@ -0,0 +1,19 @@
+import magma as m
```
rsetaluri, line 1: nit: let's move these into `test_bits.py`

tests/test_type/test_truncation.py
```diff
@@ -0,0 +1,19 @@
...
+import pytest
+
+
+def test_python_int_truncation_error():
```
rsetaluri, line 5: let's just condense these into a single test and test the issue directly:

```python
def test_python_bits_from_int_truncation_error():
    with pytest.raises(ValueError) as e:
        m.Bits[2](4)
    assert str(e.value) == (
        "Cannot construct Array[2, Bit] with integer 4 (requires truncation)")
    with pytest.raises(ValueError) as e:
        m.bits(4, 2)
    assert str(e.value) == "Cannot convert 4 to a Bits of length 2"
```

rsetaluri: I was going to say it would be nice if there was just a single place we caught this -- in the `Array`/`Bits` constructor, but it seems like the way the `bits()` conversion function works, that won't be feasible. Might be worth considering consolidating these into the type constructors instead of external conversion functions.

rsetaluri (changes requested):

tests/test_type/test_truncation.py
leonardt: Yea I think there is actually the same/similar logic to do conversion inside the constructor, let's open an issue to try to unify where these conversions are done (probably makes the most sense to have it in the type constructor).

rsetaluri (approved):

PR #972, opened 2021-06-24
Add default clock if none is present

### LINKED ISSUE #970: Automatically lift unwired clock ports to interface if a default driver does not exist
A common pattern is to be incrementally defining a module and adding a register.  The user may not have added a ClockIO to the interface, and there are no other possible default drivers coming from instances (e.g. a clock generator).  In this case, it seems useful to automatically add a ClockIO to the interface of the circuit and wire it up the undriven clock inputs.

leonardt:
Fixes #970

There's a common case where the user forgets to specify a clock port
with m.ClockIO.  The requested feature is that in this case, we
automatically lift the clock port to the external interface.

I investigate a few options for doing this.  The first idea was to do it
in the backend using a pass.  The problem with this is at this point the
IO is already bound, and cannot be modified.

The cleanest solution I was able to arrive at was during circuit
construction time.  Before we bind the IO (and it can no longer be
modified), we check whether there's any clock types.  If not, we add the
default clock port, then the standard automatic clock wiring logic will
take things from there.  If the added clock port is unused, (i.e.
there's no instances that require it), we remove the port in the
backend.


leonardt:
A few things need to be resolved before merging:
1) Is there a better way to do this that doesn't involve modifying the IO/interface objects? It seems like we could do a pass in the backend but we will still need to modify the IO/interface objects.  If we do, then this seems like the cleanest approach I could come up with
2) To remove the default clock if it's unused, it seems like the cleanest way to do this is to do a wireclocks pass then remove the port if it's not wired to anything.  This pass was originally avoided and the automatic logic done in the backend to avoid having to traverse the hierarchy for this logic.  This is mainly a performance issue, but it seems hard to do this without a pre-pass (otherwise in the backend we already add the clock port before doing the automatic clock wiring logic).  One thing to note is that this probably going to be the less common case (i.e. most people will use explicit clocks) and we only run the pass if there is a default clock, so it may not be a huge hit for most designs.
3) There's an annoying circular dependency between Bit and Circuit.  I'll have to think about the best way to resolve this but basically we need access to the Clock type (which depends on Bit) in order to do the default clock logic which happens during the Circuit construction pipeline.  However, Bit depends on Circuit for the primitive logic.  We might consider factoring the primitives out of the type definition and "patching" them on.

leonardt:
Ok i've refactored it to only add the port if needed by checking the instances.  To do this, I had to move the Interface constructor logic after the placer logic, but that seems fine.  Now we avoid having to run the wireclocks pass and introducing a possibly unused clock port.  Still need to resolve the circular dependency, it seems like Clock might not need to depend on Bit, but it used for undriven/unused (since these are CoreIR level primitives on Bit) and because Enable is a Bit type (we could factor this out to a separate file).

leonardt:
Ok, the cleanest solution I could come up with is factoring the bit primitives into a separate file and monkey patching the definitions onto the Bit class.  The other option is a delayed import inside the Bit method definitions, or moving Enable (and Clock.unused/undriven) out of clock.py.  The main annoyance seems to be that Enable is considered a ClockType (for ClockIO and default wiring and such) but is a subclass of Bit (since doing logic on enable is so common).  I think one future change we may consider is breaking enable out of clock type and into a more general category such as "ControlType", but I think this would probably need to break user code so I think that's not worth it for this purpose.

rsetaluri (approved):

rsetaluri:
@leonardt I think this is a good pattern. Made some clean up changes so I think it's in a good shape and ready to merge.

This made me think though. It might be worth it to make clocks a more first class citizen in magma. As well as the notion of combinational vs sequential along with that. It might make all this stuff more clear and explicit, and as a byproduct also more performant (e.g. don't have to search each instance for a clock...). And what I mean by first class citizen is specialize circuit types by whether they are combinational or sequential, and also store the clock of a defn and an instance up front. So I could do like `print (MySeqCircuit.clock)`.

leonardt:
that seems like a reasonable pattern, perhaps we should draft a proposal for this? E.g. Circuit2 where we can make some more drastic changes

PR #974, opened 2021-06-30
Error when using undriven input in inline verilog

leonardt:
Adds a more useful error message when trying to reference an undriven
input port in the inline verilog flow.  Before, it would give an error
related to NoneType (because value is None) without telling the user
which signal is problematic.

### DIFF
--- a/magma/inline_verilog.py
+++ b/magma/inline_verilog.py
@@ -67,9 +67,8 @@ def _insert_temporary_wires(cls, value, inline_wire_prefix):
             value = value.value()
             if value is None:
                 raise InlineVerilogError(
-                    f"Found reference to undriven input port:"
-                    f" {orig_value.debug_name}"
-                )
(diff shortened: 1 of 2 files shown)


magma/inline_verilog.py
```diff
@@ -52,14 +52,24 @@ def _make_temporary(defn, value, num, inline_wire_prefix, parent=None):
...
     """
     if isinstance(value, Type):
         if value.is_input():
+            orig_value = value
             value = value.value()
+            if value is None:
+                raise InlineVerilogError(
+                    f"Found reference to undriven input port:"
+                    f" {orig_value.debug_name}"
+                )
```
rsetaluri, line 72: nit
```suggestion
                raise InlineVerilogError(
                    f"Found reference to undriven input port: "
                    f"{orig_value.debug_name}")
```

rsetaluri (approved):

PR #975, opened 2021-06-30
Fix issue in flip then qualify of ready valid

leonardt:
Currently, calling qualify on a ready valid type (e.g. `In(T)`) will
behave correctly in that it will return a type that has all the
subfields (ready, valid, data) as inputs.  However, calling flip on the
qualified type will actually return the Flip of the original type.  This
is because producer/consumer ready valid types define flip to return the
corresponding dual.  The problem here is that when we call qualify using
the default Type.qualify, it returns a Producer/Consumer type with the
correct field types, but doesn't change the flip behavior.

The proposed change is that when you call qualify on a readyvalid type,
you get an AnonProduct back with the qualified fields that behaves like
a normal tuple.  This seems desirable since once you change a readyvalid
type to be qualified as all In or Out, it no longer really behaves like
a standard ReadyValid type and therefor doesn't obey the interfaces
(e.g. enq/deq, however there might be some use case for calling fired on
it but that could be defined as as helper function).

The use case for this pattern is inside the bind/monitor flow which
takes a generic circuit and converts all the ports to inputs (so the
monitor can read the IO).  Calling In on a ReadyValid type works fine,
but calling Flip on it (inside the Circuit interface flow) produces the
wrong behaviors (returns a Producer or Consumer, instead of something
with all inputs/outputs)

### DIFF
--- a/magma/types/ready_valid.py
+++ b/magma/types/ready_valid.py
@@ -47,9 +47,11 @@ class ReadyValidKind(ProductKind):
 
     def qualify(cls, direction):
         T = cls.undirected_data_t
-        fields = {"valid": Bit.qualify(direction),
-                  "ready": Bit.qualify(direction),
-                  "data": T.qualify(direction)}
+        fields = {
+            "valid": Bit.qualify(direction),
+            "ready": Bit.qualify(direction),
+            "data": T.qualify(direction)
+        }
         return AnonProduct[fields]
 
     @property
--- a/magma/types/ready_valid.py
+++ b/magma/types/ready_valid.py
@@ -4,7 +4,7 @@ from magma.logging import root_logger
 from magma.primitives.mux import mux
 from magma.t import Type, In, Out
 from magma.protocol_type import MagmaProtocol
-from magma.tuple import Product, ProductKind
+from magma.tuple import Product, ProductKind, AnonProduct
 
 
 _logger = root_logger()
@@ -44,6 +44,13 @@ class ReadyValidKind(ProductKind):
 
     def flip(cls):
         raise TypeError("Cannot flip an undirected ReadyValid type")
+
+    def qualify(cls, direction):
+        T = cls.undirected_data_t
+        fields = {"valid": Bit.qualify(direction),
+                  "ready": Bit.qualify(direction),
+                  "data": T.qualify(direction)}
+        return AnonProduct[fields]
 
     @property
     def undirected_data_t(cls):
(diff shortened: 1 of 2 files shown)


magma/types/ready_valid.py
```diff
@@ -45,6 +45,13 @@ def __getitem__(cls, T: Union[Type, MagmaProtocol]):
     def flip(cls):
         raise TypeError("Cannot flip an undirected ReadyValid type")
 
+    def qualify(cls, direction):
+        T = cls.undirected_data_t
+        fields = {"valid": Bit.qualify(direction),
+                  "ready": Bit.qualify(direction),
+                  "data": T.qualify(direction)}
```
rsetaluri, line 52: nit
```suggestion
        fields = {
            "valid": Bit.qualify(direction),
            "ready": Bit.qualify(direction),
            "data": T.qualify(direction)
        }
```

magma/types/ready_valid.py
```diff
@@ -45,6 +45,13 @@ def __getitem__(cls, T: Union[Type, MagmaProtocol]):
     def flip(cls):
         raise TypeError("Cannot flip an undirected ReadyValid type")
 
+    def qualify(cls, direction):
```
rsetaluri, line 48: a general question: do these types get cached? for any of the ready-valids

rsetaluri (approved):
Seems like a reasonable pattern. The only issue I can think of is:

```python
T = m.In(m.ReadyValid)
assert isinstance(T, m.ReadyValid)  # fails! I think...
```

I think it's fine in order to unblock, but there might be a more principled solution to this. Would consider opening a sep. issue or adding a comment here.

magma/types/ready_valid.py
leonardt: The Product metaclass handles caching for normal types, we should investigate whether there's any undesired interasctions between Products with the same field names/types as readyvalid (i.e. i think the caching should take the class into account, but it may currently only be tired to field names/types)

leonardt:
I think perhaps a follow up change we can do that should be fine for this use case (because it would only break internal code), is to not use `In(T)` for the monitor logic, and instead use some interface like `Monitor(T)` which returns a view of the type as inputs.  Then, we could say `In(T)` is not valid when `T: ReadyValid` for the reasons mentioned above (e.g. enq/deq don't make sense).  A similar alternative would have `In(T)` return a `ReadyValidMonitor` type which is a special subtype of ReadyValid (like consumer/producer) that only has input views.  This might be a reasonable approach that doesn't require changing the monitor code.

PR #980, opened 2021-07-15
Issue979

### LINKED ISSUE #979: Magma protocol losing type under ite
Bit.ite does not "cast" protocols back to their protocol type

Specifically:
```Python
assert type(proto0) == type(proto1)
v = Bit().ite(proto0, proto1)
assert type(v) == type(proto0)._to_magma_
```
where I would expect:
```Python
assert type(proto0) == type(proto1)
v = Bit().ite(proto0, proto1)
assert type(v) == type(proto0)
```

cdonovick:
closes #979

### DIFF
--- a/magma/bit.py
+++ b/magma/bit.py
@@ -20,7 +20,7 @@ from magma.debug import debug_wire
 from magma.family import get_family
 from magma.interface import IO
 from magma.language_utils import primitive_to_python
-from magma.protocol_type import magma_type
+from magma.protocol_type import magma_type, MagmaProtocol
 from magma.operator_utils import output_only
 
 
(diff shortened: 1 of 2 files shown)


magma/bit.py
```diff
@@ -161,7 +161,13 @@ def ite(self, t_branch, f_branch):
         # Note: coreir flips t/f cases
         # self._mux monkey patched in magma/primitives/mux.py to avoid circular
         # dependency
-        return self._mux([f_branch, t_branch], self)
+        r = self._mux([f_branch, t_branch], self)
+
+        # cast protocol values back to protocol
```
rsetaluri, line 166: wondering if we want to add the conversion function in `Mux` generator (https://github.com/phanrahan/magma/blob/0975d81d192918dcee16194e03b4f1912c547841/magma/primitives/mux.py#L38) or `mux` helper function (https://github.com/phanrahan/magma/blob/0975d81d192918dcee16194e03b4f1912c547841/magma/primitives/mux.py#L126).

@leonardt

cdonovick: doesn't matter to me I was just was looking the fastest way to get a working solution

rsetaluri: ok give me a little bit, i'll fix the mux and it should fix your issue

cdonovick: Thanks, sounds good to me.  As long as it passes my test I don't care where the casting logic is.

leonardt (approved):
LGTM, moving it to the mux seems reasonable and should provided consistent behavior when using the mux operator directly

PR #983, opened 2021-07-26
Improve  handling of In(ReadyValid)

leonardt:
Fixes #975

Avoids the following inconsistency:
```python
T = m.In(m.ReadyValid)
assert isinstance(T, m.ReadyValid)  # fails!
```

by introducing the concept of ReadyValidMonitor (ala ReadyValid
Producer/Consumer)

### DIFF
--- a/magma/types/ready_valid.py
+++ b/magma/types/ready_valid.py
@@ -2,9 +2,9 @@ from typing import Union
 from magma.bit import Bit
 from magma.logging import root_logger
 from magma.primitives.mux import mux
-from magma.t import Type, In, Out
+from magma.t import Type, In, Out, Direction
 from magma.protocol_type import MagmaProtocol
-from magma.tuple import Product, ProductKind, AnonProduct
+from magma.tuple import Product, ProductKind
 
 
 _logger = root_logger()
@@ -46,13 +46,9 @@ class ReadyValidKind(ProductKind):
         raise TypeError("Cannot flip an undirected ReadyValid type")
 
     def qualify(cls, direction):
-        T = cls.undirected_data_t
-        fields = {
-            "valid": Bit.qualify(direction),
-            "ready": Bit.qualify(direction),
-            "data": T.qualify(direction)
-        }
-        return AnonProduct[fields]
+        if direction is not Direction.In:
+            raise TypeError(f"Cannot qualify ReadyValid with {direction}")
+        return Monitor(cls)
 
     @property
     def undirected_data_t(cls):
@@ -80,6 +76,7 @@ class ReadyValid(Product, metaclass=ReadyValidKind):
     The actual semantics of ready/valid are enforced via the use of concrete
     subclasses.
     """
+
     def fired(self):
         """
         Returns a bit that is high when ready and valid are both high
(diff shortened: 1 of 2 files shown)


rsetaluri:
the check should be `issubclass` right?

magma/types/ready_valid.py
```diff
@@ -245,6 +242,25 @@ def no_enq(self, when=True):
             self.data @= 0
 
 
+class ReadyValidMonitorKind(ReadyValidKind):
+    def __getitem__(cls, T: Union[Type, MagmaProtocol]):
+        fields = {"valid": In(Bit), "ready": In(Bit)}
+        _maybe_add_data(cls, fields, T, In)
+        return type(f"{cls}[{T}]", (cls, ), fields)
+
+    def flip(cls):
```
rsetaluri, line 251: should `qualify()` also raise an error here? should that be the case for all of the derived types? e.g. `m.Out(m.In(m.Bit))` is valid but `m.Out(m.In(m.ReadyValid[...]))` should be disallowed?

tests/test_type/test_ready_valid.py
```diff
@@ -209,7 +209,6 @@ def test_in_ready_valid():
     assert T.ready.is_input()
     assert T.valid.is_input()
     assert T.data.is_input()
-    T = T.flip()
-    assert T.ready.is_output()
-    assert T.valid.is_output()
-    assert T.data.is_output()
+    with pytest.raises(TypeError) as e:
```
rsetaluri, line 212: should we check the `issubclass` relationship? `assert issubclass(m.In(m.ReadyValid[T]), m.ReadyValid[T]))`

rsetaluri (approved):

tests/test_type/test_ready_valid.py
leonardt: Turns out this subclass relationship didn't quite hold for the existing types, had to add some base class logic to resolve it.

magma/types/ready_valid.py
leonardt: This inherits the qualify method from ReadyValidKind, which only allows calling In on a readyvalid type.  so this works
```
    T = m.Consumer(m.ReadyValid[m.Bits[8]])
    T = m.In(T)
    T = m.In(T)
```
but this doesn't work
```
    m.Out(T)
```
and raises:

```
    def qualify(cls, direction):
        if direction is not Direction.In:
>           raise TypeError(f"Cannot qualify ReadyValid with {direction}")
E           TypeError: Cannot qualify ReadyValid with Direction.Out

magma/types/ready_valid.py:50: TypeError
```

So right now qualify does raise an error if you use anything other than m.In

magma/types/ready_valid.py
```diff
@@ -96,7 +96,7 @@ class ReadyValidProducerKind(ReadyValidKind):
     def __getitem__(cls, T: Union[Type, MagmaProtocol]):
         fields = {"valid": Out(Bit), "ready": In(Bit)}
         _maybe_add_data(cls, fields, T, Out)
-        t = type(f"{cls}[{T}]", (cls, ), fields)
+        t = type(f"{cls}[{T}]", (cls, cls.__bases__[-1][T],), fields)
```
rsetaluri, line 99: how does hwtypes get this to work with `Array` and `Tuple`?

magma/types/ready_valid.py
leonardt: In general, I think m.Out(m.In(m.Bit)) makes sense because you may want to get a specific version of a generic type (e.g. take any directed type and turn it into an Out).  However, for readyvalid, having an Out of readyvalid type doesn't really make sense, so we override the default behavior to raise an error because in this case the semantics are not well formed (we can monitor a ReadyValid as a In, but it doesn't make sense really to drive both the In and the Out port, well maybe it would for some mock driver for testing, but unless this use case pops up I say we don't support it for now).

magma/types/ready_valid.py
leonardt: With some pretty complicated logic, here's the magma implementation: https://github.com/phanrahan/magma/blob/master/magma/tuple.py#L47-L71 which is based off the generic hwtypes implementation

rsetaluri: oh boy, ok. fine with this, can robust-ify it down the line if it becomes an issue.

PR #987, opened 2021-08-02
Resolve regression with ReadyValid.undirected_t

leonardt:
This is used in the product/tuple base class logic (see
https://github.com/leonardt/magma_riscv_mini/runs/3216036083?check_suite_focus=true
for a failure example).  I think it is reasonable to be able to get the
"undirected" version of a ReadyValid and the recent change to the
monitor logic removed this capability.  These changes add back the
capability while preserving the prevention of using the In/Out qualifier.

### DIFF
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -96,8 +96,7 @@ from magma.primitives import (LUT, Mux, mux, dict_lookup, list_lookup,
                               Memory, set_index, register, Wire)
 
 from magma.types import (BitPattern, Valid, ReadyValid, Consumer, Producer,
-                         Decoupled, EnqIO, DeqIO, Irrevocable)
-from magma.types.ready_valid import Monitor
+                         Decoupled, EnqIO, DeqIO, Irrevocable, Monitor)
 import magma.smart
 from magma.compile_guard import compile_guard
 from magma.set_name import set_name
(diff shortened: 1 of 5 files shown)


rsetaluri (commented):
This looks fine, but I've lost track a bit about what all combinations are possible. It would be ideal to have both complete documentation and full test coverage on what combinations are valid (and what their semantics are), and what combinations are not. I think we already have some ready valid stuff in the docs (https://github.com/phanrahan/magma/blob/master/docs/types_and_values.md). If we can fill this out to include all the recent changes that would be great.

Re: tests, even just a blanket parameterized test that checks the different allowed/disallowed combinations would be great. Paraphrasing heavily but something like

```python
@pytest.parametrize("qualifiers,valid", [((m.Consumer,m.ReadyValid),True), ...])
def test_qualifiers(qualifiers, valid):
    T = m.Bit
    if valid:
        for q in reversed(qualifiers):
            T = q(T)
        return
    with pytest.raises(TypeError) as e:
            for q in reversed(qualifiers):
                T = q(T)
    assert str(e) = ...
    ```

**Comment by leonardt** [comment]
I will add docs+tests that cover the various allowed directions

File: magma/__init__.py [hunk 1]
```diff
@@ -97,6 +97,7 @@ def set_mantle_target(t):
 
 from magma.types import (BitPattern, Valid, ReadyValid, Consumer, Producer,
                          Decoupled, EnqIO, DeqIO, Irrevocable)
+from magma.types.ready_valid import Monitor
```
rsetaluri, line 100: just include this in `magma/types/__init__.py` and then merge with line above

rsetaluri (approved):
great!

PR #988, opened 2021-08-11
Fix bug in inline_verilog reference of tuple slice

leonardt:
Before this change, magma assumes that a DefnRef will not be inlined
(since ports aren't inlined), so skips emitting a blacklisted temporary
wire.  The problem is that CoreIR's inline passthrough logic sometimes
inserts temporary wires that do get inlined.  Then, in the compiler, the
inline verilog code generation emits a reference to the temporary wire,
which is then inlined without updating the inline verilog code since
it's not part of the AST.

This change avoids the issue by always inserting a blacklisted temporary
wire regardless of what kind of reference there is.  This is actually
seems like a reasonable choice becuase we cannot assume what might
happen downstream in the compiler (i.e. compiler passes should be free
to insert temporary wires where needed, that are then inlined out).
With this change, magma is impervious to these transformations by always
requiring the reference in the inline verilog to be preserved.

Most of these changes are gold files that requried updating with the
introduction of the new inline wires.

### DIFF
--- a/magma/inline_verilog.py
+++ b/magma/inline_verilog.py
@@ -50,10 +50,10 @@ class InlineVerilogError(RuntimeError):
 
 def _insert_temporary_wires(cls, value, inline_wire_prefix):
     """
-    Insert a temporary Wire instance so the signal isn't inlined out
+    Insert a temporary Wire instance so the signal isn't inlined out.
 
     We have to do this for DefnRef because the coreir inline.cpp logic
-    sometimes inserts temporary wires for DefnRef that eventually get inlined
+    sometimes inserts temporary wires for DefnRef that eventually get inlined.
     """
     if isinstance(value, Type):
         if value.is_input():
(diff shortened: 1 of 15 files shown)


magma/inline_verilog.py
```diff
@@ -58,8 +50,10 @@ class InlineVerilogError(RuntimeError):
 
 def _insert_temporary_wires(cls, value, inline_wire_prefix):
     """
-    For non DefnRef, insert a temporary Wire instance so the signal isn't
-    inlined out
+    Insert a temporary Wire instance so the signal isn't inlined out
+
+    We have to do this for DefnRef because the coreir inline.cpp logic
+    sometimes inserts temporary wires for DefnRef that eventually get inlined
```
rsetaluri, line 56: nit
```suggestion
    Insert a temporary Wire instance so the signal isn't inlined out.

    We have to do this for DefnRef because the coreir inline.cpp logic
    sometimes inserts temporary wires for DefnRef that eventually get inlined.
```

rsetaluri (approved):

Issue #977, opened 2021-07-09
Add logging/warning statements for auto clock wiring


rsetaluri:
Log the following:

- *INFO*: whenever we do an auto clock wiring
- *WARNING*: whenever we do *not* do auto clock wiring because we find multiple clocks

rsetaluri:
Closed by #989

PR #989, opened 2021-08-11
[#977] Add logging to clock wiring pass

rsetaluri:
* Adds logging to clock wiring pass
  * `INFO` for succesful auto-wiring clock
  * `WARNING` for no clocks found/multiple clocks found.
    * NOTE: this warning will be raised even if there are no underlying clocks that need to be driven. We should look into making this warning contingent on that; otherwise, all combinational code will result in warnings.
    * [EDIT]: This has since been fixed!
* Refactors, renames, and moves the clock wiring code


rsetaluri:
See #977

magma/passes/clock.py
```diff
@@ -1,10 +1,201 @@
...
+    try:
+        first = next(it)
+    except StopIteration:  # early exit to avoid traversing other elements
+        return
+    yield first
+    yield from it
+    # NOTE(leonardt): Since magma arrays don't support zero-length values, we
+    # have to check the size before slicing.
+    if len(value) <= 1:
+        return
```
leonardt, line 104: Hmm, minor comment but if it's to be believed, then maybe we should change it to `len(value) == 1` and put an assertion that `len(value) >= 1` (or update the comment if it's possible to have a length 0 array somewhere.

magma/passes/clock.py
```diff
@@ -1,10 +1,201 @@
...
+    except EmptyIterableException:
+        _logger.warning(
+            f"Found no clocks in {defn.name}; skipping auto-wiring "
+            f"{clock_type}")
+        return
+    except NonSignletonIterableException:
+        _logger.warning(
+            f"Found multiple clocks in {defn.name}; skipping auto-wiring "
+            f"{clock_type}")
+        return
```
leonardt, line 160: Could this be done at the top level to avoid having to traverse the instances?  (i.e. check once if there's a default clock, otherwise skip)  Also we shouldn't need to find these for every instance (instead reuse them across all instances)

leonardt: We could run a pass that does get_output_clocks_in_defn and stores it in the defn object

leonardt (approved):
LGTM, just one comment on avoiding having to traverse the instance ports when there's no default clock, as well as avoiding finding the default drivers more than once.

magma/passes/clock.py
leonardt: I guess there's a small tradeoff with the common case where clocks are manually wired, so we don't always need to store this information.  Another option is to run this logic once and cache it for future needs (so it's only run on demand when we encounter an undriven clock, but not run more than once)

magma/passes/clock.py
rsetaluri: I think this comment is more about that if you do `value[1:]` in python it would just return an empty list, but with a magma array it will result in an error regarding 0-length arrays. Therefore we have to do an early exit check rather than just having a zero-length array be returned.

magma/passes/clock.py
rsetaluri: Yeah that all makes sense. The reason I did it in this order was for diagnostics. If there are no undriven clocks in the defn then it doesn't matter whether or not there is a default clock. So we need to do that check first before checking for default clocks.

We could first check for default clocks, and then check for undriven clocks but that makes the code pretty ugly and results in some duplicated code -- but I do agree this is a perf. tradeoff.

My longer term plan is to make clocks more of a first class citizen. That is, when a definition is created we stash some information about it's clocks explicitly and then have a well-defined API to query about those clocks. That will also be useful when thinking about multiple clock domains etc. So as long as we don't observe a performance regression with this code, my plan is to forego more work on this until we make this change.

Issue #967, opened 2021-06-08
Forgetting parens (instance) when using Generator2 provides a confusing error


leonardt:
Here's an example:
```python
import magma as m


class Foo(m.Generator2):
    def __init__(self, width):
        self.io = m.IO(x=m.In(m.Bits[width]), y=m.Out(m.Bits[width]))
        self.io.y @= self.io.x


class Bar(m.Generator2):
    def __init__(self, width):
        self.io = m.IO(a=m.In(m.Bits[width]),
                       z=m.Out(m.Bit))
        self.io.z @= True
        # Missing a paren
        foo = Foo(width)
        # This wires up an invalid value to Foo.y
        m.wire(foo.y, self.io.a)
        # Correct
        foo = Foo(width)()
        m.wire(foo.x, self.io.a)


m.compile("build/Bar", Bar(4))
```

produces

```
base ❯ python a.py
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[0], Old Output: Foo.x[0], New Output: LazyCircui
t.a[0]
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[1], Old Output: Foo.x[1], New Output: LazyCircui
t.a[1]
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[2], Old Output: Foo.x[2], New Output: LazyCircui
t.a[2]
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[3], Old Output: Foo.x[3], New Output: LazyCircui
t.a[3]
build/Bar.json

/tmp
base ❯ python a.py
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[0], Old Output: Foo.x[0], New Output: LazyCircui
t.a[0]
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[1], Old Output: Foo.x[1], New Output: LazyCircui
t.a[1]
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[2], Old Output: Foo.x[2], New Output: LazyCircui
t.a[2]
WARNING:magma:Wiring multiple outputs to same wire, using last conne
ction. Input: Foo.y[3], Old Output: Foo.x[3], New Output: LazyCircui
t.a[3]
ERROR: Cannot select a From self
[...]

PR #990, opened 2021-08-19
Fix 967

leonardt:
WIP fix for #967 

Adds logic for tracking the associated inst/defn for values and verifying that values being wired are from a valid context (e.g. instances in the same definition, see the original issue for more a thorough list)

Need to do:
- [x] Tmp value handling
- [x] ~~Handling laziness (right now it assumes LazyDefnRef/LazyCircuit corresponds to the current active definition, is this okay?)~~ not an issue when done at compile time
- [x] Handling portviews -- these actually break the rules because they allow for hierarchical references, we could check that the portview reference is from a valid place in the hierarchy
- [x] Finish adding error messages for various cases

Adding these wiring rules actually revealed some random bugs in test suite (e.g. instancing and wiring things outside of definitions), which required some minor changes to the test code.  

We also needed to add cls.open() in places where we modify wiring so the checks can verify that the two values are being wired in the proper context (i.e. the top definition).

There was some nasty changes related to logging that I still need to review/refactor.  The issue was that in the existing logic, unstaging a logger unsets a flag to stage, but this is actually undesired in the case of nested definitions (e.g. when we call a generator from inside a circuit definition, unstaging the logger shouldn't prevent the subsequent code in the calling definition from being staged, further more, ti should only unstage the logs for it's local definition, and not touch the logs for the enclosing definition).  I changed the staged logging logic to use a stack based flow to handle this, but this reveal some inconsistencies in the pipeline where things where staged/unstaged (e.g. some of tests leave some circuit logs staged, meaning that not every "stage_logger" is followed by an "unstage_logger".).  It's probably worth trying to list out all the places where "stage_logger" is called and where "unstage_logger" is called.   I think a major portion of the confusion/complexity arises from the various ways in which we can use a circuit definition (i.e. some methods follow different code paths, resulting in different combinations of staged/unstage being called).  Finally, there's the issue of having an exception thrown inside a definition.  In this case, unstage won't be called, but I'm not sure if we can catch this exception in the new style class logic because the class code is called implicitly (we need some hook such that, if an exception is thrown in the class body, unstage the logs for the circuit).  

Opening a draft now in case you have any suggestions related to the staged logging and/or the remaining tasks (e.g. handling laziness)


leonardt:
Also, I think it's worth documenting the wiring rules written in the original issue in the wiring documentation for users.

leonardt:
Figured out that we can handle portviews by tracing to the "root" instance used to create the reference, then doing the same "instance" context check for normal instances.

magma/backend/coreir/coreir_transformer.py
```diff
@@ -37,6 +39,45 @@
 _generator_callbacks = {}
 
 
+def _check_wiring_context(i, o):
```
rsetaluri, line 42: I think ideally this should be called during a checking pass, independent of CoreIR, i.e. before lowering to CoreIR. But I can see performance implications for that being a separate pass and having to iterate over all the connections multiple times, so need to think about it for a bit.

At least for now, can we move this function into a separate file? Maybe `magma/backend/check_wiring_context.py: check_wiring_context`.

magma/t.py
```diff
@@ -134,6 +134,28 @@ def is_driven_anon_temporary(self):
...
+        otherwise None
+        """
+        if not isinstance(self.name.root(), DefnRef):
+            return None
+        return self.name.root().defn
+
+    def temp(self):
+        return (type(self.name.root()) is NamedRef or 
+                type(self.name.root()) is AnonRef)
+
```
rsetaluri, line 158: can we make these all free functions with names `get_ref_inst`, `get_ref_defn`, `ref_is_inst` and move them to `ref.py`? Then wherever we call `x.fn()` we replace with `ref.fn(x.name)`

tests/test_wire/test_967.py
```diff
@@ -0,0 +1,90 @@
+import pytest
```
rsetaluri, line 1: can we move this file to `tests/test_wire/test_check_wiring_context.py`

magma/backend/coreir/coreir_transformer.py
leonardt: To make it modular while avoiding the performance implications, we could provide it somewhere in the class ancestry as part of the super logic.  Maybe as part of an API to "get the driver" of a value (since we need the source to wire up the value/sink).  Then as part of the API, we run the error check and this can be reused by backends.

magma/backend/coreir/coreir_utils.py
```diff
@@ -8,6 +8,7 @@
                        PortViewRef)
 from magma.tuple import Tuple
 from magma.protocol_type import magma_type, magma_value
+from magma.protocol_type import MagmaProtocol, MagmaProtocolMeta
```
rsetaluri, line 11: seems like an extraneous change?

magma/backend/coreir/coreir_utils.py
```diff
@@ -187,3 +188,11 @@ def __init__(self, value, low, high):
     def get_coreir_select(self,):
         return (magma_name_to_coreir_select(self.value.name) +
                 f".{self.low}:{self.high}")
+
+
+def _unwrap(x):
```
rsetaluri, line 193: same; seems like an extraneous change?

magma/circuit.py
```diff
@@ -216,11 +215,11 @@ def _get_interface_type(cls, add_default_clock):
 
 
 def _setup_interface(cls):
-    # Called from DefinitionContext.finalize.
     IO = _get_interface_type(cls, add_default_clock=True)
-    if IO is None:
+    if not hasattr(cls, "IO") and not hasattr(cls, "io"):
```
rsetaluri, line 219: why this check instead of `IO is None`? seems like duplicated logic from `_get_interface_type`.

magma/circuit.py
```diff
@@ -232,12 +231,11 @@ def _add_intermediate_value(value):
...
     """
     root = value.name.root()
-    # root is either None, or a NamedRef. If it is a NamedRef which refers to a
-    # constant, then we can proceed with the empty set. Otherwise, we seed the
-    # set with the root value.
-    if root is None or root.value().const():
-        return OrderedIdentitySet()
-    return OrderedIdentitySet((root.value(),))
+    # If it is a NamedRef which doesn't refer to a constant, then we seed the
+    # set with the root value. Otherwise, we can proceed with the empty set
```
rsetaluri, line 235: nit
```suggestion
    # If it is a NamedRef which doesn't refer to a constant, then we seed the
    # set with the root value. Otherwise, we can proceed with the empty set.
```

magma/circuit.py
```diff
@@ -232,12 +231,11 @@ def _add_intermediate_value(value):
...
+    if type(root) is NamedRef and not root.value().const():
```
rsetaluri, line 236: should this be `isinstance(root, NamedRef)`? this seems brittle.

If it's this way because `InstRef` and `DefnRef` are subclasses of `NamedRef` but we don't want to include them, then there should be a more well-defined way of dealing with these.

I think this is fine for now, but we should in the future clean this up. Just dropping this here for now to document: every instance of `Ref` should have 3 attributes: named, bound, derived. They should be assigned as follows:
* Everything besides AnonRef has named=True; AnonRef has named=False.
* (Lazy)InstRef and (Lazy)DefnRef should all have bound=True. Their "derived" refs (e.g. array slices or tuple getitems) should also have bound=True.
* Only ArrayRef and TupleRef should have derived=True. All others should have derived=False.

I think this covers all our cases and avoids a lot of the ad-hoc logic floating around regarding refs!

(Again, fine to leave as is for now. Probably will this roll this into the backend change).

magma/circuit.py
```diff
@@ -321,6 +319,8 @@ def __new__(metacls, name, bases, dct):
             # Override staged context with '_context_' from namespace if
             # available.
             cls._context_ = dct.get("_context_", context)
+            cls._context_.place_instances(cls)
```
rsetaluri, line 322: is there a reason to split these? It seems best to have everything encapsulated in finalize.

magma/ref.py
```diff
@@ -213,8 +199,33 @@ def qualifiedname(self, sep="."):
...
+
+
+def get_ref_inst(ref):
+    """
+    If value is part of a port on an instance, return that instance,
+    otherwise None
+    """
+    if not isinstance(ref.root(), InstRef):
+        return None
+    return ref.root().inst
```
rsetaluri, line 216: can cache `ref.root()` so we don't compute it twice

```suggestion
    """
    If value is part of a port on an instance, return that instance,
    otherwise None.
    """
    root = ref.root()
    if not isinstance(root, InstRef):
        return None
    return root.inst
```

magma/ref.py
```diff
@@ -213,8 +199,33 @@ def qualifiedname(self, sep="."):
...
+
+
+def get_ref_defn(ref):
+    """
+    If value is part of a port on an definition, return that definition,
+    otherwise None
+    """
+    if not isinstance(ref.root(), DefnRef):
+        return None
+    return ref.root().defn
```
rsetaluri, line 226: same
```suggestion
def get_ref_defn(ref):
    """
    If value is part of a port on an definition, return that definition,
    otherwise None.
    """
    root = ref.root()
    if not isinstance(root, DefnRef):
        return None
    return root.defn
```

magma/ref.py
```diff
@@ -213,8 +199,33 @@ def qualifiedname(self, sep="."):
...
+
+
+def is_temp_ref(ref):
+    return (type(ref.root()) is NamedRef or
+            type(ref.root()) is AnonRef)
```
rsetaluri, line 231: same
```suggestion
def is_temp_ref(ref):
    root = ref.root()
    return (type(root) is NamedRef or type(root) is AnonRef)
```

magma/tuple.py
```diff
@@ -8,6 +8,8 @@
     Product,
     ProductMeta,
 )
+from hwtypes import BitVector, Bit
+
```
rsetaluri, line 12: nit: unnec. new line

tests/test_deprecated/test_old_io_syntax/test_old_io_syntax_circuit_generator.py
```diff
@@ -68,6 +68,7 @@ def definition(add):
 def test_add_generator():
     with pytest.warns(DeprecationWarning):
         Add8cin = DefineAdd(8, has_cin=True, has_cout=False)
+        print(repr(Add8cin))
```
rsetaluri, line 71: spurrious?

magma/circuit.py
leonardt: In line 206, we return None because cl.IO is already an InterfaceKind.  This happens when you do a subclass of a Circuit that already has an IO defined (so _get_interface_type was already run in the parent class).  However, we still want lines 223/224 to run for this subclass (we need to create a new `interface` attribute with `defn=cls`, otherwise we inherit the parent interface attribute with `defn=parent_cls`, we also need to update the port attribute or else the refer to the parent class).

magma/circuit.py
leonardt: > If it's this way because InstRef and DefnRef are subclasses of NamedRef but we don't want to include them, then there should be a more well-defined way of dealing with these
This is why

I think we can add a `TempRef` that is a `NamedRef` subclass, but not in the tree of Inst/Defn Ref to resolve this.

magma/circuit.py
leonardt: In the old-style syntax, we need to call setup_interface before running the definition, then we place the instances afterwards.  In the new-style syntax, we want to call setup_interface after placing instances so we can add the automatically lifted clock if necessary.  For the old-style syntax, we need to call _setup_interface before the definition because otherwise the definition won't have the write ports to wire up.  For the new-style syntax, we resolve this with the staged placer logic which handles everything after definition.

magma/circuit.py
rsetaluri: got it. think this logic can prob be cleaned up but looks correct.

magma/circuit.py
rsetaluri: Makes sense. Hoping we can completely clean out old-style stuff in v2, so it might simplify it then.

rsetaluri (approved):

PR #991, opened 2021-08-25
Add module linking

rsetaluri:
Depends on branch: https://github.com/leonardt/pycoreir/tree/add-link-definitions-api, PR: https://github.com/leonardt/pycoreir/pull/157

### DIFF
--- a/magma/linking.py
+++ b/magma/linking.py
@@ -1,29 +1,66 @@
-from typing import Mapping, Optional
+from typing import Iterable, Mapping, Optional
 
 from magma.circuit import CircuitKind
+
+
+class _LinkInfo:
+    def __init__(self):
+        self.default_linked_module = None
+        self.linked_modules = {}
 
 
 _link_map = {}
 
 
-def link_module(source: CircuitKind, key: str, target: CircuitKind):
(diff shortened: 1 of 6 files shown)


magma/linking.py
```diff
@@ -0,0 +1,70 @@
...
+
+def get_linked_modules(source: CircuitKind) -> Mapping[str, CircuitKind]:
+    info = _get_or_create_link_info(source)
+    return info.linked_modules
+
+
+def link_default_module(source: CircuitKind, target: CircuitKind):
+    # TODO(rsetaluri): Catch errors early here?
+    info = _get_or_create_link_info(source)
+    info.default_linked_module = target
```
leonardt, line 34: Maybe we allow "overriding" the default linked module?  We could issue a warning by default and allow the user to pass a flag to suppress it ("override_default=True")

leonardt (approved):

magma/linking.py
rsetaluri: My only concern is that the rules basically will get duplicated on the front-end and backend. (In terms of checking default, checking linking against the same key, etc.). I think that kind of interface^ is reasonable, but I'd rather delay more logic in the front-end until this becomes more of a hardened feature -- especially if we're migrating backends.

PR #992, opened 2021-09-21
Add is_producer/is_consumer helper methods

leonardt:


magma/types/ready_valid.py
```diff
@@ -85,6 +85,14 @@ def fired(self):
         """
         raise NotImplementedError
 
+    @classmethod
```
rsetaluri, line 88: I think these are better off as free functions. We could then dispatch on both types and kinds (i.e. `is_producer(T()) \equiv is_producer(T)`)

rsetaluri (approved):
Good minus making the functions free.

PR #994, opened 2021-10-01
Compile guard drive output

leonardt:


rsetaluri (approved):

rsetaluri:
https://github.com/phanrahan/magma/blob/5de7002ea2426cedf2b1c813d9e68918b65ce66e/magma/compile_guard.py#L190-L196

this is a great example as for why these should be primitives, and more-over not emitting verilog directly. Hopefully with MLIR this becomes much cleaner.

PR #996, opened 2021-11-12
Mlir fixes

rsetaluri:


leonardt (approved):

leonardt:
I'm guessing these will get covered by the MLIR backend tests, but since they're small innocuous changes I think it's okay to merge now so we don't have the code base get out of sync

PR #999, opened 2021-12-17
Compile guard temp value support

rsetaluri:


leonardt (approved):
Looks good, but it seems this only works if the temporary value is driven when it's used inside the compile guard?

so something like this wouldn't work?
```
temp = m.Bit()
with m.compile_guard(...):
    ...
    reg.I @= temp

# temp wired up after
temp @= io.I
```

Regardless, I think it's still useful, but we should add a note in the compile guard documentation about the supported/unsupported patterns.

In the future, it seems like we could handle the general case by resolving the externality as a backend pass.

PR #1002, opened 2022-01-27
Merge mlir backend

rsetaluri:
Merged from https://github.com/rsetaluri/magma-mlir-experiments `main`, 6c64f0192764850104bd18ce1a0e9f9e7edfd6d4.


magma/backend/mlir/sv.py
```diff
@@ -0,0 +1,253 @@
...
+
+@dataclasses.dataclass
+class BPAssignOp(MlirOp):
+    operands: List[MlirValue]
+
+    def print_op(self, printer: PrinterBase):
+        printer.print(f"sv.bpassign ")
+        print_names(self.operands, printer)
+        printer.print(" : ")
+        print_types(self.operands[1], printer)
```
leonardt, line 70: Seems like the above ops could share code and define an "op_str" field

rsetaluri:
Thanks for the comments. I think I'm pretty satisfied with the state of it for a first cut. I definitely agree a lot of the visitor and printing code paths could be made cleaner and much more modular. If you're ok with it, I'd prefer merging and doing a re-org once we have some confidence in the functionality.

PR #1004, opened 2022-02-10
Fix bug in mixed direction insert wrap casts

leonardt:
Array2 added the early return for when the port isn't driven, however in
the case of a mixed direction tuple, driven will be false but we still
need to recurse and check the inputs in the children.

### DIFF
--- a/magma/backend/coreir/insert_wrap_casts.py
+++ b/magma/backend/coreir/insert_wrap_casts.py
@@ -27,15 +27,20 @@ class InsertWrapCasts(DefinitionPass):
                 value_store.set_value(self.out, input_val)
         return Wrap
 
+    def _recurse(self, port, definition):
+        wrapped = False
+        for t in port:
+            wrapped |= self.wrap_if_named_type(t, definition)
+        return wrapped
+
     def wrap_if_named_type(self, port, definition):
+        if port.is_mixed():
+            return self._recurse(port, definition)
         if not port.driven():
             return False
         value = port.value()
         if isinstance(port, Tuple):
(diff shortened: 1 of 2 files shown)


magma/backend/coreir/insert_wrap_casts.py
```diff
@@ -27,15 +27,20 @@ def simulate(self, value_store, state_store):
                 value_store.set_value(self.out, input_val)
         return Wrap
 
+    def _recurse(self, port, definition):
+        wrapped = False
```
rsetaluri, line 31: `return any(self.wrap_if_named_type(t, definition) for t in port)`

rsetaluri (approved):

magma/backend/coreir/insert_wrap_casts.py
leonardt: We actually don't want it to short circuit because we want it to wrap all children that have named types. The wrapped bool logic was for early exiting in arrays (since if one child wasn't wrapped we didn't need to traverse the rest)

rsetaluri: we could do

```python
return any([self.wrap_if_named_type(t, definition) for t in port])
```

or even

```
# NOTE(leonardt): We can't short circuit, since wrap_if_named_type needs to be called on all elements, so we materialize them into a list rather than a generator.
t_wrapped = [self.wrap_if_named_type(t, definition) for t in port]
return any(t_wrapped)
```

might be more readable but nbd either way, leaving it up to you

PR #1012, opened 2022-03-01
Fix return type for Array wireable functions

leonardt:
Before, the wireable functions like `value` always returned `Array[N,
T.flip()]`, however in the case of a subclass it's preferable (and
expected by previous behavior) to return the subclass type rather than
the base class `Array` type (i.e. `value` on a `Bits` should return a
`Bits` not an array)

### DIFF
--- a/tests/test_type/test_array2.py
+++ b/tests/test_type/test_array2.py
@@ -393,3 +393,16 @@ def test_array2_wire_to_anon():
         for i, driving in enumerate(io.I.driving()):
             assert len(driving) == 1
             assert driving[0] is io.O[i]
+
+
+def test_array2_value_return_type():
+    class Foo(m.Circuit):
+        io = m.IO(I=m.In(m.Bits[2]))
+
+    class Bar(m.Circuit):
+        io = m.IO(I=m.In(m.Bit))
+        foo = Foo()
+        foo.I[0] @= io.I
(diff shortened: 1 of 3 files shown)


tests/test_type/test_array2.py
```diff
@@ -393,3 +393,16 @@ class Foo(m.Circuit):
         for i, driving in enumerate(io.I.driving()):
             assert len(driving) == 1
             assert driving[0] is io.O[i]
+
+
+def test_array2_value_return_type():
```
rsetaluri, line 398: can we move this test to somewhere else and make it more specific to the issue at hand? maybe `tests/test_wire/test_introspection.py`.

And test can be as simple as

```python
@pytest.mark.parameterize("T", (m.Bits,))
def test_value_type(T):
    x = m.Out(T(name="x"))
    y = m.In(T(name="y"))
    y @= x
    assert isinstance(y.value(), T)
```

rsetaluri (approved):
looks good modulo comment re: test

tests/test_type/test_array2.py
leonardt: Moved, but the test can't be quite that simple.  The issue above only arises when array isn't wired as whole, so it needs to test an array that's had elaborated children.

PR #1013, opened 2022-03-02
Support m.Flip(<IO>)

leonardt:


magma/interface.py
```diff
@@ -491,6 +491,10 @@ def __getattr__(self, key):
             return self._ports[key]
         return super().__getattribute__(key)
 
+    def flip(self):
```
rsetaluri, line 494: I think we should also make `flip(self) -> 'IOInterface'` a default method on `IOInterface` base class that raises `NotImplenetedError`. This way the derived classes don't implement them..

rsetaluri:
@leonardt actually have second thoughts on this, pasted here from slack convo:

> wait actually thinking about it…I think it’s not preferred to operations on `IO` directly since (a) it is staged, and (b) it is an internal object to the circuit (we never do `Foo.io`)
> 
> where as interface is so, `Foo.interface` is something we look at, and it is not staged. So I think I prefer flip to be defined on `Interface` not `IO`.
> 
> also there is the problem that we play some weird games with the derived classes of `IO` (e.g. `SingletonInstanceIO`)

leonardt:
My two main concerns with `Flip(foo.interface)`.

One, `interface` is a "magic" attribute provided by magma, whereas a user would naturally know about `Foo.io` since it's a "pythonic" accessible attribute in the code they right.  This requires them to hold more implicit context in their head.

Two, `interface` objects hold the "definition context" view of the ports, so technically the types are already "flipped" w.r.t. to the declaration.  This could be confusing both to the user and maintainers.

rsetaluri:
Yeah this is a good point. It seems like according to our offline conversation that we want to do like:

```python
def gen_io(...):
    return m.IO(...)

class A(m.Circuit):
    io = gen_io(...)

class B(m.Circuit):
    io = gen_io(...).flip()
```

not sure there's a better way to do this than to implement flip on IO. I think if we only implement it on the "normal" IO class (literally `IO` and none of its children), then I would be ok. It's still a little unsavory that when you call IO() you instantiate all the values (ports), and when you call flip you drop it and create a new IO from just the spec, but I can live with the weirdness if it's providing value.

tests/test_io/test_io_flip.py
```diff
@@ -0,0 +1,12 @@
+import magma as m
```
rsetaluri, line 1: let's just name this file `test_io.py` since we don't have a basic test for `IO()` class. Eventually we can add more testing for the IO and derived classes. (not really sure what the other test files in this directory are doing, but probably can clean them up and move them later).

magma/interface.py
```diff
@@ -491,6 +494,12 @@ def __getattr__(self, key):
...
```
rsetaluri, line 497: let's add a utility function called fields which is basically like

```python
def fields(self) -> Dict[str, Kind]:
    return {name: T for name, T in zip(self._decl[::2], self._decl[1::2])}
```

this will probably be useful to be public for users as well

magma/interface.py
```diff
@@ -491,6 +494,12 @@ def __getattr__(self, key):
...
+        return IO(**{
+            name: T.flip()
+            for name, T in zip(self._decl[::2], self._decl[1::2])
+        })
```
rsetaluri, line 501: if we add the function `fields`

```suggestion
        return IO(**{name: T.flip() for name, T in self.fields().items()})
```

tests/test_io/test_io_flip.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_io_flip():
+    class Foo(m.Circuit):
+        io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit))
+
+
+    class Bar(m.Circuit):
+        io = m.Flip(Foo.io)
+        assert io.I.is_input()
+        assert io.O.is_output()
```
rsetaluri, line 12: don't believe we need a circuit here.

```suggestion
def _assert_io_fields(io: m.IO, fields: Dict[str, m.Kind]):
    """Asserts that @io has ordered fields in @fields."""
    for ((name0, T0), (name1, T1)) in zip(io.fields().items(), fields.items()):
        assert name0 == name1
        assert T0 is T1


def test_io_flip():
    A = m.Product.from_fields("anon", dict(x=m.In(m.Bit), y=m.Out(m.Bit))
    B = m.In(m.Bits[8])
    io = m.IO(a=A, b=B)
    flipped = io.flip()
    # Check flipped io.
    _assert_io_fields(flipped, {"a": A.flip(), "b": B.flip()})
    # Check that flip(flip) = id.
    _assert_io_fields(flipped.flip(), io.fields())
```

rsetaluri (changes requested):

magma/interface.py
```diff
@@ -133,11 +133,14 @@ def _make_wires(value, wired):
     return s
 
 
+def _dict_from_decl(decl):
```
rsetaluri, line 136: btw, I think we can get rid of OrderedDict now because in > python3.7 builtin dict is ordered.

rsetaluri (approved):

PR #1014, opened 2022-03-02
Pyverilog on demand

leonardt:


rsetaluri (approved):
I like this pattern a bit better, it's more clear

rsetaluri:
@leonardt let's squash and merge whenever we do

PR #1015, opened 2022-03-02
Add support for connection_iter in magma protocol

leonardt:


tests/test_type/test_magma_protocol.py
```diff
@@ -35,3 +35,14 @@ def __call__(self, cond: m.Bit, i0: T, i1: T) -> T:
                 foo = i1
 
             return foo.non_standard_operation()
+
+
+def test_magma_protocol_connection_iter():
```
rsetaluri, line 40:
```suggestion
def test_connection_iter():
```

tests/test_type/test_magma_protocol.py
```diff
@@ -35,3 +35,14 @@ def __call__(self, cond: m.Bit, i0: T, i1: T) -> T:
...
+    class Foo(m.Circuit):
+        io = m.IO(I=m.In(T), O=m.Out(T))
+        # Since this is not bulk wired, this will trigger the
+        # connection_iter logic
+        for i in range(8):
+            io.O._val[i] @= io.I._val[7 - i]
+
+    m.compile("build/test_magma_protocol_connection_iter", Foo)
```
rsetaluri, line 48: let's try to avoid compilation in tests where it's not needed, and test the API directly
```suggestion
    I = T(m.Out(T.T)(name="I"))
    O = T(m.In(T.T)(name="O"))
    # Trigger the connection_iter logic by doing a non-bulk wiring.
    for i in range(T.T.N):
        O._val[i] @= I._val[T.T.N - 1 - i]
    for idx, (o, i) in enumerate(O.connection_iter()):
        assert o is O._val[idx]
        assert i is I._val[T.T.N - 1 - idx]
```

magma/protocol_type.py
```diff
@@ -115,6 +115,9 @@ def __imatmul__(self, other):
         self.wire(other)
         return self
 
+    def connection_iter(self):
```
rsetaluri, line 118: is it correct that `connection_iter()` always returns `Iterable[m.Bit]`?

rsetaluri (changes requested):

magma/protocol_type.py
rsetaluri: ok, but we can expect it to be the base type and not the protocol type always?

rsetaluri (approved):

PR #1019, opened 2022-03-03
Promote concat return type

### LINKED ISSUE #1006: `concat` should return most specific type possible
Right now `concat` seems to return `Array[N, T]` for all invocations, e.g. `concat(SInt(), UInt()) : Array[N, Bit]`. We should instead maintain the most specific type as possible:

* If any argument to concat is `Array[Bit]` then the output should be `Array[Bit]`
* Otherwise, if any argument to concat is `Bits` then the output should be `Bits`
* If all arguments are `SInt` or all arguments are `UInt`, then return `SInt` and `UInt`, respectively
* In all other cases (e.g. mixed `SInt` and `UInt`) return `Bits`

rsetaluri:
Fixes #1006

### DIFF
--- a/magma/common.py
+++ b/magma/common.py
@@ -1,7 +1,8 @@
 import abc
 import collections
 import collections.abc
-from functools import wraps, partial
+from functools import wraps, partial, reduce
+import operator
 from typing import Iterable
 import warnings
 
@@ -205,3 +206,8 @@ class FinalizableDelegator(Finalizable):
     def finalize(self):
         for child in self._children.values():
             child.finalize()
+
+
+def lca_of_types(classes):
+    return next(iter(reduce(
+        operator.and_, (collections.Counter(cls.mro()) for cls in classes))))
--- a/magma/common.py
+++ b/magma/common.py
@@ -209,5 +209,7 @@ class FinalizableDelegator(Finalizable):
 
 
 def lca_of_types(classes):
+    # NOTE(rsetaluri): This implementation is inspired by:
+    #   https://stackoverflow.com/questions/58290137/how-to-find-most-recent-common-ancestor-base-type-of-several-types-in-python
     return next(iter(reduce(
         operator.and_, (collections.Counter(cls.mro()) for cls in classes))))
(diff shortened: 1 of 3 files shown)


magma/common.py
```diff
@@ -205,3 +206,8 @@ def get_child(self, key: str) -> Finalizable:
     def finalize(self):
         for child in self._children.values():
             child.finalize()
+
+
+def lca_of_types(classes):
+    return next(iter(reduce(
+        operator.and_, (collections.Counter(cls.mro()) for cls in classes))))
```
leonardt, line 213: Sometimes with code like this I find it useful to leave the link in case any future reader needs to understand it: https://stackoverflow.com/questions/58290137/how-to-find-most-recent-common-ancestor-base-type-of-several-types-in-python

magma/conversions.py
```diff
@@ -35,6 +35,36 @@
...
+
+class _Concatter(Finalizable):
+    def __init__(self):
+        self._ts = []
+        self._types = []
+
+    def _consume_impl(self, value):
+        if isinstance(value, ht.BitVector):
+            return value.bits(), Bits
+        if isinstance(magma_value(value), Bit):
```
leonardt, line 47: Should we just call `magma_value` once at the top?

magma/conversions.py
```diff
@@ -35,6 +35,36 @@
...
+        )
+
+    def consume(self, value):
+        new_ts, t = self._consume_impl(value)
+        self._ts.extend(new_ts)
+        self._types.append(t)
+
+    def finalize(self):
+        T = lca_of_types(self._types)
+        return convertbits(self._ts, None, T, checkbit=True)
```
leonardt, line 65: I don't think this works with non array of bits?

rsetaluri: yeah fixed that. good call

magma/conversions.py
rsetaluri: I'm not sure if that works with the checking for BitVector, bool cases here. Also, this maintains parity with the old code. I agree this should be made more uniform (not really clear where protocol types are allowed here). Will keep for now.

leonardt (approved):

PR #1032, opened 2022-03-14
[MLIR] Switch over to MLIR compiler

rsetaluri:
Place holder for all MLIR fixes. Until committed, any fix should merge into this branch, not master.


magma/backend/mlir/build_magma_graph.py
```diff
@@ -100,19 +110,27 @@ def _visit_driver(
...
-        cache_key = (ref.array, ref.index)
+        index = ref.index
+        if isinstance(index, slice):
+            assert index.step is None
+            assert index.start <= index.stop
+            index = index.start, index.stop
+            getter_cls, getter_args = MagmaArraySliceOp, index
+        else:
+            getter_cls, getter_args = MagmaArrayGetOp, (index,)
+        cache_key = (ref.array, index)
```
leonardt, line 121: Maybe move this into a helper function `get_cache_key`

magma/backend/mlir/hardware_module.py
```diff
@@ -373,57 +421,91 @@ def visit_magma_mux(self, module: ModuleWrapper) -> bool:
...
-            results=module.results)
+        data, select = module.operands[:-1], module.operands[-1]
+        # NOTE(rsetaluri): This is a specialized code path for the case where
+        # all tuples are flattened.
+        if self._ctx.opts.flatten_all_tuples and len(data) != height:
+            assert len(data) % height == 0
+            stride = len(data) // height
+            assert len(module.results) == stride
+            for i in range(stride):
+                operands = data[i::stride]
```
leonardt, line 432: operands is unused?

magma/backend/mlir/hardware_module.py
```diff
@@ -365,6 +402,17 @@ def visit_primitive(self, module: ModuleWrapper) -> bool:
...
+        def make_mux(data, select, result):
+            mlir_type = hw.ArrayType((len(data),), data[0].type)
+            array = self._ctx.new_value(mlir_type)
+            hw.ArrayCreateOp(
+                operands=data,
+                results=[array])
+            hw.ArrayGetOp(
+                operands=[array, select],
+                results=[result])
+
```
leonardt, line 415: Should we move this out of the enclosing function def since it doesn't use any references to the enclosing scope?

magma/backend/mlir/hardware_module.py
```diff
@@ -373,57 +421,91 @@ def visit_magma_mux(self, module: ModuleWrapper) -> bool:
...
+                if has_enable:
+                    ctx = push_block(sv.IfOp(operands=[enable]).then_block)
+                with ctx:
+                    sv.PAssignOp(operands=[reg, data])
+            if has_reset:
+                with push_block(always.reset_block):
+                    sv.PAssignOp(operands=[reg, const])
+            with push_block(sv.InitialOp()):
+                sv.BPAssignOp(operands=[reg, const])
+            sv.ReadInOutOp(operands=[reg], results=[result])
```
leonardt, line 464: Similar, should we move this to be an internal method?

magma/backend/mlir/hardware_module.py
```diff
@@ -373,57 +421,91 @@ def visit_magma_mux(self, module: ModuleWrapper) -> bool:
...
+                data_out.append(operands.pop(0))
+                init_out.append(make_selector(i).select(init))
+
+            visit_magma_value_or_value_wrapper_by_direction(
+                defn.I,
+                _visit_input,
+                _visit_input,
+                flatten_all_tuples=self._ctx.opts.flatten_all_tuples
+            )
+            return T_out, data_out, init_out
```
leonardt, line 480: Does this need to be a nested definition?

magma/backend/mlir/magma_common.py
```diff
@@ -67,18 +75,28 @@ def __iter__(self) -> Iterable['ValueWrapper']:
...
+        output_visitor: Callable[[Type], Any],
+        **kwargs):
+
+    def descend(v):
+        if isinstance(v, (m_Tuple, Array)):
+            for item in v:
+                visit_value_or_value_wrapper_by_direction(
+                    item, input_visitor, output_visitor, **kwargs)
+            return
+        raise TypeError(value)
```
leonardt, line 87:
```suggestion
        if not isinstance(v, (m_Tuple, Array)):
            raise TypeError(value)
        for item in v:
            visit_value_or_value_wrapper_by_direction(
                item, input_visitor, output_visitor, **kwargs)
```

leonardt: Could also be not a nested func

magma/view.py
```diff
@@ -43,13 +45,16 @@ def __getattr__(self, key):
...
+        path = [self.port.name.name]
         curr = self.parent
-        hierarchical_path = curr.inst.name + ";"
-        while isinstance(curr.parent, InstView):
-            hierarchical_path = curr.parent.inst.name + ";" + hierarchical_path
+        while isinstance(curr, InstView):
+            path.insert(0, curr.inst.name)
             curr = curr.parent
-        return hierarchical_path
+        return tuple(path)
```
leonardt, line 54: Could use tuple addition instead of insert then tuple conversion to avoid the intermediate list

e.g.
```
path = (self.port.name, )
...
    path = (curr.inst.name, ) + path
```

leonardt (approved):
Minor comments

magma/backend/mlir/magma_common.py
rsetaluri: good point -- doesn't strictly need to be, but it's easier to enclose the kwargs this way

magma/backend/mlir/build_magma_graph.py
rsetaluri: not sure if what you meant, but factored out the get/set functionality

magma/backend/mlir/hardware_module.py
rsetaluri: this definitely looks wonky, but needs to be a closure. many of the common args such as always_operands, enable, has_reset are shared across the invokations of this function. ideally would be all factored out, into functions with good interfaces -- this is hacky but is mostly clean

PR #1034, opened 2022-03-15
Fix bug in inline_verilog clock sharing logic

leonardt:
There is logic in inline verilog to share the same wire for default clock that
forgot to check if the clock type was an input.  This caused clock outputs to
be seen as undriven and use the shared wire.  This sharing logic should only be
used for undriven inputs.

### DIFF
--- a/magma/inline_verilog.py
+++ b/magma/inline_verilog.py
@@ -94,7 +94,8 @@ def _insert_temporary_wires(
                     f"{repr(orig_value)}")
 
         key = value
-        is_undriven_clock = isinstance(value, ClockTypes) and value.is_input() and not value.driven()
+        is_undriven_clock = (isinstance(value, ClockTypes) and value.is_input()
+                             and not value.driven())
         if is_undriven_clock:
             # Share wire for undriven clocks so we don't
             # generate a separate wire for the eventual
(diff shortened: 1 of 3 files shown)


magma/inline_verilog.py
```diff
@@ -94,7 +94,8 @@ def _insert_temporary_wires(
                     f"{repr(orig_value)}")
 
         key = value
-        if isinstance(value, ClockTypes) and not value.driven():
+        if (isinstance(value, ClockTypes) and value.is_input()
+                and not value.driven()):
```
rsetaluri, line 98:
```suggestion
        is_undriven_clock = isinstance(value, ClockTypes) and value.is_input() and not value.driven()
        if is_undriven_clock:
```

rsetaluri: btw something I've found recently is that `value.driven() and value.trace() is None` can be true. e.g.

```python
x = m.Bit()
y = m.Bit()
y @= x
assert y.driven()  # driven by x
assert y.trace() is None  # x is not driven by anything
```

rsetaluri (approved):

magma/inline_verilog.py
rsetaluri: yeah, agreed. we could change the name but as long as we have `.value()` and `.trace()` then we have all the info we need. seem like reasonable names

PR #1036, opened 2022-03-15
[Types] Update tuple to use lazy children pattern

leonardt:


leonardt:
Bump

rsetaluri:
I have a set of comments on the code, but I'm not sure I understand this change at a high-level.

It seems like we're trying to not have any sub-fields by avoiding constructing the children (unless they're passed into the constructor). But it seems like every method, (including `wire`) constructs the children by calling `iter(self)` which calls `__len__` and `__getitem__`, which constructs the child

I imagine we want wire to do a bulk wire if it's not elaborated?

leonardt:
Yes we should do a bulk wire when possible, but we can't do this for mixed direction tuples.

There are cases where tuple values never expand the children, for example when we use a tuple in the interface of a definition (the flipped versions eventually get wired) so there's still some benefit in this change to avoid elaboration by default, but it doesn't add the bulk wiring logic which I planned to add as a follow up.

leonardt:
Updated this to now share the common bulk wiring logic between array and tuple.

As a follow up, I'd like to do some audting to see where recursive operations are happening (causing expansion) and see if there's some common cases occurring that we should be avoiding.

magma/wire_container.py
```diff
@@ -191,3 +191,53 @@ def wire(self, o, debug_info):
         self._wire.connect(o._wire, debug_info)
         self.debug_info = debug_info
         o.debug_info = debug_info
+
+
+class WireableWithChildren(Wireable):
```
rsetaluri, line 196: let's rename this class `AggregateWireable`

magma/wire_container.py
```diff
@@ -191,3 +191,53 @@ def wire(self, o, debug_info):
...
+    def _has_elaborated_children(self):
```
rsetaluri, line 197: let's decorate these with `@abc.abstractmethod`

unfortunately, without inheriting from `abc.ABC` also, it will be a no-op, but at-least it serves as good documentation.

magma/wire_container.py
```diff
@@ -191,3 +191,53 @@ def wire(self, o, debug_info):
...
+        value = self._wire.value()
+        Wireable.unwire(self, value)
+
+        # Update children
+        for i, child in self._enumerate_children():
+            child.wire(value[i])
+
+    def _resolve_driving_bulk_wire(self):
+        driving = self._wire.driving()
+        # NOTE: we need to remove drivees before doing the recursive wiring of
```
rsetaluri, line 214: nit: please always give author for NOTE/TODO. e.g. `NOTE(leonardt)`

magma/wire_container.py
```diff
@@ -191,3 +191,53 @@ def wire(self, o, debug_info):
...
+
+class WireableWithChildren(Wireable):
+    def _has_elaborated_children(self):
+        raise NotImplementedError()
+
+    def _enumerate_children(self):
+        raise NotImplementedError()
+
+    def _resolve_driven_bulk_wire(self):
+        # Remove bulk wire since children will now track the wiring
```
rsetaluri, line 204: nit: please end all standalone comments with `.`.

magma/wire_container.py
```diff
@@ -191,3 +191,53 @@ def wire(self, o, debug_info):
...
+            self._resolve_driving_bulk_wire()
+
+
+def wireable_with_children_wrapper(fn):
+    def wrapper(self, *args, **kwargs):
+        if self._has_elaborated_children():
+            return fn(self, *args, **kwargs)
+        wireable_fn = getattr(WireableWithChildren, fn.__name__)
+        return wireable_fn(self, *args, **kwargs)
+    return wrapper
```
rsetaluri, line 243: nit

```suggestion
def wireable_with_children_wrapper(fn):

    @functools.wraps(fn)
    def wrapper(self, *args, **kwargs):
        if self._has_elaborated_children():
            return fn(self, *args, **kwargs)
        fn = getattr(WireableWithChildren, fn.__name__)
        return fn(self, *args, **kwargs)

    return wrapper
```

rsetaluri: let's also rename this function to `aggregate_wireable_method`

magma/wire_container.py
```diff
@@ -122,9 +122,9 @@ def trace(self, skip_self=True):
             return self._driver.trace(skip_self=False)
         if not skip_self and (self._bit.is_output() or self._bit.is_inout()):
             return self._bit
-        if not skip_self and self._bit.has_children():
-            # Could be an Array driven non-bulk, so check if children can be
-            # traced
+        if (not skip_self and self._bit.has_children() and
+                self._bit.driven()):
+            # Driven non-bulk, so trace children
```
rsetaluri, line 127: nit: I've been preferring this style lately (it's also the style our clients use)

```suggestion
        if (
            not skip_self and self._bit.has_children() and self._bit.driven()
        ):
            # Driven non-bulk, so trace children
```

magma/tuple.py
```diff
@@ -403,10 +420,10 @@ def unflatten(cls, value):
         return cls(*values)
 
     def flatten(self):
-        return sum([t.flatten() for t in self.ts], [])
+        return sum([t.flatten() for t in self], [])
 
     def const(self):
-        for t in self.ts:
+        for t in self:
```
rsetaluri, line 426: `return all(t.const() for t in self)`

magma/tuple.py
```diff
@@ -188,28 +188,28 @@ def __eq__(cls, rhs):
...
-            for k, t, T in zip(self.keys(), largs, self.types()):
+            for i, (k, t, T) in enumerate(zip(self.keys(),
+                                              largs,
+                                              self.types())):
                 if isinstance(t, (IntegerTypes, BitVector, Bit)):
                     t = T(t)
-                self.ts.append(t)
-                if not isinstance(self, AnonProduct):
-                    setattr(self, k, t)
-        else:
```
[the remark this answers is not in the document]
leonardt: Because this else case is responsible for populating `ts` in the default case (when it's not constructed with largs), but now we populate ts on demand

magma/array.py
```diff
@@ -3,19 +3,19 @@
...
-from .bitutils import int2seq, seq2int
+from .bitutils import int2seq
 from .debug import debug_wire, get_callee_frame_info
 from .logging import root_logger
 from .protocol_type import magma_type, magma_value
 
 from magma.operator_utils import output_only
-from magma.wire_container import WiringLog, Wire, Wireable
+from magma.wire_container import (WiringLog, AggregateWireable,
+                                  aggregate_wireable_method)
```
rsetaluri, line 18: nit:

```suggestion
from magma.wire_container import (
    WiringLog,
    AggregateWireable,
    aggregate_wireable_method,
)
```

and should go after `magma.protocol_type` import

magma/tuple.py
```diff
@@ -314,36 +340,66 @@ def unwire(self, o=None, debug_info=None):
...
-        return True
+        return all(t.driven() for t in self)
 
+    @aggregate_wireable_method
     def wired(self):
-        for t in self.ts:
-            if not t.wired():
+        return all(t.wired() for t in self)
+
+    # test whether the values refer a whole tuple
```
rsetaluri, line 351: prob an unnec comment, method name is as explanatory as comment

rsetaluri (approved):
last few nits

leonardt:
Note, this is encountering the same issue as https://github.com/phanrahan/magma/issues/1125 in the memory test, will provide a generic when fix then integrate it here.

PR #1041, opened 2022-03-23
Add DebugCircuit and DebugGenerator helpers

leonardt:


magma/__init__.py
```diff
@@ -79,7 +79,7 @@ def set_mantle_target(t):
 from magma.family import get_family
 _Family_ = get_family()
 
-from .generator import Generator, Generator2
+from .generator import Generator, Generator2, DebugGenerator
```
rsetaluri, line 82: For uniformity sake, I'd say let's call it DebugGenerator2. That way we can just Generator2 -> Generator everywhere once we switch

magma/circuit.py
```diff
@@ -926,3 +926,29 @@ def finalize(self, dont_instantiate: bool = False):
         for k, v in self._inst_attrs.items():
             setattr(inst, k, v)
         return inst
+
+
+class DebugCircuitKind(DefineCircuitKind):
```
rsetaluri, line 931: would call it DebugDefineCircuitKind

I actually think we should get rid of the DefineCircuitKind <-> CircuitKind indirection, but that can be later.

magma/circuit.py
```diff
@@ -926,3 +926,29 @@ def finalize(self, dont_instantiate: bool = False):
...
+        set_debug_mode(True)
+        # NOTE: Using super() here doesn't work:
+        #   result = super().__prepare__(name, bases, **kwargs)
+        #   TypeError: super(type, obj): obj must be an instance or subtype of
+        #   type
+
+        result = DefineCircuitKind.__prepare__(name, bases, **kwargs)
+        peek_definition_context_stack().set_metadata("prev_debug_mode",
+                                                     prev_debug_mode)
+        return result
```
rsetaluri, line 943:
```suggestion
        cls = DefineCircuitKind.__prepare__(name, bases, **kwargs)
        ctx = peek_definition_context_stack()
        ctx.set_metadata("prev_debug_mode", prev_debug_mode)
        return cls
```

magma/circuit.py
```diff
@@ -926,3 +926,29 @@ def finalize(self, dont_instantiate: bool = False):
...
+
+    def __new__(metacls, name, bases, dct):
+        set_debug_mode(
+            peek_definition_context_stack().get_metadata("prev_debug_mode")
+        )
+        result = DefineCircuitKind.__new__(metacls, name, bases, dct)
+        return result
```
rsetaluri, line 950:
```suggestion
        ctx = peek_definition_context_stack()
        set_debug_mode(ctx.get_metadata("prev_debug_mode"))
        return DefineCircuitKind.__new__(metacls, name, bases, dct)
```

magma/generator.py
```diff
@@ -71,24 +72,26 @@ def _make_key(cls, *args, **kwargs):
 def _make_type(cls, *args, **kwargs):
     dummy = type.__new__(cls, "", (), {})
     name = cls.__name__
-    bases = (Circuit,)
-    dct = DefineCircuitKind.__prepare__(name, bases)
+    bases = (cls._base_cls,)
+    dct = cls._base_kind.__prepare__(name, bases)
```
rsetaluri, line 76: kind -> metacls

rsetaluri: and let's use `sunderscore` for DSL stuff per caleb, so `_base_cls_`, `_base_metacls_`

tests/test_circuit/test_debug_circuit.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_debug_generator():
+    assert m.config.get_debug_mode() is False
+
+    class Foo(m.DebugGenerator):
+        def __init__(self, n: int):
+            assert m.config.get_debug_mode() is True
+            self.io = m.IO(I=m.In(m.Bits[n]))
+    Foo(4)
+    assert m.config.get_debug_mode() is False
```
rsetaluri, line 21:
```suggestion
def test_debug_circuit():
    assert m.config.get_debug_mode() is False

    class Foo(m.DebugCircuit):
        assert m.config.get_debug_mode() is True
        io = m.IO(I=m.In(m.Bit))

    assert m.config.get_debug_mode() is False


def test_debug_generator():
    assert m.config.get_debug_mode() is False

    class Foo(m.DebugGenerator):
        def __init__(self, n: int):
            assert m.config.get_debug_mode() is True
            self.io = m.IO(I=m.In(m.Bits[n]))

    Foo(4)
    assert m.config.get_debug_mode() is False
```

rsetaluri (changes requested):

magma/circuit.py
```diff
@@ -926,3 +926,27 @@ def finalize(self, dont_instantiate: bool = False):
...
+
+class DebugDefineCircuitKind(DefineCircuitKind):
+    def __prepare__(name, bases, **kwargs):
+        prev_debug_mode = get_debug_mode()
+        set_debug_mode(True)
+        # NOTE: Using super() here doesn't work:
+        #   cls = super().__prepare__(name, bases, **kwargs)
+        #   TypeError: super(type, obj): obj must be an instance or subtype of
+        #   type
+
```
rsetaluri, line 939:
```suggestion
        # NOTE(leonardt): Using super() here doesn't work:
        #   cls = super().__prepare__(name, bases, **kwargs)
        #   TypeError: super(type, obj): obj must be an instance or subtype of
        #   type
```
nit

rsetaluri (approved):

PR #1042, opened 2022-03-25
[MLIR] Add option to defer to CoreIR compiler for bind flow

rsetaluri:


rsetaluri:
@leonardt I realized we can just use the existing CoreIR bind flow just to get things working, since the code path to compile to bind modules is silo'd (and also always run as a pass anyway).

My plan is to do a refactor into bind2 or something -- something that is more amenable for MLIR compiler (specifically regarding cross module ref)

magma/backend/mlir/mlir_compiler.py
```diff
@@ -15,6 +15,9 @@
 
 class MlirCompiler(Compiler):
     def __init__(self, main: DefineCircuitKind, basename: str, opts: Dict):
+        # TODO(rsetaluri): Make this a better error.
+        assert "basename" not in opts
+        opts["basename"] = basename
```
leonardt, line 20: Is there a reason basename isn't just passed using the opts dictionary?

leonardt (approved):
Seems fine as a stop gap, one main thing to check is whether we'll need to support compile_guard inside bound modules? Then this will still be blocking it seems, but my initial thought would be that compile_guard matters more for guarding code inside RTL, whereas bound modules are "guarded" by including them or not.

PR #1045, opened 2022-03-28
Fix Array2 unwire issue

rsetaluri:


rsetaluri:
@leonardt lgtm, can't approve, but looks good

leonardt (approved):

leonardt:
@rsetaluri can you double check the changes to the .mlir gold files and see if they make sense? I'm not familiar enough with the syntax to understand the diff

rsetaluri:
> @rsetaluri can you double check the changes to the .mlir gold files and see if they make sense? I'm not familiar enough with the syntax to understand the diff

It looks semantically correct (also update corresponding verilog gold files -- verilog tests aren't run in CI b/c it requires circt binary).

However, the case where we wire an array slice seems to be no longer covered, i.e. https://github.com/phanrahan/magma/blob/7b657f19c70e77964d21f0ee6383d0a54268cc15/magma/backend/mlir/build_magma_graph.py#L129-L133 codepath is not trigerred.

Do you have a suggestion for a circuit where that'll still happen?

rsetaluri:
Also, I'm guessing we should be using a different API here for performance?

leonardt:
> Also, I'm guessing we should be using a different API here for performance?
> 
> https://github.com/phanrahan/magma/blob/7b657f19c70e77964d21f0ee6383d0a54268cc15/magma/backend/mlir/build_magma_graph.py#L106-L108

You could use `enumerate_children` to iterate by index or slice depending on the array structure.  There's also `connection_iter` if you want to the drivers as well

leonardt:
> > @rsetaluri can you double check the changes to the .mlir gold files and see if they make sense? I'm not familiar enough with the syntax to understand the diff
> 
> It looks semantically correct (also update corresponding verilog gold files -- verilog tests aren't run in CI b/c it requires circt binary).
> 
> However, the case where we wire an array slice seems to be no longer covered, i.e.
> 
> https://github.com/phanrahan/magma/blob/7b657f19c70e77964d21f0ee6383d0a54268cc15/magma/backend/mlir/build_magma_graph.py#L129-L133
> 
> codepath is not trigerred.
> Do you have a suggestion for a circuit where that'll still happen?

It looks like `test_array2_wire_to_slice` which wires two slices to slices should be covering this but the slices are being expanded now, let me investigate.

leonardt:
I fixed the unnecessary resolution of the slice objects that were being triggered by internal code paths.  In doing this, I exposed the "has_elaborated_children" and "enumerate_children" APIs to be used by other parts of the code (to avoid resolving slices when possible).  I'll need to update the Tuple2 PR to be compatible with this new API (for this code, the Tuple implementation always assume elaborated children).

leonardt:
> I fixed the unnecessary resolution of the slice objects that were being triggered by internal code paths. In doing this, I exposed the "has_elaborated_children" and "enumerate_children" APIs to be used by other parts of the code (to avoid resolving slices when possible). I'll need to update the Tuple2 PR to be compatible with this new API (for this code, the Tuple implementation always assume elaborated children).

This has exposed some cases in the coreir backend where we were "expanding" slices but then reconstructing them in the generated code.  I'll need to do some looking into whether we can recreate the old output code while still avoiding these "expansions" or if we just need to bulk update the generated code.

leonardt:
Hmm it looks like the array slice mlir op is not being used in the generated code, so let me see if I can figure out why

leonardt:
I think it might be because the mlir backend doesn't use the connection_iter API? I think when it recurses across Arrays to connect them, it should be using connection_iter (or something like enumerate_children) rather than enumerating by index to get the child slice objects

leonardt:
@rsetaluri does MLIR have support for an array create op that takes in slices?  I think the issue for this code:
```
            creator = MagmaArrayCreateOp(T)
            for i, element in enumerate(driver):
                creator_input = getattr(creator, f"I{i}")
                _visit_driver(ctx, creator_input, element, creator)
```
is that creator expects `T.N` inputs, where if we use enumerate_children, we can have less than N inputs.  Perhaps an alternative is use a `concat` operation for the drivers instead of an array create?

leonardt:
Psuedo code:
```
            conns = list(value.connection_iter())
            # concat input types based on children (single index or slice)
            creator = MagmaConcatOp([type(child) for child, _ in conns])
            for i, (_, driver) in enumerate(value.connection_iter()):
                creator_input = getattr(creator, f"I{i}")
                _visit_driver(ctx, creator_input, driver, creator)
```

leonardt:
It looks like I could define an array create op that is parametrized by type to do this, but I'm guessing somewhere I need to link this with an mlir implementation

leonardt:
Ok found the corresponding logic in magma//backend/mlir/hardware_module.py

leonardt:
Based on: https://circt.llvm.org/docs/Dialects/HW/#hwarray_create-circthwarraycreateop
Traits: SameTypeOperands might be an issue (since single index vs slice operands will be different)

leonardt:
Maybe we can try hw.array_concat

leonardt:
This adds the lifting logic: https://github.com/phanrahan/magma/pull/1045/commits/b1e3f888377c704dcdf82d7bf3a620824b0e745b

This does suggest a review of whether it makes more sense to have to lift single indices or expand slices from a perf perspective.  Ideally we'd have an Op that can either concat or create from single elements and slices (i.e. we could push this logic into MLIR to at least avoid a Python level performance hit)

leonardt:
I think the issues is popping up here because in magma/coreir, we use connection semantics where index -> index and slice -> slice are separate connections rather than having to fan-in into a create/concat node that handles both types.

leonardt:
@rsetaluri any feedback on the MLIR related changes discussed above? I'd like to merge this so we can get the fix in the next release, so we could continue to iterate on the MLIR stuff afterwards

PR #1048, opened 2022-04-01
Fix bug in enumerate children slice index logic

leonardt:
Before, we assumed the value of the slice was itself a slice object, but
after the changes in https://github.com/phanrahan/magma/pull/1045 this
is no longer true since an Array can be constructed with an
existing value as the slice.

The included shows one such case, where two slices are wired up to
InstRef values, and calling trace() returns an object where the slice
value is an instref.

This updates the logic to store the original slice_ key with the slice_
value in the data structure used for the slice lookup.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -389,8 +389,8 @@ class Array(Type, Wireable, metaclass=ArrayMeta):
 
         self._slices = {}
         self._unresolved_slices = {}
-        # Store mapping from slice start index to object for faster lookup when
-        # checking overlapping indicies/slices
+        # Store mapping from slice start index to slice key and object for
+        # faster lookup when checking overlapping indicies/slices
         self._slices_by_start_index = {}
 
     @classmethod
@@ -773,7 +773,7 @@ class Array(Type, Wireable, metaclass=ArrayMeta):
             slice_value = slice_T(name=ArrayRef(self, slice_))
             self._slices[key] = slice_value
             if not self._resolve_overlapping_indices(slice_, slice_value):
-                self._slices_by_start_index[key[0]] = slice_value
+                self._slices_by_start_index[key[0]] = (slice_, slice_value)
                 self._unresolved_slices[key] = slice_value
         return slice_value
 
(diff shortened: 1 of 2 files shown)


magma/array.py
```diff
@@ -773,7 +773,7 @@ def _get_slice(self, slice_):
             slice_value = slice_T(name=ArrayRef(self, slice_))
             self._slices[key] = slice_value
             if not self._resolve_overlapping_indices(slice_, slice_value):
-                self._slices_by_start_index[key[0]] = slice_value
```
rsetaluri, line 776: can negate teh condition on L771 and do early return instead.

tests/test_type/test_array2.py
```diff
@@ -444,3 +444,18 @@ class Foo(m.Circuit):
...
+
+    class Foo(m.Circuit):
+        T = m.Array[4, m.Bit]
+        io = m.IO(I=m.In(T), O=m.Out(T))
+        bar = Bar()
+        bar.I @= io.I[2:]
+        io.O[2:] @= bar.O
+        io.O[:2] @= bar.O
+        assert list(io.O.trace().connection_iter()) == [
+            (bar.O, None), (bar.O, None)]
```
rsetaluri, line 461: why is the second element None? shouldn't it be io.O[slice(0, 2)]? (or slice(2, 4))?

rsetaluri, line 461: why is the second element None? shouldn't it be io.O[slice(0, 2)]? (or slice(2, 4))?

leonardt: Trace returns the driver, the driver itself is not connected (not driven by anything)

leonardt: (Connection iter is a recursive trace over the children)

rsetaluri: i thought connection x.iter returns (driver of x_i, x_i) where x_i can be slice or element or whole array

leonardt: Yes this calls connection iter on the return value of a trace call, not the original value

rsetaluri: ooh missed that, ok makes sense. so does io.O.connection_iter returns [(bar.O, io.O[slice(0, 2)), (bar.O, io.O[slice[2, 4])]?

I'm just trying to understand what this fixes. What would have been different w.r.t. this test before this fix?

leonardt: `io.O.trace()` returns: `m.array([io.I[slice(2, 4)], io.I[slice(0, 2)])`
calling connection iter on that anon value, io.I is not driven

leonardt: I clarified the test by using `driven` (which covers the enumerate children code path) instead of `trace`

rsetaluri (approved):

PR #1049, opened 2022-04-01
Prototype inline_comb2

leonardt:
Here's a proposal for how we can architect the frontend for inline comb -> MLIR.

The overall structure is:
* Use a traditional "classmethod" definition syntax to provide a handle to the circuit definition (cls)
  * We choose this over the current inline_combinational semantics since that has some wonky pitfalls with trying to inherit the parent scope for delayed execution semantics.  Instead the user only has access to the scope through the `cls` object which is simple to provide when processing the code
* Use an AST pass to determine references to `cls`, these are used to generate a wrapper circuit/instance that is wired up to the referenced values (useful for maintaining consistency of the representation, e.g. tracing an output driver to the inside of the code)
* "inline" wrapper during MLIR code generation by directly translating the AST to MLIR code and replacing `cls` references to the corresponding MLIR values

Thoughts?

### DIFF
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -105,3 +105,4 @@ from magma.circuit_utils import circuit_stub, stubify, CircuitStub
 from magma.compile import MagmaCompileException
 from magma.linking import link_module, link_default_module, clear_link_info
 import magma.math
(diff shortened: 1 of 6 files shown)


leonardt:
We are dropping this for an when statement approach

PR #1050, opened 2022-04-04
Prototype debug rewriter for generators

leonardt:
This shows how we can get automated names for AnonRefs based on the assignment target.  There's still some work to generalize this to various assign patterns.

We could also add logic to add filename/lineno to the context so instances can grab that debug info (prepend every line with a function call that updates the context).


leonardt:
Closing in favor of preprocessor approach

PR #1053, opened 2022-04-19
Fix bug in overlapping slice resolution

leonardt:
Follow up to https://github.com/phanrahan/magma/pull/1045 and
https://github.com/phanrahan/magma/pull/1048

Because the collect_children logic now creates Arrays where the slice children
could be from other values, we need the slice resolution logic to decide
whether the slice value is from the current array (e.g. resolving x[0:2] for x)
or from another value (e.g. resolving y for [0:2] of an anon value).

This hotfixes the build issue on master for FB, but will need to add a unit
test to cover this case.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -679,10 +679,11 @@ class Array(Type, Wireable, metaclass=ArrayMeta):
 
     def _resolve_slice_children(self, start, stop, slice_value):
         for i in range(start, stop):
-            if slice_value._ts:
+            if (not isinstance(slice_value.name, ArrayRef) or
+                    slice_value.name.array is not self):
                 self._ts[i] = slice_value[i - start]
             else:
-                slice_value._ts[i - start] = self[i]
+                slice_value._ts[i - start] = self._get_t(i)
 
     def _resolve_slice_driver(self, start, stop, value):
         # When we encounter an overlapping slice that is already bulk driven,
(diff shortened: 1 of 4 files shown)


magma/array.py
```diff
@@ -724,10 +725,6 @@ def _filter(item):
         assert len(overlaps) == 1
         key, value = overlaps[0]
         self._remove_slice(key)
-        if value._ts:
-            self._ts[index] = value[index - key[0]]
-        else:
-            value._ts[index - key[0]] = self._ts[index] = self._make_t(index)
```
leonardt, line 730: This is redundant with the `_resolve_slice_children` logic, we don't need to duplicate it here.

magma/array.py
```diff
@@ -679,10 +679,11 @@ def _make_t(self, index):
 
     def _resolve_slice_children(self, start, stop, slice_value):
         for i in range(start, stop):
-            if slice_value._ts:
+            if (not isinstance(slice_value.name, ArrayRef) or
+                    slice_value.name.array is not self):
                 self._ts[i] = slice_value[i - start]
             else:
-                slice_value._ts[i - start] = self[i]
+                slice_value._ts[i - start] = self._get_t(i)
```
leonardt, line 686: Call _get_t directly here to avoid the more general __getitem__ logic that's already being covered in the calling code path (don't need to call that code more than once)

rsetaluri (approved):

PR #1055, opened 2022-04-20
Rewrite error logging

rsetaluri:
This change refactors a bunch of the internals regarding logging, especially during the circuit creation pipeline.

Now, we explicitly stage logs from the caller side of the context manager. There is also generic machinery to capture logs (at any time) and now we use that to attach logs to a definition context. These logs can be used for diagnostics in general, but we primarily use them to raise logs as exceptions (in RaiseLogsAsExceptionsPass, previously FindErrorsPass). I think these new abstractions are more clear and maintainable.

The only externally visible change is that I got rid of multiline logging since it is generally bad practice and IMO looked a bit clunky. Instead I split those out into multiple debug log statements. Since we don't have any parallelism, it is probably easy enough to inspect the logs to see the breakdown of the connected-/unconnected-ness of complex types nearby. I plan to either (a) add an ID to those messages so they can all be grouped, or (b) write to a separate debug file so that the port can be debugged.

### DIFF
--- a/magma/find_unconnected_ports.py
+++ b/magma/find_unconnected_ports.py
@@ -124,7 +124,7 @@ def _make_unconnected_port_diagnostic_visitor_cls():
     return _Visitor
 
 
-def check_unconnected(ckt):
+def find_and_log_unconnected_ports(ckt):
     infos = find_unconnected_ports_of_circuit(ckt)
     for info in infos:
         port = info.port
--- a/magma/find_unconnected_ports.py
+++ b/magma/find_unconnected_ports.py
@@ -1,4 +1,6 @@
 import dataclasses
+import functools
+import itertools
 from typing import Optional
 
 from magma.clock import is_clock_or_nested_clock, ClockTypes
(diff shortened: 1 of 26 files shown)


magma/find_unconnected_ports.py
```diff
@@ -0,0 +1,136 @@
...
+            if value.trace() is not None:
+                yield UnconnectedPortDiagnostic(value, True, self._depth)
+                return
+            wrapped = self.wrap(value)
+            results = []
+            has_connected = False
+            self._depth += 1
+            for child in wrapped.children:
+                for diagnostic in self.visit(child):
+                    has_connected = has_connected or diagnostic.connected
```
leonardt, line 115:
```suggestion
                    has_connected |= diagnostic.connected
```

leonardt (approved):

magma/find_unconnected_ports.py
rsetaluri: @leonardt actually this doesn't work since it relies on diagnostic.connected being possibly a NoneType (and that doesn't work for bitwise or).

PR #1056, opened 2022-04-21
Fix bug in wire children logic missing unwire

leonardt:
Before, we were relying on the "overwriting" log message when wiring a
second driver to a child.  However, this is problematic because we don't
actually unwire the child being overwritten.  As as result, in various
code paths (e.g. error reporting) we can trigger one of the overwritten
drivers to expand itself, which in turn calls the
_resolve_driven_bulk_wires logic and corrupts the representation.  By
unwiring the overwritten driver, we avoid any code acciddentally
triggering the resolve logic.

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -15,7 +15,8 @@ from .logging import root_logger
 from .protocol_type import magma_type, magma_value
 
 from magma.operator_utils import output_only
-from magma.wire_container import WiringLog, Wire, Wireable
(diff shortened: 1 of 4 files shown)


leonardt:
I've generalized the fix by moving it out of the array logic and into the wireable logic.  Basically, when we encounter an "overwrite" wire, we should unwire the existing driver

rsetaluri (approved):

PR #1059, opened 2022-04-25
Add default unwire API

leonardt:


magma/wire_container.py
```diff
@@ -39,7 +39,11 @@ def __str__(self):
     def anon(self):
         return self._bit.anon()
 
-    def unwire(self, other):
+    def unwire(self, other=None):
+        if other is None:
+            if self._driver is None:
+                raise RuntimeError("Cannot unwire undriven value")
```
rsetaluri, line 45: what's the argument for raising an error here? I could see it both ways, either (a) unwiring an unwired value raises, or (b) it's valid since it's like calling clear() on an empty list.

tests/test_wire/test_unwire.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+import magma as m
+
+
+@pytest.mark.parametrize('T', [m.Bit, m.Bits[4], m.Tuple[m.Bit, m.Bits[4]],
+                               m.Array[4, m.Bits[4]]])
+@pytest.mark.parametrize('func', [lambda x, y: x.unwire(y), m.unwire])
+def test_unwire_with_arg(T, func, caplog):
```
rsetaluri, line 8: * Let's write these tests without circuits, e.g.
```python
x, y, z = T(), T(), T()
...
```
* Let's use `not any(record.level >= logging.WARNING for record in caplog.records)` (`import logging` at top)
* It's probably worth saving the T and func parameterizations as variables. will reduce LOC and make it easier to write new tests if needed.

magma/array.py
```diff
@@ -620,10 +620,13 @@ def wire(self, o, debug_info):
...
 
-    def unwire(self, o):
+    def unwire(self, o=None):
         if self._has_elaborated_children():
             for i, child in self._enumerate_children():
-                child.unwire(o[i])
+                if o is not None:
+                    child.unwire(o[i])
+                else:
+                    child.unwire()
```
rsetaluri, line 629:
```suggestion
                o_i = None if o is None else o[i]
                child.unwire(o_i)
```

magma/tuple.py
```diff
@@ -301,8 +301,13 @@ def wire(i, o, debug_info):
             o_elem = magma_value(o_elem)
             wire(o_elem, i_elem, debug_info)
 
-    def unwire(i, o):
-        for i_elem, o_elem in zip(i, o):
+    def unwire(self, o=None):
+        if o is None:
```
rsetaluri, line 305: any particular reason to do if-for here rather than for-if as in array? probably a little more readable to make them uniform

rsetaluri (changes requested):

magma/wire_container.py
leonardt: I don't feel strongly either way.  We could make it a warning log just like overriding a driven value?

rsetaluri (approved):

PR #1062, opened 2022-04-26
Alias m.Register to m.Reg

rsetaluri:
Adds an alias to `m.Register` as `m.Reg`. This is done in a special section of `magma/__init__.py` with a simple equality statement. We could also do an `import .. as ...` but this seems simpler.

### DIFF
--- a/magma/__init__.py
+++ b/magma/__init__.py
@@ -105,3 +105,11 @@ from magma.circuit_utils import circuit_stub, stubify, CircuitStub
 from magma.compile import MagmaCompileException
 from magma.linking import link_module, link_default_module, clear_link_info
 import magma.math
+
+################################################################################
+# BEGIN ALIASES
+################################################################################
+Reg = Register
+################################################################################
+# END ALIASES
+################################################################################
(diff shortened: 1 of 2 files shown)


magma/__init__.py
```diff
@@ -105,3 +105,11 @@ def set_mantle_target(t):
...
 from magma.linking import link_module, link_default_module, clear_link_info
 import magma.math
+
+################################################################################
+# BEGIN ALIASES
+################################################################################
+Reg = Register
+################################################################################
+# END ALIASES
+################################################################################
```
leonardt, line 115: Should we consider providing this in magma.primitives.register and importing it here? This would allow internal code (e.g. other primitives) to reference it without using the top level import

leonardt (approved):

magma/__init__.py
rsetaluri: yeah I could go either way on it. I do kind of like the isolation of having a set of aliases only for external users (i.e. through __init__), which seems like a different case than desiring it internally.

If we see a reason to use Reg internally then I can move it there, otherwise I'd argue it should be in `__init__.py`

PR #1069, opened 2022-05-05
Fix #1068

### LINKED ISSUE #1068: Calling `m.bits` on a UInt should convert the value to Bits
This code: https://github.com/phanrahan/magma/blob/master/magma/conversions.py#L126-L129

Is causing a short circuiting and skipping the conversion.

leonardt:
Force upcasting when calling `m.bits` on a UInt/SInt

### DIFF
--- a/magma/smart/smart_bits.py
+++ b/magma/smart/smart_bits.py
@@ -497,7 +497,7 @@ class SmartBits(_SmartBitsExpr, metaclass=_SmartBitsMeta):
         super().__init__(self)
         if value is None:
             value = type(self)._to_magma_()()
-        self._value = bits(value)
+        self._value = value
 
     def typed_value(self):
         if type(self)._signed:
--- a/tests/test_type/test_bit.py
+++ b/tests/test_type/test_bit.py
@@ -380,8 +380,3 @@ EndCircuit()\
 def test_errors(op):
     with pytest.raises(ValueError):
         op(m.Bit(name="b"))
-
-
-def test_bit_upcast():
-    x = m.Enable()
-    assert type(m.bit(x)) is m.Bit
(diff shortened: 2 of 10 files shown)


magma/smart/smart_bits.py
```diff
@@ -497,7 +497,7 @@ def __init__(self, value=None):
         super().__init__(self)
         if value is None:
             value = type(self)._to_magma_()()
-        self._value = bits(value)
+        self._value = value
```
leonardt, line 500: @rsetaluri does this make sense? We can't do a `m.bits` conversion here because it will cause an error in the interface construction pipeline, see https://github.com/phanrahan/magma/runs/6312244536?check_suite_focus=true#step:5:1191

```
self = <magma.interface.IO object at 0x7f5fe9840ee0>
defn = _Test = DeclareCircuit("_Test", "I0", SmartBits[8, False], "O1", SmartBits[4, False], "O2", SmartBits[16, False])

    def bind(self, defn):
        if self._bound:
            raise Exception("Can not bind IO multiple times")
        for port in self._ports.values():
>           port.name.set_defn(defn)
E           AttributeError: 'AnonRef' object has no attribute 'set_defn'

magma/interface.py:454: AttributeError
```

Basically, calling `m.bits` on the port value will create an anon bits value, that's no longer a DefnRef.  Based on this change, the original code was doing a NOOP anyways since value was an instance of Bits, so removing the conversion should be okay?

tests/test_type/test_bit.py
```diff
@@ -380,3 +380,8 @@ class TestITE(m.Circuit):
 def test_errors(op):
     with pytest.raises(ValueError):
         op(m.Bit(name="b"))
+
```
rsetaluri, line 383: move this to conversions as well?

rsetaluri (commented):
Looks good modulo minor comment on test and nit on comment.

rsetaluri (approved):

PR #1070, opened 2022-05-09
Support wiring Bit and Bits[1]

leonardt:
Two implementation issues:
1. Should we support wiring T to Array[1, T] in general? Or have this be
   a Bits/Bit specific feature
2. Is there a way to avoid a circular import in this logic? The naive
   solution is to have Bit check if the driver is Bits[1], but this
   clearly requires a circular import.  Perhaps we could use some sort
   of `T.convert_to(other_T)` API?

### DIFF
--- a/magma/bit.py
+++ b/magma/bit.py
@@ -14,13 +14,8 @@ from .digital import VCC, GND  # TODO(rsetaluri): only here for b.c.
 from magma.compatibility import IntegerTypes
 from magma.debug import debug_wire
 from magma.family import get_family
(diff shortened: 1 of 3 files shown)


rsetaluri:
I'm still not sure if we really want to support this pattern (our original ideas about type strictness).

But if we do, I think we definitely should support all `T <-> Array[1, T]` wirings.

leonardt:
Perhaps we can investigate other options with the designers to avoid the issue without making this fundamental change.

PR #1072, opened 2022-05-13
Add inline verilog expression primitive

rsetaluri:
This is a temporary "hack" to allow inline verilog strings to be interpreted as outputs, e.g.

```python
io.O @= m.InlineVerilogExpression("THIS_THING", T)
```

generates

```verilog
assign O = THIS_THING;
```


tests/gold/test_inline_veriog_expression_compilation_coreir-verilog.v
```diff
@@ -0,0 +1,60 @@
...
+
+module MyWrapperGen_1 (
+    input [0:0] I,
+    output [0:0] O,
+    input CLK
+);
+wire [0:0] _magma_inline_wire0;
+InlineVerilogExpression_Bits1_a489b4461c0192fc InlineVerilogExpression_Bits1_a489b4461c0192fc_inst0 (
+    .O(O)
+);
```
leonardt, line 34: Do we intend to support this through the CoreIR pipeline? If so, should this be inlined code instead? If not, let's remove this part of the test.

rsetaluri: Btw, this is just a temporary change to make the following code work:

```python
class Reg(m.Circuit):
    io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit)) + m.ClockIO()
    verilog = """\
  reg R;
  always @(posedge CLK)
    R <= I;
  assign O = R;"""
```

work with the MLIR backend by transforming it to:

```python
class Reg(m.Circuit):
    io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit)) + m.ClockIO()
    m.inline_verilog("""\
  reg R;
  always @(posedge CLK)
    R <= I;""")
    io.O @= m.inline_verilog_expression("R")
```

The reason we need to do this is that MLIR needs to know that outputs are driven, so simply doing `sv.verbatim <verilog>` when `<verilog>` assigns to outputs doesn't work. _Inline verilog expressions_ (as I have named them) allow us to get the value contained by a piece of inline verilog and use it in magma, which enables us to assign to outputs (so that MLIR is happy).

Now, supporting this pattern in CoreIR would either require transforming inline verilog expressions to normal inline verilog, or supporting inline verilog expressions directly in CoreIR as a primitive. (The former has some quirks, won't dive too deep into them.)

**This feature is meant to be temporary**, (to enable some features at a customer) and should be reverted ASAP once those things are fixed at the customer. So I just wanted to make sure that magma -> CoreIR -> verilog wasn't "broken" (doesn't crash, although isn't complete), so I had this test here.

leonardt (approved):

leonardt:
Perhaps it's better to detect this in the CoreIR backend and raise a NotImplementedError just so people know it's not supported?

PR #1075, opened 2022-05-19
When

leonardt:
* magma maintains a global stack that contains a pointer to the active when condition stack (a stack of stacks is used to match per definition when condition stacks)
* the wiring logic is aware of this stack, and does a conditional wire if there is an active when condition
* a finalize step is added that resolves conditional drivers using a generated defn/inst that emits verilog code that matches the original when statement structure
  * my goal here was two fold: (1) provide a basis for the MLIR implementation that should look similar but construct SV dialect instead of inline verilog, (2) provide a prototype implementation that we can use now to test out the frontend
  * we may want to delay any significant code review changes until we have the MLIR version, but general/high level design feedback will likely still be useful for the future MLIR impl

Note, I was able to avoid a circular import everywhere except for generating the implementation in the finalize step (Circuit depends on impl generator which depends on Circuit), we will likely be able to avoid this by generating MLIR instead, which avoids having this circular import (if we can avoid generating the intermediate circuit/inst)


leonardt:
Note #1108 fixes a bug where the key in the conditional driver dictionary could be clobbered by defn/instances with the same port names.  We should either cherry-pick that fix in before merging or merge that with this (the bug came up in the register test where the port name collision was found)

leonardt:
Just ran into an issue in `sum` where using an instance doesn't behave properly because the inputs/outputs are wired conditionally.  Working on an improvement that I'll cherry-pick into here so we don't depend on `sum` but I think the resolution will be to only conditionally assign values when `@=` is used (versus calling wire, which can happen explicitly by the user or implicitly in expressions).  I think this is a reasonable choice since it helps us distinguish between the conditional assignments versus general wiring (e.g. for instsances), and it can be easily explained as only supported for the `@=` operator rather than general wiring.

leonardt:
Turns out this issue is a bit more complicated than I initially though, take this example:

```
m.when(cond):
    x @= foo(y)
```

where `foo` could be a simple helper function that does something like
```
def foo(y):
    w = Bit()
    w @= y.reduce_and()
    return w
```

Here we actually want the wiring to be done unconditionally even though it's called inside a when statement.

I think the proper fix is to have a notion of values created inside a when context being unconditionally assigned inside that context (unless there's another nested when, at which point they'll be conditionally assigned without the enclosing when contexts condition).

I think it's fairly straightforward to have value creation aware of enclosing when contexts (which would also handle the above mentioned instance wiring case).  There's the next wrinkle of how this would handled in the MLIR/SV backend.  Basically, values would be created, but only assigned in the appropriate sub-when contexts when you consider nesting.  This could lead to potential latches even though the values are never used outside of the enclosing `if` statements.  One workaround would be to emit a separate if statement that drops the enclosing conditions (which loses the original structure).  I'll see if I can come up with some other option (What would verilog do if we assign `x` for the unused conditions, i.e. we don't care? Or we could just provide some default value for those cases, but this would need to be done before latch detection)

leonardt:
> There's the next wrinkle of how this would handled in the MLIR/SV backend. Basically, values would be created, but only assigned in the appropriate sub-when contexts when you consider nesting. 

This is a fairly HW specific issue here, because in software we'd do something like define the variable inside the body of the enclosing `if` statement and only refer to inside there (so other cases wouldn't matter).  However, in our code structure we declare values initially (as registers) outside the if statement body.  That's why we could consider providing `x` or default for the cases we don't care.  Unless SV supports declaring a variable inside an `if` statement.

rsetaluri:
Closing for #1113, since for some reason this can't be rebased...

PR #1077, opened 2022-05-25
[MLIR] Add undriven support

rsetaluri:
This change adds support for the coreir.undriven primitive in the MLIR backend. The MLIR (IR) we emit contains dangling wires that hook up to "undriven" ports.

NOTE: SV::WireOp canonicalization converts undriven wires to be driven by X. So if we run canonicalization (which we definitely should), the dangling wires will be assigned to `X`.

### DIFF
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -358,6 +358,12 @@ class ModuleVisitor:
             return self.visit_coreir_wire(module)
         if defn.coreir_name == "term":
             return True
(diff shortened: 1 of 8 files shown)


rsetaluri:
I actually figured out how to not have dangling wires driven by X: [df71d2a (https://github.com/phanrahan/magma/pull/1077/commits/df71d2aaa27e9d73d7e38b8a51fd25a2215f24d5)

leonardt (approved):

PR #1078, opened 2022-05-31
Add bits rewire test/fix

leonardt:


tests/test_type/test_bits.py
```diff
@@ -474,3 +474,11 @@ class Foo(m.Circuit):
...
         # Dummy.__rand__
         assert (io.I & Dummy()) is io.I
+
+
+def test_bits_rewire_warning(caplog):
+    x = m.Bit(name="x")
+    x @= 0
+    x = m.as_bits(x)
+    x @= 0
+    assert len(caplog.messages)
```
rsetaluri, line 484: can use `has_warning()`?

rsetaluri (approved):
Seems like it doesn't break anything but I'm wary to see if client is affected...

rsetaluri:
Yup, that was my thinking. If there were internal places we should do unwire.

PR #1080, opened 2022-06-01
[Stubify] Add dont override driven option

rsetaluri:


leonardt (approved):
LGTM

Does it make sense to have the zero_stubifier be default? This wouldn't break existing code, and I'd perhaps think that the no_override_driven behavior is a special case but since we're probably the only users right now I don't think it's a big issue.

rsetaluri:
> LGTM
> 
> Does it make sense to have the zero_stubifier be default? This wouldn't break existing code, and I'd perhaps think that the no_override_driven behavior is a special case but since we're probably the only users right now I don't think it's a big issue.

Yeah, it could definitely be seen as a special option, but as you said for now we only have one client so it's probably fine to make this default. I tried to design it in a way that swapping out a new stubbifier would be easy.

PR #1083, opened 2022-06-08
Add name argument for conversions

### LINKED ISSUE #1081: [Types] Add name argument for conversions
Example: `m.array(val, name="val_arr")`

Convenience argument for naming the converted value to avoid having to define a separate intermediate value for after conversion.

leonardt:
Fixes #1081

### DIFF
--- a/magma/conversions.py
+++ b/magma/conversions.py
@@ -60,7 +60,7 @@ def can_convert_to_bit_type(value):
     return issubclass(magma_type(value), (Digital, Array, Tuple))
 
 
-def convertbit(value, totype):
+def convertbit(value, totype, name):
     # NOTE: We don't do `isinstance` here because we want an upcast to cause a
     # conversion
     value = magma_value(value)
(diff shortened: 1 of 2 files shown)


magma/conversions.py
```diff
@@ -200,30 +203,34 @@ def convertbits(value, n, totype, checkbit):
             Direction.InOut: InOut
         }[T.direction](Bit)
 
-    value = totype[len(Ts), T](ts)
+    value = totype[len(Ts), T](ts, name=name)
     if n is not None and len(value) < n:
+        # TODO: The extended value isn't named, but perhaps we'd like to move
```
rsetaluri, line 208: yeah I think we want to move to making the conversion operators not do extensions. imagine it will break a lot right now so probably will be a sizeable chagne (and affects outside people).

magma/conversions.py
```diff
@@ -200,30 +203,34 @@ def convertbits(value, n, totype, checkbit):
...
+        # to an explicit convert + extend rather than doing them in a single
+        # operation? If so, then we could provide the same name interface for
+        # the extension operators
```
rsetaluri, line 211: nit
```suggestion
        # TODO(leonardt): The extended value isn't named, but perhaps we'd like to move
        # to an explicit convert + extend rather than doing them in a single
        # operation? If so, then we could provide the same name interface for
        # the extension operators.
```

tests/test_conversion.py
```diff
@@ -84,3 +84,13 @@ def test_bits_upcast(T):
 def test_bit_upcast():
     x = m.Enable()
     assert type(m.bit(x)) is m.Bit
+
+
+def test_name():
```
rsetaluri, line 89: can we parameterize this test by a handful of conversions? and additionally check that the conversion happened properly?

rsetaluri (approved):

PR #1090, opened 2022-06-10
[MlirToVerilog] Revamp MlirToVerilog flow

rsetaluri:


magma/backend/mlir/mlir_to_verilog.py
```diff
@@ -48,10 +58,31 @@ def _make_stream(filename, mode, default):
...
+
 def mlir_to_verilog(istream: io.RawIOBase, ostream: io.RawIOBase = sys.stdout):
-    circt_home = _get_circt_home()
-    opt_cmd = _make_opt_cmd(circt_home)
-    _subprocess_run(opt_cmd, istream, ostream)
+    circt_home = _circt_home()
+    cmd = _circt_opt_cmd(circt_home)
+    returncode = _run_subprocess(cmd, istream, ostream)
+    if returncode != 0:
+        raise MlirToVerilogException()
```
leonardt, line 85:
```suggestion
        raise MlirToVerilogException(f"Error running {cmd}, got returncode {returncode}")
```

magma/testing/utils.py
```diff
@@ -125,3 +126,21 @@ def non_standard_operation(self):
...
+        def wrapper(*args, **kwargs):
+            prev_value = getattr(config, key)
+            setattr(config, key, value)
+            ret = fn(*args, **kwargs)
+            setattr(config, key, prev_value)
+            return ret
+
+        return wrapper
+
+    return deco
```
leonardt, line 146: One thing, when using the Python fixture approach with explicit setup/teardown stages, this should handle the case when the test throws an uncaught exception.  Otherwise, pytest won't know to run the teardown code even when the test fails (not normally a problem, but in theory this could cause downstream tests to also fail if the config state isn't reverted after one of these tests fail)

tests/test_backend/test_mlir/test_mlir_to_verilog.py
```diff
@@ -0,0 +1,59 @@
...
+def test_bad_input():
+    _skip_if_circt_opt_binary_does_not_exist()
+
+    istream = io.TextIOWrapper(io.BytesIO())
+    istream.write("blahblahblah")
+    istream.seek(0)
+    ostream = io.TextIOWrapper(io.BytesIO())
+
+    with pytest.raises(MlirToVerilogException):
+        mlir_to_verilog(istream.buffer, ostream.buffer)
```
leonardt, line 59:
```suggestion


def _run_test(write=None):
    istream = io.TextIOWrapper(io.BytesIO())
    ostream = io.TextIOWrapper(io.BytesIO())
    if write:
        istream.write("module {}\n")
        istream.seek(0)
    mlir_to_verilog(istream.buffer, ostream.buffer)
    return istream, ostream

@with_config("circt_home", "/this_cant_be_a_real_directory/")
def test_no_binary():
    with pytest.raises(FileNotFoundError):
        _run_test()


def test_basic():
    _skip_if_circt_opt_binary_does_not_exist()
    _run_test()


def test_module():
    _skip_if_circt_opt_binary_does_not_exist()
    _, ostream = _run_test("module {}\n")

    ostream.seek(0)
    assert ostream.read() == ""


def test_bad_input():
    _skip_if_circt_opt_binary_does_not_exist()

    with pytest.raises(MlirToVerilogException):
        _run_test("blahblahblah")
```

leonardt (changes requested):
LGTM, minor comments

magma/testing/utils.py
rsetaluri: :+1: using that now

leonardt (approved):

PR #1093, opened 2022-06-14
[CompileGuard] Compile guard revamp (2)

rsetaluri:


magma/compile_guard.py
```diff
@@ -1,40 +1,79 @@
...
+        external @= driver
+
+    def _visit_output_connection(self, driver: Type, drivee: Type):
+        new_port = self._builder.add_port(Out(type(driver)))
+        new_port @= driver
+        external = getattr(self._builder, new_port.name.name)
+        old_driver = drivee.value()
+        drivee.unwire(old_driver)
+        new_driver = external
+        drivee @= new_driver
```
leonardt, line 49: We could consider adding an a "rewire" API, something like:

```python
drivee.rewire(new_driver)
```

which would suppress the rewiring warning by internal doing the unwire call.

magma/compile_guard.py
```diff
@@ -1,40 +1,79 @@
...
+        # to handle the undriven port.
+        if not isinstance(port, ClockTypes):
+            _logger.debug(f"found undriven port: {port}")
+            return
+        T = type(port).undirected_t
+        if T in self._clock_types:
+            return  # only add at most one port for each clock type
+        self._clock_types.add(T)
+        name = str(port.name)
+        _ = self._builder.add_port(In(T), name=name)
```
leonardt, line 65:
```suggestion
       self._builder.add_port(In(T), name=name)
```

magma/compile_guard.py
```diff
@@ -1,40 +1,79 @@
...
 
 
 class _CompileGuardBuilder(CircuitBuilder):
+    __default_defn_name_counter = 0
+
+    @staticmethod
+    def make_default_defn_name() -> str:
+        counter = _CompileGuardBuilder.__default_defn_name_counter
+        _CompileGuardBuilder.__default_defn_name_counter += 1
```
leonardt, line 74: hat tip caleb for a nice pattern to do this using [itertools.count](https://docs.python.org/3/library/itertools.html#itertools.count)

```suggestion
    __default_defn_name_counter = itertools.count()

    @staticmethod
    def make_default_defn_name() -> str:
        counter = next(_CompileGuardBuilder.__default_defn_name_counter)
```

magma/compile_guard.py
```diff
@@ -45,146 +84,51 @@ def __init__(self, name, cond, type):
...
+        return self._instances.copy()
 
+    def open(self):
+        return self._open()
 
-class _CompileGuard:
-    __index = 0
+    def _new_port_name(self):
+        name = f"port_{self._num_ports}"
+        self._num_ports += 1
```
leonardt, line 102: Could use `itertools.count` here too

magma/compile_guard.py
```diff
@@ -45,146 +84,51 @@ def __init__(self, name, cond, type):
...
-                  type: Optional[str] = "defined"):
-    return _CompileGuard(cond, defn_name, inst_name, type)
+def _make_builder(
+        cond: str,
+        defn_name: Optional[str],
+        inst_name: Optional[str],
+        type: str
+) -> _CompileGuardBuilder:
+    if defn_name is None:
+        defn_name = _CompileGuardBuilder.make_default_defn_name()
```
leonardt, line 113: Could move this inside the `_CompileGuardBuilder` init logic to make more reuseable

magma/passes/group.py
```diff
@@ -0,0 +1,118 @@
...
+            yield from self._run_on_input(elt)
+
+    def _run_on_output(self, port: Type) -> Iterable[Connection]:
+        for driver, drivee in _get_driving(port):
+            is_external = self._is_external(drivee)
+            if is_external is not None:
+                if is_external:
+                    yield (driver, drivee)
+                continue
+            assert not drivee.name.bound()
```
leonardt, line 89: I think this is redundant with the _is_external assertion

magma/ref.py
```diff
@@ -201,6 +237,12 @@ def anon(self):
     def parent(self):
         return self.tuple.name
 
+    def named(self) -> bool:
+        return self.parent().named()
+
+    def bound(self) -> bool:
+        return self.parent().bound()
+
```
leonardt, line 245: Could add an intermediate class for Array/Tuple ref named "RecursiveRef" that defines these methods

leonardt (approved):
LGTM, minor comments

leonardt:
I believe this provides the anon value support (driver) needed for https://github.com/phanrahan/magma/pull/999, should we double check those tests pass too? (or are covered by the added tests)

PR #1099, opened 2022-06-28
[Bind] Bind2

rsetaluri:


magma/backend/mlir/xmr_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    if isinstance(ref, TupleRef):
+        path = _ascend_to_leaf(ref.tuple, leaves)
+        return path + [f".{ref.index}"]
+    if isinstance(ref, ArrayRef):
+        path = _ascend_to_leaf(ref.array, leaves)
+        return path + [f"[{ref.index}]"]
+    raise TypeError(ref, value)
+
+
+def _advance(it):
```
leonardt, line 56: Dead code?

magma/backend/mlir/xmr_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    return seperator.join(path)
+
+
+def _get_leaf_descendants(tree, node, include_self=False):
+    is_leaf = True
+    for succ in tree.successors(node):
+        is_leaf = False
+        yield from _get_leaf_descendants(tree, succ, include_self=True)
+    if is_leaf and include_self:
+        yield node
```
leonardt, line 32: Could just use succ as the flag
```suggestion
    succ = None
    for succ in tree.successors(node):
        yield from _get_leaf_descendants(tree, succ, include_self=True)
    if succ is not None and include_self:
        yield node
```

magma/backend/mlir/hardware_module.py
```diff
@@ -613,6 +620,39 @@ def make_register(T, data, init, result):
...
+        try:
+            self._ctx.parent.xmr_paths[xmr]
+        except KeyError:
+            pass
+        else:
+            return True
+        paths = get_xmr_paths(self._ctx, xmr)
+        assert len(paths) == len(module.operands)
+        self._ctx.parent.xmr_paths[xmr] = paths
+        return True
```
leonardt, line 638: Could we just move the code inside KeyError?
```suggestion
        try:
            self._ctx.parent.xmr_paths[xmr]
        except KeyError:
            paths = get_xmr_paths(self._ctx, xmr)
            assert len(paths) == len(module.operands)
            self._ctx.parent.xmr_paths[xmr] = paths
        return True
```

magma/backend/mlir/xmr_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        leaves = _get_leaf_descendants(tree, xmr.port)
+        leaves = sorted(leaves, key=lambda n: tree.nodes[n]["index"])
+        return [
+            (_path_to_string(_get_path(tree, leaf), "_"),)
+            for leaf in leaves
+        ]
+    # (2)
+    leaves = list(_get_leaf_descendants(tree, root, include_self=True))
+    leaf, *path = _ascend_to_leaf(xmr.port, leaves)
+    path = _path_to_string(_get_path(tree, leaf), "_") + "".join(path)
```
leonardt, line 83: Seems like we could use a helper here for repeated pattern.
```
def _make_path(tree, value):
    return _path_to_string(_get_path(tree, value), "_")
```

leonardt (approved):
Minor comments

magma/backend/mlir/hardware_module.py
rsetaluri: semantically, yes I think that would be fine. Just a style preference to do it outside, with an early return in the no-exception case.

PR #1102, opened 2022-07-12
[MLIR] Add support for unconnected module

rsetaluri:
Adds support for the case where a module definition has no connections.
This could happen, for example, where the module is simply a container
for undriven instances (as in the added test).

### DIFF
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -14,6 +14,7 @@ from magma.backend.mlir.builtin import builtin
 from magma.backend.mlir.comb import comb
 from magma.backend.mlir.common import wrap_with_not_implemented_error
 from magma.backend.mlir.compile_to_mlir_opts import CompileToMlirOpts
+from magma.backend.mlir.errors import MlirCompilerInternalError
 from magma.backend.mlir.graph_lib import Graph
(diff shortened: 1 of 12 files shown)


magma/backend/mlir/hardware_module.py
```diff
@@ -789,20 +789,25 @@ def visit_module(self, module: ModuleWrapper) -> bool:
...
+        for src, _, data in self._graph.in_edges(magma_module, data=True):
             info = data["info"]
             src_port, dst_port = info["src"], info["dst"]
             src_value = self._ctx.get_or_make_mapped_value(src_port)
             self._ctx.set_mapped_value(dst_port, src_value)
-        assert self.visit_module(ModuleWrapper.make(module, self._ctx))
+
+    def visit(self, module: MagmaModuleLike):
+        if module in self._visited:
+            raise RuntimeError(f"Can not re-visit module")
```
leonardt, line 807: Would this be an internal error? If so, perhaps we should make it an assertion rather than a runtime error.  That way if a user encounters it it's less likely that they think it's their own fault.

rsetaluri: Changed it to raise a specific internal error type. I think we should use this pattern for other parts as well, to distinguish external errors from internal errors.

PR #1104, opened 2022-07-18
Add CreditValid type

leonardt:


rsetaluri:
I'm a bit skeptical on adding a whole new type just for the purpose of a name alias. It adds a new set of code paths that need to be tested and maintained (even if base code is shared), and is goes against the "one way to do things" principle. Unless there's a functional difference between the two types, I would vote against adding it

leonardt:
Per George, there's a suite of standard library modules that use a credit valid interface, so it makes sense to have it as a standard type even if it's nominal so that the reusable components can be standardized.  We could potentially move it to mantle?

leonardt:
It's similar to how Irrevocable/Decoupled are equivalent types (in this case, both structurally and nominally) but semantically different (there's a different set of assertions that can be applied to the types).  CreditValid is similar except that it also differs nominally, but more useful is that there's a different set of properties for type.  For example, a user might leverage automated coverage properties of all creditvalid types (which are different than standard ready valid types)

leonardt:
There's probably some clever way to refactor this so that there's less internal code to maintain

magma/types/handshake.py
```diff
@@ -36,14 +36,101 @@ def _maybe_add_data(cls, fields, T, qualifier):
...
+        ns['__qualname__'] = name
+
+        # _cache_handler implemented in hwtypes.adt_meta
+        return cls._cache_handler(True, fields, name, bases, ns)
+
+    def __new__(mcs, name, bases, namespace, fields=None, **kwargs):
+        result = super().__new__(mcs, name, bases, namespace, fields, **kwargs)
+
+        if not result._field_names_:
+            # Skip for base class (variants already exist and pointer is not
```
rsetaluri, line 105: have a bunch of other nits on comments. at least, can we end comments in periods that are full sentences? otherwise if they're short include them inline, e.g. on L112, that can be after the if or then inline, e.g. `return result  # these don't need variants`

magma/types/handshake.py
```diff
@@ -36,14 +36,101 @@ def _maybe_add_data(cls, fields, T, qualifier):
...
+        ns = {}
+        ns['__module__'] = cls.__module__
+        name = f'{cls}[{T}]'
+        ns['__qualname__'] = name
+
+        # _cache_handler implemented in hwtypes.adt_meta
+        return cls._cache_handler(True, fields, name, bases, ns)
+
+    def __new__(mcs, name, bases, namespace, fields=None, **kwargs):
+        result = super().__new__(mcs, name, bases, namespace, fields, **kwargs)
```
rsetaluri, line 102: nit: name `cls` instead of `result`

rsetaluri (approved):
bunch of modifications requested. good after that

PR #1105, opened 2022-07-18
Add naming benchmark

leonardt:


leonardt:
I believe this has to do with the insert coreir wires logic flattening complex types and inserting instances for each leaf, I was thinking we may consider an alternative strategy for MLIR

PR #1110, opened 2022-08-05
[When] Detect latches

leonardt:
Adds a conservative latch detection check.

Note, we do this on the generated MLIR because it has a regular syntax structure we can use to do the analysis as a standard recursive tree traversal.  Doing this at the magma level is difficult for a few reasons: 
1. It's difficult to do at when context enter/exit time because we have optional elsewhen/otherwise statements that make it difficult for us to determine when to run the analysis (i.e. when the chain has finished).  
2. It's difficult to do at definition close time because the representation used internally does not immediately map to the original when structure (which is useful for doing the standard recursive tree traversal approach).  The next step would be to convert the IR into the original when structure to do the analysis, but that's what the MLIR transformation does, so then it seems simpler to do on the MLIR representation since we already have the desired structure at that time.

Finally, this is likely to be a generally useful pass for the MLIR SV dialect, so the long term plan may be to migrate this logic into MLIR for more general reuse/improvement.

### DIFF
--- a/magma/backend/mlir/check_latches.py
+++ b/magma/backend/mlir/check_latches.py
@@ -10,7 +10,7 @@ class LatchError(Exception):
         latches_str = ", ".join(latch.debug_name for latch in latches)
         defn_str = str(ctx.magma_defn_or_decl)
         msg = (f"Error while compiling {defn_str}, "
-               f"detected latches: [{latches_str}]")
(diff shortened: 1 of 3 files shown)


rsetaluri:
Closed by merging `when-detech-latches` -> `when-latest`.

PR #1121, opened 2022-08-25
[When] Fix bug with lazy array value creation

leonardt:
Lazily created values should inherit the when context of its parent
(i.e. the active context when the original value was created), rather
than getting the currently active when context (problematic if the value
is referenced in a different context than the original parent value's
context)

### DIFF
--- a/magma/wire_container.py
+++ b/magma/wire_container.py
@@ -170,12 +170,10 @@ class Wire:
 class Wireable:
     def __init__(self):
         self._wire = Wire(self)
-        if isinstance(self.name, ArrayRef):
-            # Lazily constructed children should inherit the parent value's
-            # context, TODO(leonardt): update for Tuple2 when merged
-            self._when_context = self.name.array._when_context
-        else:
-            self._when_context = get_curr_when_block()
+        self._when_context = get_curr_when_block()
+
+    def set_when_context(self, ctx):
+        self._when_context = ctx
 
     def wired(self):
         return self._wire.wired()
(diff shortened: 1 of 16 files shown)


magma/wire_container.py
```diff
@@ -169,7 +170,12 @@ def bit(self):
 class Wireable:
     def __init__(self):
         self._wire = Wire(self)
-        self._when_context = get_curr_when_block()
+        if isinstance(self.name, ArrayRef):
```
rsetaluri, line 173: I think the derived class (Array) should handle this rather than the parent class (it's kind of backwards). I would propose exposing a (private) method `_set_when_context()` and when Array creates it's children, it should `_set_when_context(self._when_context)`. The main reason I prefer it this way is that the laziness of the array should be fully encapsulated (as much as possible) to the internals of Array, so no other functions or classes should really have to consider that implementation detail.

There is still the problem (either the way you have it here or the way I propose^ that further nested types don't get the appropriate context.

leonardt: For dealing with nesting, we can add a set_when_context API. For arrays, it sets the wireable when context which propagates it to children when they're created.  For tuples, it eagerly calls it on the children (will change in Tuple2, which will share the same code with Array).  For other types, they are leafs that will inherit the standard set_when_context API from Wireable.

leonardt:
In adding a test for protocol children, there were a couple protocol APIs needed to be added for compilation support

magma/wire_container.py
```diff
@@ -4,6 +4,7 @@
 from magma.debug import debug_wire
 from magma.logging import root_logger, StagedLogRecord
 from magma.when import get_curr_block as get_curr_when_block
+from magma.ref import ArrayRef
```
rsetaluri, line 7: looks like a spurrious import
```suggestion
```

rsetaluri (approved):

PR #1122, opened 2022-08-25
[when] Fixes bug in bulk wire resolution for array

leonardt:
Related to https://github.com/phanrahan/magma/pull/1121, when array
children are wired during bulk wire resolution, they should skip the
when context check because the conditional wiring logic was already
handled during the origin bulk wire and the children should
unconditionally use the current wiring.


leonardt:
Note, based on https://github.com/phanrahan/magma/pull/1121

PR #1131, opened 2022-09-09
[When] Temporarily hotfix lazy arrays

leonardt:
Until we fix the issue with bulk wire resolution, we can unblock current users by always wiring the children (effectively forcing bulk wire resolution before entering the when logic).

### DIFF
--- a/magma/array.py
+++ b/magma/array.py
@@ -602,12 +602,8 @@ class Array(Type, Wireable, metaclass=ArrayMeta):
         for i, child in self._enumerate_children():
             child.wire(o[i], debug_info)
 
-    @debug_wire
-    def wire(self, o, debug_info):
-        o = magma_value(o)
-        if not self._check_wireable(o, debug_info):
-            return
-        if (
+    def _should_wire_children(self, o):
+        return (
             self._has_elaborated_children() or
             o._has_elaborated_children() or
             self.T.is_mixed() or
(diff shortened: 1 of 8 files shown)


magma/array.py
```diff
@@ -606,9 +607,15 @@ def wire(self, o, debug_info):
         o = magma_value(o)
         if not self._check_wireable(o, debug_info):
             return
-        if (self._has_elaborated_children() or
-                o._has_elaborated_children() or
-                self.T.is_mixed()):
+        if (
```
rsetaluri, line 610: let's extract this into a temp named bool, will be more readable. perhaps `should_wire_children`

magma/array.py
```diff
@@ -606,9 +607,15 @@ def wire(self, o, debug_info):
...
+            self._has_elaborated_children() or
+            o._has_elaborated_children() or
+            self.T.is_mixed() or
+            # TODO(leonardt): temporary fix for when issue with bulk wire
+            # resolution, if we have a conditional wire let the children handle
+            # it for now
```
rsetaluri, line 616: nit
```suggestion
            # TODO(leonardt): This is a temporary fix for when with bulk wire
            # resolution. The workaround is if we have a conditional wire, we just
            # let the children handle it.
```

rsetaluri (approved):

PR #1134, opened 2022-09-13
[MLIR] Revert to CIRCT verison (hash) 2eea3019

rsetaluri:
* Revert hw.array_slice syntax to [] vs at
* Remove circt-opt loweringOptions argument
* Revert golds


leonardt (approved):

rsetaluri:
@leonardt we should merge this without squashing whenever we do so as to keep `5c71593` around for reverting.

PR #1139, opened 2022-09-14
[Primitives] Fix slice primitives

leonardt:
* Promote int-like values in get_slice/set_slice
* Update doc to refer to set_index instead of set_bit
* Support set_slice of length 1 for constant values


magma/primitives/slice.py
```diff
@@ -40,15 +42,32 @@ def slice(*args, **kwargs):
...
+        if width != 1:
+            raise ValueError(
+                "Width must equal 1 for set_slice with value of length 1")
+        if not (
+            isinstance(value, type(target)) or
+            isinstance(target, type(value))
+        ):
+            raise ValueError(
+                "Incompatible types found for target and value")
+        return value
```
leonardt, line 67: We could allow a non constant value, but that would be "unsafe" if start was ever non zero

docs/primitives.md
```diff
@@ -249,23 +249,26 @@ Slice function interfaces:
    * width: constant slice width
 
 
-### set_bit
+### set_index
```
rsetaluri, line 252: I would be in favor of renaming the function (publicly, i.e. in the docs) to `set_bit`. We could deprecate `set_index` for now with warning (doubt anyone is using it anyway...)

rsetaluri:
Moving this conversation here:

> I'm not sure if set_slice of length makes sense unless the start value is a constant 0. Otherwise, the smallest start value we could have is a Bit which could take on value 1 or 0. A start index of 1 would be invalid for setting a slice on length 1.

I think my note was unclear. I meant the case when `width` is 1, i.e. `value` is of length 1 -- not `target` being length 1. This seems like it should be uniform with any other `width` value. Right now it seems to be broken.

docs/primitives.md
leonardt: It was changed to set_index because the pattern is more general than just set_bit (i.e. you can use it on generic arrays, rather than just on Bits), so I think it makes more sense to keep it as set_index for generality

rsetaluri: I see. In that case, why do we need a separate function? I.e. why is it not covered by set slice with width 1?

rsetaluri: Makes sense -- but I would argue `x[1]` and `x[1:2]` is the same interface with different arguments (in fact it literally is the function `__getitem__`).

Maybe we don't need to do this in this change, but seems reasonable to refactor to have the following interface:

```python
"""
* target: Array[N, T]
* value: Union[Array[M, T], T]
* start: UInt[W]

where M < N and W <= clog2(N)
"""
def set(target: Array, value: Union[Type, Array], start: UInt):
    ...
```

Now that I think about it there are few new concerns here:
* Should this function work with all `Array`s and not just `Bits`? I kind of assumed it did, but looking at the code closer I'm not certain.
* Do we need a `width` argument? Can it just be inferred from `value`? If `value` is an array of the same dimensionality as `target` then it is a slice set; else if it is of dimensionality one less than `value` then it is an index set (of course assuming the contained types match up and all).
* In the case `start` being a constant value, should we optimize for the simple wiring (non-dynamic mux) case?

PR #1140, opened 2022-09-15
Add instance callback pass

rsetaluri:
Adds the `InstanceCallbackPass`, called by default in `m.compile()`. It can be defined as follows on an instance:

```python
def callback(path, data):
    log = data.setdefault("log", list())
    log.append(f"Visited inst '{path[-1]}' at hierarchical path '{path}'")


class Foo(m.Circuit):
    ...
    bar = Bar(name="bar")
    baz = Baz(name="baz")
    m.register_instance_callback(bar, callback)
    m.register_instance_callback(baz, callback)
    ...


class Top(m.Circuit):
    ...
    foo = Foo()
    ...


res = m.compile("top", Top, ...)
callback_log = res["instance_callback_pass_data"]["log"]
# This should print:
#   Visited inst 'bar' at hierarchical path '[Top, Foo_inst0, bar]'
#   Visited inst 'baz' at hierarchical path '[Top, Foo_inst0, baz]'
for entry in callback_log:
    print (entry)
```

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -8,6 +8,7 @@ import functools
 import operator
 from collections import namedtuple
 import os
+from typing import Callable, Dict, List
 
 import six
 from . import cache_definition
@@ -55,6 +56,7 @@ __all__ += ['CopyInstance']
 __all__ += ['circuit_type_method']
 __all__ += ['circuit_generator']
 __all__ += ['CircuitBuilder', 'builder_method']
+__all__ += ['register_instance_callback', 'get_instance_callback']
 
 circuit_type_method = namedtuple('circuit_type_method', ['name', 'definition'])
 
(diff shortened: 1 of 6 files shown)


magma/circuit.py
```diff
@@ -921,3 +923,32 @@ def __new__(metacls, name, bases, dct):
...
+        callback: InstanceCallback,
+):
+    try:
+        get_instance_callback(instance)
+    except AttributeError:
+        pass
+    else:
+        raise AttributeError(
+            f"Instance {instance} already has callback registered"
+        )
```
leonardt, line 942:
```suggestion
    if hasattr(instance, "_callback_"):
        raise AttributeError(
            f"Instance {instance} already has callback registered"
        )
```

Perhaps a `hasattr` would be better here? Less code and I don't think going with an eager try/catch provides any performance benefit since we still have to check the attribute anyways.

magma/circuit.py
```diff
@@ -921,3 +923,32 @@ def __new__(metacls, name, bases, dct):
...
+    setattr(instance, "_callback_", callback)
+
+
+def get_instance_callback(instance: CircuitType) -> InstanceCallback:
+    try:
+        return getattr(instance, "_callback_")
```
leonardt, line 948:
```suggestion
        return instance._callback_"
```

Could skip getattr since we're catching the AttributeError anyways

leonardt (approved):

magma/circuit.py
rsetaluri: :+1:

magma/circuit.py
rsetaluri: :+1:

PR #1142, opened 2022-09-20
[MLIR] Add ifdef inside compile guarded modules

rsetaluri:
This is a workaround for #1143.

Instead of generating

```verilog
module Top;
    `ifdef COND
    !! some invalid verilog !!
    `endif
endmodule;
```

this change will generate

```verilog
module Top;
   `ifdef COND
   CompileGuard0 CompileGuard0_inst0 (...);
   `endif
endmodule;

module CompileGuard0 (...);
   `ifdef COND
    !! some invalid verilog !!
   `endif
endmodule;
```

Note that rather than inlining the compile guarded definition, we simply "reguard" it with the same condition.

### DIFF
--- a/magma/compile_guard.py
+++ b/magma/compile_guard.py
@@ -1,10 +1,10 @@
 import contextlib
 import itertools
-from typing import Optional, Tuple
+from typing import Dict, Optional, Tuple, Union
 
 from magma.bits import BitsMeta
 from magma.clock import ClockTypes
-from magma.circuit import CircuitBuilder
+from magma.circuit import CircuitKind, AnonymousCircuitType, CircuitBuilder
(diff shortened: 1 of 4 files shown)


magma/compile_guard.py
```diff
@@ -95,6 +94,21 @@ def _new_port_name(self) -> str:
         return f"port_{next(self._num_ports)}"
 
 
+def get_compile_guard_data(
+        inst_or_defn: Union[AnonymousCircuitType, CircuitKind]
+) -> Optional[Dict]:
```
leonardt, line 99: Could consider moving this API to the circuit objects and have them implement it.

leonardt (approved):
Minor comments

Issue #1132, opened 2022-09-09
[When] Inferred latches incorrectly detected for array slices


rsetaluri:
```python
import magma as m


class _(m.Circuit):
    T = m.Bits[8]
    t_i = m.Out(T)(name="t_i")
    t_o = m.In(T)(name="t_o")
    S = m.Bits[2](name="S")

    with m.when(S[0]):
        t_o @= t_i
    with m.otherwise():
        t_o[:4] @= t_i[4:]
        t_o[4:] @= t_i[4:]
```

I have a feeling this is a tough one to detect...Or the latch methodology may have to consider the flattened values.

leonardt:
This should be handled by the unwire,rewire fix. But, yet basically if there's overlaps we have to check flattened values, that's why the hot fix just forces flattening for arrays.

PR #1144, opened 2022-09-20
[When] Lazy type integration

### LINKED ISSUE #1125: [When] Mixing array bulk and elaborated wiring for arrays results in error
```python
import magma as m


class Foo(m.Circuit):
    io = m.IO(I=m.In(m.SInt[2]), S=m.In(m.Bit), O=m.Out(m.SInt[2]))
    with m.when(io.S):
        io.O @= io.I
    with m.otherwise():
        io.O @= m.sint(m.uint(io.I) >> 1)


m.compile("build/Foo", Foo, output="mlir-verilog")
```

produces

```verilog
module Foo(	// <stdin>:1:1
  input  [1:0] I,
  input        S,
  output [1:0] O);

  reg [1:0] _GEN;	// <stdin>:8:10
  reg       _GEN_0;	// <stdin>:10:11
  reg       _GEN_1;	// <stdin>:12:11

  always_comb begin	// <stdin>:14:5
    _GEN_0 = _GEN[0];	// <stdin>:2:10, :9:10, :15:9
    _GEN_1 = _GEN[1];	// <stdin>:6:10, :9:10, :16:9
    if (S)	// <stdin>:17:9
      _GEN = I;	// <stdin>:18:13
    else begin	// <stdin>:17:9
      _GEN_0 = I[1];	// <stdin>:4:10, :20:13
      _GEN_1 = 1'h0;	// <stdin>:7:10, :21:13
    end
  end // always_comb
  assign O = {_GEN_1, _GEN_0};	// <stdin>:11:10, :13:10, :24:11, :25:5
endmodule
```

Notice that `_GEN` is used in the first case and `_GEN_0` and `_GEN_1` are used in the second.  We need to update the when context and conditional driver when children are elaborated.

leonardt:
Updates the array2 logic to properly handle conditionally driven values.  The main change requires that we ensure that children are "rewired" based on the conditionally driven information of the parent.  That is, if a parent was conditional wired in various when contexts, the children need to be wired in the same when contexts using the original drivers.  This recursively invokes the elaboration process on the drivees and drivers, maintaining consistency in the when contexts (e.g. for latch detection with children and slices).  In doing this, I discovered a few code paths that could share this same resolution logic for children (_resolve_driven_bulk_wire), which simplifies some of the more complicated parts of the array code.  

One important choice was how elaboration was handled with the when build primitive.  When a value is elaborated, it is removed from the when builder outputs, and replaced with its children.  This requires us to add the ability to remove a port from a builder, something we didn't have before.  The other option was to leave the port (since it was unwired and not used), but this required adding logic in various places to be aware of this (e.g. latch detection would need to know that this wire was replaced by the child wires).  This was also a problem related to input values that are elaborated, because it threw off the indexing logic.  Related to removing a port, we also needed to rename ports.  We could choose not do this, and merely update the value_to_index dictionaries, but I thought this would be useful for debugging consistency.  Again, we could choose to special case this logic in the backend, but by adding a port removal API, the downstream backend logic can remain unchanged and just assume the ports are consistent, which requires less changes in the code.

Another decision was to have Wireable objects track their "wired_when_contexts" and updating them when they are elaborated.  Another option was to have the root when context handle the updates, which seemed like it might be cleaner since it contains the when logic to the when code objects, but that trades off the performance cost of having to traverse the entire when tree to determine which contexts a value was assigned in.  It's nominal since the user is unlikely to have an extreme amount of when statements, but it seemed simpler to just track, and update the contexts of interests (e.g. a value could have default and only be assigned in one when contexts, so using the root context could require traversing a large number of unnecessary contexts).

An elaboration pass was added for tuples so it can occur before the when finalize logic, which was also moved to the backend versus at circuit definition time.  This is for various reasons, one: the user may open and edit a definition which could require changes to the when context either from elaboration or new when statements, two: the flatten tuple logic in the backend was causing elaboration after when had been finalized, so we want to do this earlier so that we can then finalize the when statements (which was easier done as a set of two passes, rather than doing this in the backend logic)

Fixes https://github.com/phanrahan/magma/issues/1132
Fixes #1125

### DIFF
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -560,7 +560,7 @@ class ModuleVisitor:
             # `value` occurs as more than one input (for example, it could be
             # used as a conditional driver for multiple values).  We could
             # optimize the logic to always share the same argument, but for now
-            # we just use the "last" index
+            # we just use the "last" index.
             value_to_index[value] = next(counter)
 
         counter = itertools.count()
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -555,7 +555,6 @@ class ModuleVisitor:
         output_to_index = {}
 
         def _visit(value, counter, value_to_index):
-            assert value not in value_to_index
             value_to_index[value] = next(counter)
 
         counter = itertools.count()
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -555,7 +555,12 @@ class ModuleVisitor:
         output_to_index = {}
 
         def _visit(value, counter, value_to_index):
-            assert value not in value_to_index
+            # NOTE: value may already be in value_to_index, in which case we
+            # update it to a new index.  This is okay and it just means that
+            # `value` occurs as more than one input (for example, it could be
+            # used as a conditional driver for multiple values).  We could
+            # optimize the logic to always share the same argument, but for now
+            # we just use the "last" index
             value_to_index[value] = next(counter)
 
         counter = itertools.count()
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -555,10 +555,10 @@ class ModuleVisitor:
         output_to_index = {}
 
         def _visit(value, counter, value_to_index):
-            # NOTE(leonardt): value may already be in value_to_index, in which case we
-            # update it to a new index.  This is okay and it just means that
-            # `value` occurs as more than one input (for example, it could be
-            # used as a conditional driver for multiple values).  We could
+            # NOTE(leonardt): value may already be in value_to_index, in which
+            # case we update it to a new index.  This is okay and it just means
+            # that `value` occurs as more than one input (for example, it could
+            # be used as a conditional driver for multiple values).  We could
             # optimize the logic to always share the same argument, but for now
             # we just use the "last" index.
             value_to_index[value] = next(counter)
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -555,6 +555,7 @@ class ModuleVisitor:
         output_to_index = {}
 
         def _visit(value, counter, value_to_index):
+            assert value not in value_to_index
             value_to_index[value] = next(counter)
 
         counter = itertools.count()
--- a/magma/backend/mlir/mlir_compiler.py
+++ b/magma/backend/mlir/mlir_compiler.py
@@ -39,11 +39,11 @@ class MlirCompiler(Compiler):
         return "mlir"
 
     def _run_passes(self):
-        # NOTE(leonardt): We elaborate tuples before finalizing when statements 
-        # because they may cause array expansion (e.g. when we have an array of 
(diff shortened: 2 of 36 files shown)


leonardt:
Actually it turns out trying to remove the drivee is problematic when the drivee is being used multiple times in the builder (this happens when it's used in multiple different when contexts), so I think I'm going to update the logic to keep the discarded drivees and filter out the undriven ones during codegen.

magma/backend/mlir/hardware_module.py
```diff
@@ -555,7 +555,12 @@ def visit_when(self, module: ModuleWrapper) -> bool:
         output_to_index = {}
 
         def _visit(value, counter, value_to_index):
-            assert value not in value_to_index
+            # NOTE: value may already be in value_to_index, in which case we
```
rsetaluri, line 558: nit
```suggestion
            # NOTE(leonardt): value may already be in value_to_index, in which case we
```

magma/backend/mlir/mlir_compiler.py
```diff
@@ -37,6 +39,14 @@ def suffix(self):
         return "mlir"
 
     def _run_passes(self):
+        # We elaborate_tuples before finalizing when statements because they
+        # may cause array expansion (e.g. when we have an array of tuples).  If
+        # we remove the flatten_all_tuples requirement, we may finalize when
+        # earlier, but then we may still need this code path as an option
```
rsetaluri, line 45:
```suggestion
        # NOTE(leonardt): We elaborate tuples before finalizing when statements 
        # because they may cause array expansion (e.g. when we have an array of 
        # tuples).  If we remove the flatten_all_tuples requirement, we may     
        # finalize when earlier, but then we may still need this code path as an                                                                               
        # option.                                                               
```

magma/definition_context.py
```diff
@@ -123,6 +123,10 @@ def set_default_metadata(self, key: str, default: Any) -> Any:
     def place_instances(self, defn):
         self._placer = self._placer.finalize(defn)
         for builder in self._builders:
+            if getattr(builder, "_is_when_builder_", False):
```
rsetaluri, line 126: can we hide this in an API like `is_when_builder()` in `primitives/when.py`?

rsetaluri (changes requested):

magma/circuit.py
```diff
@@ -901,6 +901,9 @@ def finalize(self, dont_instantiate: bool = False):
             setattr(inst, k, v)
         return inst
 
+    def is_when_builder(self):
```
rsetaluri, line 904: I think we should use a free function in `primitives/when.py`. it might result in a circular import, but prefer that over adding another method to circuit.

magma/definition_context.py
```diff
@@ -123,6 +123,10 @@ def set_default_metadata(self, key: str, default: Any) -> Any:
     def place_instances(self, defn):
         self._placer = self._placer.finalize(defn)
         for builder in self._builders:
+            if builder.is_when_builder():
+                # Wait to finalize when primitive until compile since circuit
+                # modifications may introduce changes
```
rsetaluri, line 128: nit
```suggestion
                # NOTE(leonardt): We have to wait to finalize the when primitive
                # until compilation to consider downstream circuit modifications.
```

rsetaluri (approved):
Final change request. Looks good otherwise.

PR #1146, opened 2022-09-29
[MLIR] Revert wrapping compile guarded modules in ifdef

rsetaluri:
Relevant PR:
  #1142

Relevant commits:
  bb95301e3fd2ade92f73d358ffb18ea6172d4f4c
  1133049bcc8ef74617c2fd7e0bacbf12a49e91b3


rsetaluri:
@leonardt note that this was a hack to begin with, and we no longer need the hack. Reverting it now to avoid it sticking around.

leonardt (approved):

PR #1149, opened 2022-10-04
[MLIR] Add option to disallow symbols with duplicate names

rsetaluri:


leonardt (approved):
LGTM, if I understand his correctly, MLIR doesn't care about the duplicate names and just appends a counter value to make it unique in the verilog.

PR #1150, opened 2022-10-04
Prototype Sum

leonardt:
Checking in a prototype implementation of the sum type, this isn't quite ready for widespread use, but I would like to it to be in the CI flow since it depends on the when logic.  Also, I think it's been sufficiently pushed through to get a review on the overall architecture and make some early decisions, e.g.:
(1) match, case context manager patterns (dispatching to when)
(2) stateful interpretation of the sum value based on the active context (the alternative is to use a named reference, e.g. `io.I.x`, to retrieve the interpreted value, but we'd still want to check the active context to verify that it's a legal reference, so if we're maintaining that state anyways it seems useful to the user to just have it interpreted correctly for them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,4 @@ prof
 
 *.sublime-*
 benchmarks/*stats
+*bind_files.list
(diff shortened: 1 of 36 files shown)


leonardt:
Closing in favor of https://github.com/phanrahan/magma/pull/1367

PR #1154, opened 2022-10-12
[Primitives] Add default=None for list_lookup

leonardt:
Rather than use an explicit default value as the standard pattern, it makes more sense to use the final element of the list since we're buildling a mux tree anyways.  The existing explicit "default" value pattern might still be useful to emulate a certain "else" pattern that's different than the last element.

In adding this, I noticed the mux tree was being built forwards, which made the generated code a bit convoluted for this default case, so I reversed the tree instead to make it look more natural in the generated code.

### DIFF
--- a/magma/primitives/mux.py
+++ b/magma/primitives/mux.py
@@ -185,15 +185,18 @@ def dict_lookup(dict_, select, default=0):
     return output
 
 
-def list_lookup(list_, select, default=0):
+def list_lookup(list_, select, default=None):
     """
     Use `select` as an index into `list` (similar to a case statement)
 
     `default` is used when `select` does not match any of the indices (e.g.
-    when the select width is longer than the list) and has a default value of
-    0.
+    when the select width is longer than the list).  If it is `None`, the last
+    element of the list will be used.
     """
     output = default
-    for i, elem in enumerate(list_):
-        output = mux([output, elem], i == select)
+    if default is None:
+        output = list_[-1]
+        list_ = list_[:-1]
+    for i in range(len(list_) - 1, -1, -1):
+        output = mux([output, list_[i]], i == select)
     return output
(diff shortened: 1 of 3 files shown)


magma/primitives/mux.py
```diff
@@ -185,15 +185,18 @@ def dict_lookup(dict_, select, default=0):
...
+    when the select width is longer than the list).  If it is `None`, the last
+    element of the list will be used.
     """
     output = default
-    for i, elem in enumerate(list_):
-        output = mux([output, elem], i == select)
+    if default is None:
+        output = list_[-1]
+        list_ = list_[:-1]
+    for i in range(len(list_) - 1, -1, -1):
```
rsetaluri, line 200: why do we need to change the order?

rsetaluri: oh nvm I see the comment in the PR.

rsetaluri, line 200: Can we do

```python
for i, elt in reversed(enumerate(list_)):
```
?

rsetaluri: and also add an inline comment here like

```
# We chain the muxes in reverse order so that the emitted Verilog is in forward order.
```

leonardt: > ```python
> for i, elt in reversed(enumerate(list_)):
> ```

I tried this originally and enumerate objects are not reversible (presumably because reversing them would require expanding them entirely).

```
>>> for i, elt in reversed(enumerate(list_)):
...
KeyboardInterrupt
>>> list_ = [x for x in range(9)]
>>> for i, elt in reversed(enumerate(list_)):
...     print(i, elt)
...
  File "<stdin>", line 1, in <module>
TypeError: 'enumerate' object is not reversible
```

PR #1156, opened 2022-10-14
[MLIR] Finalize whens last

leonardt:
Then when finalize must occur after any passes that modify the circuit

e.g. `insert_coreir_wires` could flatten a value used in the conditional logic, which then requires updating the when builder instance

### DIFF
--- a/magma/backend/mlir/mlir_compiler.py
+++ b/magma/backend/mlir/mlir_compiler.py
@@ -45,7 +45,6 @@ class MlirCompiler(Compiler):
         insert_coreir_wires(self.main)
         insert_wrap_casts(self.main)
         wire_clocks(self.main)
-
         # NOTE(leonardt): finalizing whens must happen after any
         # passes that modify the circuit.  This is because passes
         # could introduce more conditional logic, or they could
--- a/magma/backend/mlir/mlir_compiler.py
+++ b/magma/backend/mlir/mlir_compiler.py
@@ -39,18 +39,19 @@ class MlirCompiler(Compiler):
         return "mlir"
 
     def _run_passes(self):
-        # NOTE(leonardt): We elaborate tuples before finalizing when statements
-        # because they may cause array expansion (e.g. when we have an array of
-        # tuples).  If we remove the flatten_all_tuples requirement, we may
-        # finalize when earlier, but then we may still need this code path as
-        # an option.
         if self.opts.get("flatten_all_tuples", False):
             elaborate_tuples(self.main)
-        finalize_whens(self.main)
 
         insert_coreir_wires(self.main)
         insert_wrap_casts(self.main)
         wire_clocks(self.main)
+
+        # NOTE(leonardt): finalizing whens must happen after any
+        # passes that modify the circuit.  This is because passes
+        # could introduce more conditional logic, or they could
+        # trigger elaboration on values used in existing coditiona
+        # logic (which modifies the when builder)
+        finalize_whens(self.main)
         raise_logs_as_exceptions_pass(self.main)
 
     def compile(self):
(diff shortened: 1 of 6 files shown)


tests/test_when.py
```diff
@@ -485,6 +485,7 @@ class _Test(m.Circuit):
     assert check_gold(__file__, f"{basename}.mlir")
 
 
+@pytest.mark.skip("There's a bug in this test")
```
leonardt, line 488: Turns out there's a bug in this test, `x` is defined inside a when, and then referred to in the following otherwise, I'm not sure what the original intention of this was, it might have been a temporary change while I was debugging.  Working on a fix but will put it in a separate PR for clarity.

magma/backend/mlir/mlir_compiler.py
```diff
@@ -39,18 +39,19 @@ def suffix(self):
...
-        # finalize when earlier, but then we may still need this code path as
-        # an option.
         if self.opts.get("flatten_all_tuples", False):
             elaborate_tuples(self.main)
-        finalize_whens(self.main)
 
         insert_coreir_wires(self.main)
         insert_wrap_casts(self.main)
         wire_clocks(self.main)
+
```
rsetaluri, line 48: nit: remove white line

tests/test_backend/test_mlir/test_mlir_compiler.py
```diff
@@ -0,0 +1,17 @@
+import magma as m
+
+
+def test_mlir_compiler_finalize_when_order():
```
rsetaluri, line 4: let's move this to an example in `examples.py` as the other ones, maybe named `complex_when_lazy_tuple`

tests/test_when.py
rsetaluri, line 488: nit: let's make a more descriptive skip message, maybe `"Test does not pass due to when issue (#1156)"`

tests/test_backend/test_mlir/test_mlir_compiler.py
leonardt: I moved it to examples, but it looks like the existing examples flow doesn't use the mlir_compiler interface, so it doesn't test that pass ordering, so I left it in this separate test that uses `m.compile` for now

tests/test_backend/test_mlir/test_mlir_compiler.py
```diff
@@ -0,0 +1,7 @@
+import magma as m
+from examples import complex_when_lazy_tuple
+
+
+def test_mlir_compiler_finalize_when_order():
+    m.compile("build/complex_when_lazy_tuple", complex_when_lazy_tuple,
+              output="mlir")
```
rsetaluri, line 7: nit
```suggestion
    m.compile(
        "build/complex_when_lazy_tuple",
        complex_when_lazy_tuple,
        output="mlir",
    )
```

also let's use a tempdir and kill the build dir and .gitignore

tests/test_backend/test_mlir/test_mlir_compiler.py
```diff
@@ -0,0 +1,7 @@
+import magma as m
+from examples import complex_when_lazy_tuple
```
rsetaluri, line 2: nit
```suggestion

from examples import complex_when_lazy_tuple
```

tests/test_backend/test_mlir/test_utils.py
```diff
@@ -181,7 +181,7 @@ def get_local_examples() -> List[DefineCircuitKind]:
         examples.simple_memory_wrapper,
         examples.simple_undriven_instances,
         examples.simple_neg,
-        examples.simple_array_slice,
+        examples.simple_array_slice
```
rsetaluri, line 184: nit: leave `,`

rsetaluri (approved):
final comments, looks good if addressed

PR #1157, opened 2022-10-14
[When] Update when temp value scope

leonardt:
Updates the test so the temporary value is only referenced inside the scope it's created.

Updates the insert_coreir_wires logic to handle builders (needed for the when builder which hasn't been finalized).

I think this is probably a better pattern than allowing users to abuse the fact that the Python namespace allows values to fall through to the enclosing scope, otherwise we run into potential issues with latch inference (either we need to require the user to assign the value in all when contexts, or determine when a value is referenced outside it's context and change its enclosing scope).  While possible, I doubt there's many useful patterns that motivate this added complexity.

### DIFF
--- a/magma/backend/coreir/insert_coreir_wires.py
+++ b/magma/backend/coreir/insert_coreir_wires.py
@@ -119,6 +119,12 @@ class InsertCoreIRWires(DefinitionPass):
             for value in instance.interface.ports.values():
                 if not value.is_output():
                     self._insert_wire(value, definition)
+        for builder in definition._context_._builders:
+            if builder._finalized:
+                continue
+            for value in builder._io.inst_ports.values():
+                if not value.is_output():
+                    self._insert_wire(value, definition)
         for value in definition.interface.ports.values():
             if not value.is_output():
                 self._insert_wire(value, definition)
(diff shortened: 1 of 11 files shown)


magma/backend/coreir/insert_coreir_wires.py
```diff
@@ -119,6 +119,12 @@ def __call__(self, definition):
             for value in instance.interface.ports.values():
                 if not value.is_output():
                     self._insert_wire(value, definition)
+        for builder in definition._context_._builders:
```
rsetaluri, line 122: why do we do this? can we add a comment note for this?

magma/when.py
```diff
@@ -412,6 +412,11 @@ def _get_assignees_and_latches(ops: Iterable[_Op]) -> Tuple[Set, Set]:
                 )
             )
             assignees.update(then_assignees.union(else_assignees))
+            # Filter out values created in this context
```
rsetaluri, line 415:
```suggestion
            # Filter out any latches that were created in this when context.
```

rsetaluri (approved):

magma/backend/coreir/insert_coreir_wires.py
leonardt: Explained in the PR message, but i'll add a comment.

leonardt: More generally, we may want to reconsider when builders are finalized, and have a more unified API for iterating over instances/builders.  The motivation would be that passes may want to modify builders or even introduce them.  Right now we finalize builder when a definition closes, so I'm not sure it would work if a user opened a circuit then added another builder instance? We may want to finalize builders as a final step before compile (which provides a more open interface for circuit modification before compilation)

PR #1158, opened 2022-10-17
[MLIR] Support protocol types in module interface

rsetaluri:


tests/test_backend/test_mlir/golds/simple_compile_guard.v
```diff
@@ -2,27 +2,28 @@
 module COND1_compile_guard(	// <stdin>:1:1
   input port_0,
         CLK);
-  reg Register_inst0;	// <stdin>:2:10
```
rsetaluri, line 5: this file had some lingering errors and was never updated. not related to the main change

leonardt (approved):
Since these all seem related to the interface iterators, I wonder if it could be useful to have an "underlying_ports" API or something like that that's to be used by the backend to get the underlying magma values for protocols.

rsetaluri:
> Since these all seem related to the interface iterators, I wonder if it could be useful to have an "underlying_ports" API or something like that that's to be used by the backend to get the underlying magma values for protocols.

@leonardt good idea, see #1159.

PR #1163, opened 2022-10-18
[When] Add user extensible register pattern

leonardt:
I first tried to use the subclass m.Register pattern, but this ran into an issue with the m.Generator2 key logic and an unhashable dict from the metaclass arguments.  In the future, we may consider changing this to use a subclassing patterns, but I think we need to ensure that the Generator2 pattern is amenable to subclassing and that seems like a task for a separate PR.

A simpler approach is to use a `_magma_register_` magic flag that indicates values that should be treated as registers in the when logic. These values implement a `get_enable` API to support the default enable logic.


magma/primitives/when.py
```diff
@@ -76,17 +76,16 @@ def _get_corresponding_register_default(value: Type) -> Optional[Type]:
...
         value_inst = root_ref.inst
     except AttributeError:
         return None
-    if not isinstance(type(value_inst), Register):
+    if not type(value_inst)._magma_register_:
         return None
-    if isinstance(value, Enable):
-        return Enable(False)
+    if value is value_inst.get_enable():
+        return Bit(False)
```
leonardt, line 88: With the generic `get_enable API`, we have a more reliable way to find the enable values, but to make things more extensible (e.g. users that use a Bit enable type), we change the default wiring type to Bit.

magma/when.py
```diff
@@ -143,17 +131,27 @@ def _add_reg_enables(self):
...
         # Copy with list because this can add conditional wires
-        for wire in list(self._conditional_wires):
+        _conditional_wires = list(self._conditional_wires)
+
+        # Only wire enables once otherwise an elaborated register input value
+        # may trigger multiple wirings
+        # NOTE(leonardt): If we wanted to allow users to wire explicitly, we
+        # could lookup whether a wire to enable exists in _conditional_wires
+        # (which would also avoid multiple wirings)
+        seen = set()
```
leonardt, line 141: When running into this seen issue, I realized there is a simple way to allow users to explicitly wire enable if we wanted, but it still seems odd to support it (maybe they want to choose the wired value without updating for a case?)

leonardt: (Alternatively, we may decide to force the user to use the Enable type, and it's a simple update to the user code

leonardt:
Okay, I think I got a pattern I'm fairly happy with.  I added an AbstractRegister base class that the user can extend for the implicit when support.  The concrete m.Register is not viable for subclassing since it has special handling in the MLIR backend (e.g. the user can't give the subclass a special verilog code body definition).  There is the issue of "enforcing" the get_enable API.  The simplest way is to have the method on the Abstract generator class, but this makes calling it on the instance of the circuit a bit more tedious (have to call type and pass it the instance), but since this is done internally I think it's fairly reasonable, and seems better than option (2) which was to define the abstract method inside __init__ and patch it onto the generated class (which would also only work as an "abstract" method if the subclass called super().__init__ to cause the patching).  The current method will catch when the user forgets to implement the method, where option(2) the user could forget to call super.__init__ and then not get caught.

rsetaluri (commented):
I might prefer a registration pattern rather than a subclassing pattern.

So something where we maintain a list of register types, which by default holds our register. And then we can add `m.register_register_type()` (unfortunate name 😓, maybe `reg_type`).

leonardt:
Can you provide an argument for that over a class pattern? 

This seems reasonable to me given the abstract method and the isinstance check patterns.  That is, if we go with a "registration pattern" we have to maintain our own data structure to track this information, whereas the subclassing pattern the information is stored implicitly in the class hierarchy.  There would also no longer a contract to implement the get_enable interface beyond documentation.  I feel like this is a fairly standard use case of the class/subclass pattern? At a lower level, using a registration pattern requires us to add/maintain an API and internal data structure, whereas the class pattern has an implicit "API" (by providing the ability to subclass) and the information is stored implicitly (in the class hierarchy).

rsetaluri:
> Can you provide an argument for that over a class pattern?
> 
> This seems reasonable to me given the abstract method and the isinstance check patterns. That is, if we go with a "registration pattern" we have to maintain our own data structure to track this information, whereas the subclassing pattern the information is stored implicitly in the class hierarchy. There would also no longer a contract to implement the get_enable interface beyond documentation. I feel like this is a fairly standard use case of the class/subclass pattern? At a lower level, using a registration pattern requires us to add/maintain an API and internal data structure, whereas the class pattern has an implicit "API" (by providing the ability to subclass) and the information is stored implicitly (in the class hierarchy).

That's fair. I'm ok with this pattern. Made some updates and LGTM.

rsetaluri (approved):

PR #1166, opened 2022-10-21
[CompileGuard] Support simple anon drivees

leonardt:
For now, we only support anon values that are driving values with the same status: either all must be external or internal. We also assume that more drivees will not be added after the fact. We still do not support unwired anon values, nor do we support anon values that have wirings that change after the compile_guard (e.g. an internal value is then wired to an external value after the compile guard)

In the future, we could improve support by:
(1) staging the resolving of externality until the end of definition 
(2) copy anon values that drive mixed values (one copy for external, one for internal)
(3) do final wiring for internal/external copies with the proper values


tests/test_compile_guard.py
```diff
@@ -324,3 +324,39 @@ class Top(m.Circuit):
...
+            x.undriven()  # x is driven internally
+            out = m.Register(m.Bit)()(x)
+
+        io.O @= io.I
+
+    m.compile("build/test_compile_guard_anon_driven_internal", _Top,
+              output="mlir")
+    assert m.testing.check_files_equal(
+        __file__, "build/test_compile_guard_anon_driven_internal.mlir",
+        "gold/test_compile_guard_anon_driven_internal.mlir")
```
rsetaluri, line 345: nit
```suggestion
    basename = "test_compile_guard_anon_driven_internal"
    m.compile(f"build/{basename}", _Top, output="mlir")
    assert m.testing.check_files_equal(
        __file__, f"build/{basename}.mlir", f"gold/{basename}.mlir"
    )
```

tests/test_compile_guard.py
```diff
@@ -324,3 +324,39 @@ class Top(m.Circuit):
...
+        io.O @= x  # x drives an external value
+
+        with m.compile_guard("COND", defn_name="COND_compile_guard"):
+            x @= m.Register(m.Bit)()(io.I)
+
+    m.compile("build/test_compile_guard_anon_driving_external", _Top,
+              output="mlir")
+    assert m.testing.check_files_equal(
+        __file__, "build/test_compile_guard_anon_driving_external.mlir",
+        "gold/test_compile_guard_anon_driving_external.mlir")
```
rsetaluri, line 362: nit
```suggestion
    basename = "test_compile_guard_anon_driving_external"
    m.compile(f"build/{basename}", _Top, output="mlir")
    assert m.testing.check_files_equal(
        __file__, f"build/{basename}.mlir", f"gold/{basename}.mlir"
    )
```

tests/test_compile_guard.py
```diff
@@ -324,3 +324,39 @@ class Top(m.Circuit):
...
+        io.O @= io.I
+
+    m.compile("build/test_compile_guard_anon_driven_internal", _Top,
+              output="mlir")
+    assert m.testing.check_files_equal(
+        __file__, "build/test_compile_guard_anon_driven_internal.mlir",
+        "gold/test_compile_guard_anon_driven_internal.mlir")
+
+
+def test_compile_guard_anon_driving_external():
```
rsetaluri, line 348: I think should should not be valid magma. Either we should explicitly disallow driving outside values from within a compile guard *or* implicitly convert them to `compile_guard_select`.

(BTW the generated MLIR is not "valid" in that it contains undefined SSA values and therefore can not generate Verilog. But I think this should be solved in magma/compile_guard, *not* in the backend.)

leonardt: Okay, there was already an existing test that has compile guard driving an external value, so we should remove that and we can just raise an error when detecting externals

rsetaluri: Oh I see. Do we use this in customer code? In that case we would want to continue support and just fix the case that's not working.

leonardt: I doubt it now that I look at the existing test, one of them is incomplete (doesn't fully compile). Likely a WIP effort that wasn't completed, we can bring this up next week and see if anyone wants us to clean it up, but for now I'd rather not have it in half way.

rsetaluri (approved):

PR #1169, opened 2022-10-24
[When] Add support for explicitly wired CE

leonardt:
Two cases:
(1) wired beforehand, the user wants a specific constant value 
(2) wired inside, the user wants to override the implicit 1 value

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -4,6 +4,7 @@ from .t import Direction, In
 from .digital import DigitalMeta, Digital
 from magma.bit import Bit
 from magma.array import Array
+from magma.debug import get_debug_mode, get_callee_frame_info
 from magma.tuple import Tuple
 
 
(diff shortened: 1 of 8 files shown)


leonardt:
See https://github.com/phanrahan/magma/issues/1170 for an issue that should be resolved before merging this

leonardt:
The initial resolution for https://github.com/phanrahan/magma/issues/1170 will be that we do not add any special behavior for an already wired CE.  So wiring a CE will change the CE default driver (same as explicitly wiring the register default driver).  The current logic will implicitly insert a CE @= 1 and if the user desires a different behavior, they should wire it up after the when context (they can do .CE.rewire(desired_value)).  This is likely the least confusing choice for now since it chooses a "one way" things happen (always implicit behavior) but still allows the user to override the behavior.  In the future, we may be able to discover when the user explicitly wires it versus implicit behavior from a previous when context, but I don't think we should merge a version that doesn't behave as expected in that edge case.

rsetaluri (commented):
I think rather than modifying Enable class to store this information, we should just store this local to when. It seems that is the only consumer and producer of this information. Unless we expect someone else to check for this.

leonardt:
> I think rather than modifying Enable class to store this information, we should just store this local to when. It seems that is the only consumer and producer of this information. Unless we expect someone else to check for this.

Putting the logic in Enable is necessary for simple handling of "rewiring" logic.  That is, if we only update the flag inside when, then if a user decides to rewire the enable after the when, then the flag will still be set.  So it's simplest to have Enable's wire logic default to setting the flag to False, which covers the above case. Otherwise, Enable would need to be aware that the flag could be set externally, which seems like a worse choice w.r.t. encapsulation (this approach has Enable manage everything internally and provides when the interface to change the default behavior when necessary).

rsetaluri:
> > I think rather than modifying Enable class to store this information, we should just store this local to when. It seems that is the only consumer and producer of this information. Unless we expect someone else to check for this.
> 
> Putting the logic in Enable is necessary for simple handling of "rewiring" logic. That is, if we only update the flag inside when, then if a user decides to rewire the enable after the when, then the flag will still be set. So it's simplest to have Enable's wire logic default to setting the flag to False, which covers the above case. Otherwise, Enable would need to be aware that the flag could be set externally, which seems like a worse choice w.r.t. encapsulation (this approach has Enable manage everything internally and provides when the interface to change the default behavior when necessary).

Ok this makes sense.

PR #1172, opened 2022-10-26
Add better error message for trace infinite loop

leonardt:


tests/test_wire/test_wireable.py
```diff
@@ -20,3 +21,22 @@ class Main2(m.Circuit):
 Cannot wire Main2.a (Out(SInt[16])) to Main2.b (In(UInt[16]))\
 """
     assert caplog.messages[1][-len(expected):] == expected
+
+
+def test_trace_infinite_loop():
+    class Foo(m.Circuit):
```
rsetaluri, line 27: we can do this without a circuit right? if we do need the circuit wrapper for some reason let's name it `_`

magma/wire_container.py
```diff
@@ -201,7 +210,10 @@ def wired(self):
 
     # return the input or output Bit connected to this Bit
     def trace(self, skip_self=True):
-        return self._wire.trace(skip_self)
+        try:
```
rsetaluri, line 213: I would say this is an ok way to do it, but perhaps needs better logic to actually detect a combo loop. not sure the best way to that...

.github/workflows/deploy.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       run: |
         sudo apt install -y verilator libgmp-dev libmpfr-dev libmpc-dev
         python -m pip install --upgrade pip
+        # BEGIN: Temp fix, see
```
rsetaluri, line 24: needs a rebase?

rsetaluri (approved):
looks good, please address comments before merging

magma/wire_container.py
leonardt: Actually a more precise way might be to keep a list of the current trace and error if we encounter a seen value, will try that.

rsetaluri: To be fair, I don't think this is a bad way to do this. Letting the system take care of tracking those values for us...

leonardt: It technically requires more stack frames than necessary (e.g. we could catch the loop in three recursive calls versus waiting for the RecursionError), but now that I think about it, given this is the very uncommon case, it doens't seem worth it to add extra cost here just for that error.

magma/wire_container.py
```diff
@@ -12,6 +12,15 @@
 )
 
 
+class TraceRecursionError(RecursionError):
+    def __init__(self, value):
+        msg = (
+            f"RecursionError when calling trace on {value}, do you have a"
+            " combinational loop?"
```
rsetaluri, line 19: nit```suggestion
            f"RecursionError when calling trace on {value}, do you have a "
            "combinational loop?"
```

rsetaluri (approved):
last nit

PR #1173, opened 2022-10-26
[WireClockPass] Avoid Tuple recursion/elaboration

leonardt:
Checks Tuple subtypes before recursion to avoid unnecessary traversal


rsetaluri (approved):
nits, looks good

PR #1176, opened 2022-10-28
[When] Only add necessary implicit default values

leonardt:
Do latch detection, then pass latches to the implicit default value logic, only add the implicit default value if it is a latch (avoids unnecessary code in the generated verilog, e.g. when a register is assigned in every case we don't need the default update)

### DIFF
--- a/magma/primitives/when.py
+++ b/magma/primitives/when.py
@@ -1,7 +1,6 @@
-import abc
 import functools
 import itertools
-from typing import Dict, Iterable, Optional, Set, Tuple, Union
+from typing import Dict, Optional, Union, Set
 
 from magma.bits import Bits
 from magma.circuit import Circuit, CircuitType, CircuitBuilder
@@ -12,7 +11,6 @@ from magma.primitives.register import Register
 from magma.t import Type, In, Out
 from magma.when import (
     BlockBase as WhenBlock,
-    get_all_blocks as get_all_when_blocks,
     no_when,
     find_inferred_latches,
 )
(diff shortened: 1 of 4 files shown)


magma/primitives/when.py
```diff
@@ -91,7 +93,8 @@ def _get_corresponding_register_default(value: Type) -> Optional[Type]:
     return sel.select(value_inst.O)
 
 
-def _add_default_drivers_to_register_inputs(builder: 'WhenBuilder'):
+def _add_default_drivers_to_register_inputs(builder: 'WhenBuilder',
+                                            latches: Set[Type]):
```
rsetaluri, line 97:
```suggestion
def _add_default_drivers_to_register_inputs(
    builder: 'WhenBuilder', latches: Set[Type]
):
```

magma/primitives/when.py
```diff
@@ -203,16 +207,22 @@ def add_default_driver(self, drivee: Type, driver: Type):
         self._default_drivers[drivee] = driver
 
     def _finalize(self):
+        # Detect latches which would be inferred from the context of the when
+        # block.
+        latches = find_inferred_latches(self.block)
+
```
rsetaluri, line 213:
```suggestion
```

magma/primitives/when.py
```diff
@@ -203,16 +207,22 @@ def add_default_driver(self, drivee: Type, driver: Type):
...
         # and a global (post-processing) pass should be used instead.
-        _add_default_drivers_to_memory_ports(self)
-        _add_default_drivers_to_register_inputs(self)
-        # Detect latches which would be inferred from the context of the when
-        # block.
-        latches = find_inferred_latches(self.block)
+        # NOTE(leonardt): we only add default drivers if a value is inferred to
+        # be a latch, otherwise we skip this implicit logic so the output is
+        # cleaner (e.g. if a register is assigned in every case, we don't need
+        # the default value)
```
rsetaluri, line 222:
```suggestion
        # NOTE(leonardt): We only add default drivers if a value is inferred to
        # be a latch, otherwise we skip this implicit logic so the output is
        # cleaner (e.g. if a register is assigned in every case, we don't need
        # the default value).
```

rsetaluri (changes requested):
nits only

PR #1177, opened 2022-10-28
[Generator2] Improve class creation pipeline

leonardt:
Currently we have a guard for when the Generator2 class is created.  In fact, we need a more general check for when any class is created has the metaclass Generator2Meta.

One hacky way to determine that this is a type(...) call, i.e. it is invoked in the class creation pipeline.  In that case, there are exactly 3 arguments of the form name, bases, dct In theory, it would be possible for a Generator subclass to have the same argument number and types, but having the "__module__" member of the dct would be highly unlikely, furthmore they should be using a hashable dict, so it shouldn't satisfy that check.  Note that this is why we use "type is" instead of isinstance

### DIFF
--- a/magma/generator.py
+++ b/magma/generator.py
@@ -98,13 +98,23 @@ class _Generator2Meta(type):
         return type.__new__(metacls, name, bases, dct)
 
     def __call__(cls, *args, **kwargs):
-        if cls is Generator2:
-            try:
-                name, bases, dct = args
-            except ValueError:
-                raise Exception("Can not initialize base Generator class")
-            assert not kwargs
-            return type.__new__(cls, name, bases, dct)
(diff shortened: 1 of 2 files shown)


rsetaluri:
Is this all to make subclassing Generator sub-classes possible?

So to allow

```python
class Foo(m.Generator2):
    ...

class Bar(Foo):
    ...
```

leonardt:
> Is this all to make subclassing Generator sub-classes possible?
> 
> So to allow
> 
> ```python
> class Foo(m.Generator2):
>     ...
> 
> class Bar(Foo):
>     ...
> ```

Yes

leonardt:
(Although it's also a "better" implementation for the existing logic that tries to handled the Generator2 class definition)

rsetaluri (approved):

PR #1180, opened 2022-11-05
[When] Fix bug in default_driver logic

leonardt:
In the bulk resolve logic, there was a case where the root was not being used to get the default drivers, this manifested in a latch being inferred when there was nested whens and the bulk resolution was happening inside the nesting.

### DIFF
--- a/magma/wire_container.py
+++ b/magma/wire_container.py
@@ -281,7 +281,9 @@ class AggregateWireable(Wireable):
         default_drivers = []
         for ctx in self._wired_when_contexts:
             conditional_wires.append(ctx.get_conditional_wires_for_drivee(self))
-            default_drivers.append(ctx._default_drivers.pop(self, None))
+            default_drivers.append(
+                ctx.root.get_default_drivers_dict().pop(self, None)
+            )
         return self._wired_when_contexts, conditional_wires, default_drivers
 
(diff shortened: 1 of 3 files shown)


magma/wire_container.py
```diff
@@ -281,7 +281,9 @@ def _get_conditional_drivee_info(self):
         default_drivers = []
         for ctx in self._wired_when_contexts:
             conditional_wires.append(ctx.get_conditional_wires_for_drivee(self))
-            default_drivers.append(ctx._default_drivers.pop(self, None))
+            default_drivers.append(
```
rsetaluri, line 284: good to put this behind an API rather than accessing `_default_drivers` directly.

If you wanted you could also just have a specialized `pop_default_driver(drivee, default)` API instead of exposing the whole dict.

rsetaluri (approved):

tests/test_when.py
```diff
@@ -867,3 +867,33 @@ class _Test(m.Circuit):
 
     m.compile(f"build/{_Test.name}", _Test, output="mlir")
     assert check_gold(__file__, f"{_Test.name}.mlir")
+
+
+def test_when_latch_zext():
```
rsetaluri, line 872: also let's give this a name that describes what the test is actually testing (latch_nested_multiple ?)

PR #1183, opened 2022-11-09
[When] Allow user to override implicit logic

leonardt:
In the spirit of explicit is better than implicit, I think it's best if we don't require the user to use `rewire` if they want to override the implicit wire logic.  This changes the logic so that if we encounter a wire called on something implicitly wired, we remove the implicit wire (without raising an already driven error).  For this to work, we need to update the when builder code to mark that it is wiring implicitly when adding the drivee port logic.

### DIFF
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -96,8 +96,7 @@ class Enable(Bit):
         if get_debug_mode() and debug_info is None:
             debug_info = get_callee_frame_info()
         if self._driven_implicitly_by_when and not driven_implicitly_by_when:
-            # User is overriding the implicit when logic
-            self.unwire(debug_info)
+            self.unwire(debug_info)  # user is overriding the implicit when logic
         super().wire(o, debug_info)
         self._driven_implicitly_by_when = driven_implicitly_by_when
 
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -100,7 +100,7 @@ class Enable(Bit):
             # a when statement, but now the user is explicitly wiring the enable
             # When this happens, we choose the user's explicit wiring, and
             # discard the implicit logic (rather than forcing the user to call
-            # rewire explicitly)
+            # rewire explicitly).
             self.unwire(debug_info)
         super().wire(o, debug_info)
         self._driven_implicitly_by_when = driven_implicitly_by_when
--- a/magma/clock.py
+++ b/magma/clock.py
@@ -96,7 +96,12 @@ class Enable(Bit):
         if get_debug_mode() and debug_info is None:
             debug_info = get_callee_frame_info()
         if self._driven_implicitly_by_when and not driven_implicitly_by_when:
-            self.unwire(debug_info)  # user is overriding the implicit when logic
+            # In this case, the enable value was previously implicitly driven by
+            # a when statement, but now the user is explicitly wiring the enable
+            # When this happens, we choose the user's explicit wiring, and
+            # discard the implicit logic (rather than forcing the user to call
+            # rewire explicitly)
+            self.unwire(debug_info)
         super().wire(o, debug_info)
         self._driven_implicitly_by_when = driven_implicitly_by_when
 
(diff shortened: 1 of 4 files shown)


rsetaluri:
> In the spirit of explicit is better than implicit, I think it's best if we don't require the user to use `rewire` if they want to override the implicit wire logic. This changes the logic so that if we encounter a wire called on something implicitly wired, we remove the implicit wire (without raising an already driven error). For this to work, we need to update the when builder code to mark that it is wiring implicitly when adding the drivee port logic.

Not really sure I follow: "explicit is better than implicit" --> "I think it's best if we don't require the user to use `rewire`". Seems explicit = user calls rewire, and implicit = we do rewire/unwire behind the scenes. I would understand if we wanted to "auto" re/un-wire but that doesn't seem to follow from the "explicit is better than implicit" argument.

rsetaluri:
Also, can we try a rebase?

magma/clock.py
```diff
@@ -95,6 +95,9 @@ def driven_implicitly_by_when(self):
     def wire(self, o, debug_info=None, driven_implicitly_by_when=False):
         if get_debug_mode() and debug_info is None:
             debug_info = get_callee_frame_info()
+        if self._driven_implicitly_by_when and not driven_implicitly_by_when:
+            # User is overriding the implicit when logic
+            self.unwire(debug_info)
```
rsetaluri, line 100:
```suggestion
            self.unwire(debug_info)  # user is overriding the implicit when logic
```

(btw "overriding the implicit when logic" is vague and unclear what is actually happening, a more descriptive comment might be more useful).

magma/primitives/when.py
```diff
@@ -165,7 +165,14 @@ def _generic_add(
...
         with no_when():
-            wire(getattr(self, port_name), value)
+            port = getattr(self, port_name)
+            if value.is_input() and isinstance(value, Enable):
+                value.wire(
+                    port,
+                    driven_implicitly_by_when=value.driven_implicitly_by_when
+                )
+            else:
+                wire(port, value)
```
rsetaluri, line 175:
```suggestion
            port = getattr(self, port_name)
            wire_kwargs = {}
            if value.is_input() and isinstance(value, Enable):
                wire_kwargs["driven_implicitly_by_when"] = value.driven_implicitly_by_when
            value.wire(port, **wire_kwargs)
```

believe this is equiv.

rsetaluri (changes requested):

leonardt:
The explicit is better than implicit argument is that: if the user explicitly wires the CE, then that's what we should choose as the behavior, rather than defaulting to the implicit behavior (which requires the user to call rewire).  The pattern is that the implicit logic is only added if the user does not provide anything explicit, rather than the pattern being that the implicit logic is always added and the user has to deal with that.

magma/primitives/when.py
leonardt: This doesn't work for two reasons:
(1) We need to use `wire(...)` in the general case since this method is used for both inputs/outputs
(2) The driven_implicitly_by_when kwarg is only provided by the Enable wire method

We could: add the implicitly_by_when kwarg to all the wire locations, but this seems like a big change for little benefit.

magma/clock.py
```diff
@@ -95,6 +95,13 @@ def driven_implicitly_by_when(self):
     def wire(self, o, debug_info=None, driven_implicitly_by_when=False):
         if get_debug_mode() and debug_info is None:
             debug_info = get_callee_frame_info()
+        if self._driven_implicitly_by_when and not driven_implicitly_by_when:
+            # In this case, the enable value was previously implicitly driven by
+            # a when statement, but now the user is explicitly wiring the enable
+            # When this happens, we choose the user's explicit wiring, and
+            # discard the implicit logic (rather than forcing the user to call
+            # rewire explicitly)
```
rsetaluri, line 103: nit
```suggestion
            # In this case, the enable value was previously implicitly driven by
            # a when statement, but now the user is explicitly wiring the enable
            # When this happens, we choose the user's explicit wiring, and
            # discard the implicit logic (rather than forcing the user to call
            # rewire explicitly).
```

rsetaluri (approved):
lgtm, nit

PR #1187, opened 2022-11-16
[When] Fix bug in resolve logic for when output

leonardt:
If the user call's .value() on a conditionally driven value, they may leave a hanging reference to a when output that was resolved.  When a value is resolved, it's children are added as output ports to the builder and original output "discarded".  To handle these stale references, at finalization we can traverse the outputs and update anything still driven by a stale output with the resolved children.

### DIFF
--- a/magma/primitives/when.py
+++ b/magma/primitives/when.py
@@ -207,8 +207,7 @@ class WhenBuilder(CircuitBuilder):
         self._default_drivers[drivee] = driver
 
     def _update_resolved_refs(self):
-        """
-        When a value driven by a when is resolved, it's children are added as
+        """When a value driven by a when is resolved, its children are added as
         outputs.  The user may have referenced the unresolved when output using
         .value(), so at this point, we update all the references by checking if
         any resolved outputs are being used to drive values.
@@ -218,7 +217,7 @@ class WhenBuilder(CircuitBuilder):
                 continue
             if value.name._parent not in self._output_to_name:
                 continue
-            # Found resolved port, rewire to child output
+            # Found resolved port, rewire to child output.
             port = getattr(self, self._output_to_name[value.name._parent])
             for drivee in port[value.name.index].driving():
                 drivee.rewire(getattr(self, name))
--- a/magma/primitives/when.py
+++ b/magma/primitives/when.py
@@ -8,6 +8,7 @@ from magma.circuit import Circuit, CircuitType, CircuitBuilder
 from magma.conversions import from_bits
 from magma.digital import Digital
 from magma.primitives.register import AbstractRegister
+from magma.ref import DerivedRef
 from magma.t import Type, In, Out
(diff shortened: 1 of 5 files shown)


magma/ref.py
```diff
@@ -196,8 +196,9 @@ def set_defn(self, defn):
 
 
 class DerivedRef(Ref):
-    def __init__(self, parent):
+    def __init__(self, parent, index):
         self._parent = parent
+        self.index = index
```
rsetaluri, line 201: let's make this private and use an accessor function for uniformity: `def index(self): return self._index`

(property is probably better but just for uniformity's sake)

rsetaluri: wait is this needed for this change to work?

magma/primitives/when.py
```diff
@@ -205,7 +206,25 @@ def add_default_driver(self, drivee: Type, driver: Type):
...
+        """
+        for value, name in self._output_to_name.items():
+            if not isinstance(value.name, DerivedRef):
+                continue
+            if value.name._parent not in self._output_to_name:
+                continue
+            # Found resolved port, rewire to child output
+            port = getattr(self, self._output_to_name[value.name._parent])
+            for drivee in port[value.name.index].driving():
+                drivee.rewire(getattr(self, name))
```
rsetaluri, line 224: nit
```suggestion
        """When a value driven by a when is resolved, its children are added as
        outputs.  The user may have referenced the unresolved when output using
        .value(), so at this point, we update all the references by checking if
        any resolved outputs are being used to drive values.
        """
        for value, name in self._output_to_name.items():
            if not isinstance(value.name, DerivedRef):
                continue
            if value.name._parent not in self._output_to_name:
                continue
            # Found resolved port, rewire to child output.
            port = getattr(self, self._output_to_name[value.name._parent])
            for drivee in port[value.name.index].driving():
                drivee.rewire(getattr(self, name))
```

magma/ref.py
leonardt: Yes, it's needed to generically refer to the index of a reference without knowing whether it's an array or tuple (both derivedref's have an index

magma/ref.py
```diff
@@ -196,8 +196,9 @@ def set_defn(self, defn):
 
 
 class DerivedRef(Ref):
-    def __init__(self, parent):
+    def __init__(self, parent, index):
         self._parent = parent
+        self._index = index
```
rsetaluri, line 201: yes but all sub-classes of `DerivedRef` already have an `index` member right? (just being pedantic here, not a big deal).

rsetaluri (approved):

magma/ref.py
leonardt: The code checks DerivedRef, so it's better to enforce the interface in the base class init then to expect that all subclasses happen to add a specific member

PR #1188, opened 2022-11-16
[When] Remove default driver in unwire logic

leonardt:


magma/when.py
```diff
@@ -68,6 +68,9 @@ def remove_conditional_wire(self, i):
         self._conditional_wires = list(
             filter(lambda x: x.drivee is not i, self._conditional_wires)
         )
+        if i in self.root._default_drivers:
```
rsetaluri, line 71:
```
try:
    self.root._default_drivers.pop(i)
except KeyError:
     pass
else:
    self.root.builder.remove_default_driver(i)
```
might be more pythonic

rsetaluri (approved):

magma/when.py
leonardt: used the pop default pattern for a two liner.

PR #1189, opened 2022-11-17
[Debug] Use uinspect for source file/line info

leonardt:


Kuree:
One issue of using C extension is that it has to be build with every Python platform. I'm wondering if it is possible to achieve the same result with `ctypes`?

Also based on my benchmarks, `f_locals` in Python is fast enough to use it without using C extensions.

leonardt:
So I tried hacking together a simple example:
```

import ctypes as ct
from distutils.sysconfig import get_config_var
libpython_name = "%s/%s" % (get_config_var("LIBDIR"), get_config_var("INSTSONAME"))

libpython = ct.cdll.LoadLibrary(libpython_name)
libpython.PyThreadState_Get()
```
but this crashes

I realized you need the GIL before checking thread state, but that causes the program to hang
```
import ctypes as ct
from distutils.sysconfig import get_config_var
libpython_name = "%s/%s" % (get_config_var("LIBDIR"), get_config_var("INSTSONAME"))

libpython = ct.cdll.LoadLibrary(libpython_name)
libpython.PyGILState_Ensure()
libpython.PyThreadState_Get()
```

So I think it might be problematic to try to interact with the thread state API from within python.  I guess once you take the lock through the API, control never returns to the interpreter?

Kuree:
I think you might be able to workaround this issue using `ctypes.pythonapi`. Let me experiment with it before getting back to you.

Kuree:
Here is what I have using ctypes with Python 3.10:

```python
import ctypes
import time


get_line = ctypes.pythonapi.PyFrame_GetLineNumber
get_line.restype = ctypes.c_int
get_line.argtypes = [ctypes.c_void_p]
get_frame = ctypes.pythonapi.PyEval_GetFrame
get_frame.restype = ctypes.c_void_p;

get_frame_back = ctypes.pythonapi.PyFrame_GetBack
get_frame_back.restype = ctypes.c_void_p;
get_frame_back.argtypes = [ctypes.c_void_p]

def main():
    frame = get_frame()
    frame = get_frame_back(frame)
    line = get_line(frame)
    print("line", line)


main()
```

It seems to work, but it's actually slower than inspect module... Let me poke around and see what can be done to make it faster.

leonardt:
It seems that github actions has a manylinux wheel builder we could try: https://github.com/marketplace/actions/manylinux-wheel-builder

leonardt:
One simple option that works for modern systems is to distribute an `sdist` that includes the debug-ext folder (and recursively the pybind source) and builds it on the user system when installed.  It simply requires that they have cmake installed, which they probably do if they've installed MLIR/CIRCT.  This isn't quite as convenient as the manylinux wheel but this will probably work for a short-term solution

Kuree:
It would be probably better to maintain a separate python package to deal with the debugging information. I've done some experiments/benchmarks and should be able to share tonight or tomorrow.

Kuree:
See here: https://github.com/Kuree/uinspect

I will update the benchmark info and upload it to PYPI once I have CI working properly.

Kuree:
@leonardt Can you try to install `uinspect`? It's been released for every Python version and every OS.

leonardt:
I'm not able to pip install uinspect on my mac, I wonder if it has to do with the fix wheel logic? I remember for pycoreir we had to add this script: https://github.com/leonardt/pycoreir/blob/master/.travis/fix_wheel.py

I'm on macOS 12.6.1, python 3.8.12, but not sure if that really helps.  

Looking at https://pypi.org/project/uinspect/0.0.1/#files, the mac distribution seems to be for 12.0, which I'm not sure pattern matches upwards? Looking at https://pypi.org/project/coreir/#files, I see one difference as "py3_none" vs "cp3_cp3", not sure if that's relevant.

Kuree:
That `fix_wheel.py` script is used to remove python ABI flags since pycoreir doesn't use any Python C API. In this case we do since we're calling Python APIs, which are version dependent.

Let me spin up a mac VM and see if I can tweak the wheels by artificially lowering it to some ancient macos version.

Kuree:
Also given the benchmark info, getting location is relatively cheap, i.e. 0.2us/op, I'm wondering if it's possible to keep the source location tracking on by default? This is the expected behavior for most compilers.

leonardt:
Yes, I think we should have it on my default, we could give an option to turn it off, but the performance overhead seems to be nominal in my testing so I don't think it's that important.  It's quite useful for error messages in the Python level too (for exceptions, we can print tracebacks to the orig source file/line e.g. for wiring errors)

Kuree:
I'm not able to reproduce the error on a 12.6.1 machine:
image

Let me try to lower the MacOS version using `MACOSX_DEPLOYMENT_TARGET`.

Kuree:
I'm not able to reproduce the error on a 12.6.1 machine:
image

Let me try to lower the MacOS version using `MACOSX_DEPLOYMENT_TARGET`.

leonardt:
Weird, I'm on Python 3.8, but not sure if that matters.  Also my python is installed with miniconda, maybe that changes the compatability?

Screen Shot 2022-11-29 at 2 46 42 PM

Kuree:
@leonardt I lowered it to `10.15`. Please let me know if it works for you or not.

leonardt:
Yup works now

Kuree:
Cool. Please let me know if you want any other inspect related functions. For instance, walk the stack with any given predicate functions and etc. I can try to implement them in uinspect and then benchmark the performance against `inspect` module.

magma/circuit.py
```diff
@@ -264,9 +264,8 @@ def __new__(metacls, name, bases, dct):
         return cls
 
     def __call__(cls, *largs, **kwargs):
-        if get_debug_mode():
-            debug_info = get_callee_frame_info()
-            kwargs["debug_info"] = debug_info
+        debug_info = get_debug_info(3)
+        kwargs["debug_info"] = debug_info
```
rsetaluri, line 268: nit
```suggestion
        kwargs["debug_info"] = get_debug_info(3)
```

(unless this screws up the callstack lol)

magma/debug.py
```diff
@@ -1,43 +1,31 @@
-import inspect
+import uinspect
 import collections
-import magma
-from magma.config import get_debug_mode
 
 
-debug_info = collections.namedtuple("debug_info", ["filename", "lineno", "module"])
+debug_info = collections.namedtuple("debug_info",
+                                    ["filename", "lineno", "module"])
```
rsetaluri, line 6: while we're here (`import dataclasses` above)
```suggestion
class _DebugInfo:
    filename: str
    lineno: int
    module: str
    
 
 debug_info = _DebugInfo
```

magma/debug.py
```diff
@@ -1,43 +1,31 @@
...
-    for i in range(0, 10):
-        module = inspect.getmodule(callee_frame)
-        # Go up until we're out of the magma module (assuming this is the user
-        # code)
-        if not module or module.__name__.split(".")[0] != "magma":
-            break
-        callee_frame = callee_frame.f_back
-    callee_frame = inspect.getframeinfo(callee_frame)
-    return debug_info(callee_frame.filename, callee_frame.lineno, module)
+def get_debug_info(frames_to_skip):
```
rsetaluri, line 9: do we always know `frames_to_skip` statically? vs. the popping up dynamically...I kind of like the dynamic one, unless this is pretty robust.

magma/testing/utils.py
```diff
@@ -55,15 +56,21 @@ def magma_debug_section():
     return _MagmaDebugSection()
 
 
+def _str_log(msg):
+    # Remove escaped formatting, e.g. bold in log
+    ansi_escape = re.compile(r'(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]')
+    return ansi_escape.sub('', str(msg))
```
rsetaluri, line 62: nit
```suggestion
def _unformat(msg):
    # Remove escaped formatting, e.g. bold or color in stdout log.
    ansi_escape = re.compile(r'(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]')
    return ansi_escape.sub('', str(msg))
```

rsetaluri (approved):

rsetaluri (changes requested):
changes requested from above

magma/debug.py
leonardt: The dynamic one requires traversing the stack with the inspect module which I think is one of the reasons why the current debug mode is so slow.

One option would be to have the C api do the traversal out of the magma module for us, but that doesn't quite work well for one reason: the primitives are inside the magma module and we'd want debug info to be generated there.  I think in general it's faster and easier to specify frames_to_skip the in places where we know we want to skip frames

Kuree: > the primitives are inside the magma module and we'd want debug info to be generated there.

If you can specify a list of files to allow and discard, I can introduce a C API that traverse the stack quickly. I believe it should be orders of magnitude faster than `inspect` module.

leonardt: We could try that by passing the list of core magma .py files (so basically traverse outside of the core codebase to try to figure out where the wire occured)

Kuree: @leonardt Can you try the latest `uinspect`?  I've added a simple frame walking logic that look for any frames that don't originate from the block list (see [here](https://github.com/Kuree/uinspect/blob/c18449a03dd200da35ecd4f5abee3f5a1446cea3/tests/test_uinspect.py#L51)).

Based on the benchmark, it's about 35x faster than walking the stack in Python, about 1.5us/op. So it's still 5x more expensive than static methods.

leonardt: Could we try adding an API to register the files to skip, then invoke the stack walking without having to pass the file names each time? (Perhaps avoiding passing the strings each time would speed things up a bit)

Kuree: You only need to create the `FrameWalker` object once, by providing a set of strings. Then any subsequent calls of `get_location()` will walk the stack. Here is some usage:

```python
walker = uinspect.FrameWalker({"foo.py"})


def foo():
   loc = walker.get_location()


def bar():
  loc = walker.get_location()

```


I think the slowdown is caused by casting Python string into std::string, and then search in an unordered_set. If a more efficient string hash data structure is used, I think we can speed it up more. Let me do more experiments.

magma/backend/verilog.py
```diff
@@ -123,7 +123,8 @@ def arg(k,v):
     debug_str = ""
     for k, v in self.interface.ports.items():
         if getattr(v, "debug_info", False) and get_codegen_debug_info():
-            filename, lineno, module = v.debug_info
+            filename, lineno, module = \
```
rsetaluri, line 126: nit: prefer paren's here over line break

magma/backend/verilog.py
```diff
@@ -176,7 +177,10 @@ def compiledefinition(cls):
 
     args = ', '.join(vmoduleargs(cls.interface))
     s = ''
-    if get_codegen_debug_info() and cls.debug_info.filename and cls.debug_info.lineno:
+    if (
+        get_codegen_debug_info() and cls.debug_info and
+        cls.debug_info.filename and cls.debug_info.lineno
```
rsetaluri, line 182: nit
```suggestion
        get_codegen_debug_info()
        and cls.debug_info
        and cls.debug_info.filename
        and cls.debug_info.lineno
```

magma/config.py
```diff
@@ -89,7 +89,7 @@ def __setitem__(self, key, value):
 
 config = ConfigManager(
     compile_dir=RuntimeConfig("normal"),
-    debug_mode=RuntimeConfig(False),
+    debug_mode=RuntimeConfig(True),
```
rsetaluri, line 92: I think switching this default in a separate PR would be preferred

magma/debug.py
rsetaluri: do we have a conclusion here?

leonardt: I think we should just use static frames to skip for now and add this support as an improvement

rsetaluri: let's make an issue for this to track then.

magma/debug.py
```diff
@@ -1,43 +1,43 @@
...
-            break
-        callee_frame = callee_frame.f_back
-    callee_frame = inspect.getframeinfo(callee_frame)
-    return debug_info(callee_frame.filename, callee_frame.lineno, module)
+
+debug_info = _DebugInfo
+
+
+def get_debug_info(frames_to_skip):
+    if config.use_uinspect:
```
rsetaluri, line 19: let's actually register this config in this file locally, rather than in the global config.

rsetaluri (approved):
Final change request, lg otherwise!

PR #1194, opened 2022-11-19
Infer names in metaclass pipeline

leonardt:
Since this is quite useful for error messages and generated code readability and it seems to incur nominal overhead, I figured it'd be best to turn it on by default.  We could guard it behind the debug mode switch if this is problematic for existing flows (perhaps the name change will break scripts that assume the default naming schemes).  Alternatively, and perhaps better, would be to add a flag to disable this which we could have anybody who depends on the old naming scheme use.


leonardt:
Closing in favor of https://github.com/phanrahan/magma/pull/1201

PR #1199, opened 2022-11-22
[InlineVerilog] Avoid inserting named wires

rsetaluri:


leonardt (approved):
Looks good, can we open an issue to track the test migration for the skipped tests?

PR #1201, opened 2022-11-28
Namer dict

leonardt:
Adds name inference to Circuit and Generator (requires setattr on `self` to work).

Since this is quite useful for error messages and generated code readability and it seems to incur nominal overhead, I figured it'd be best to turn it on by default. We could guard it behind the debug mode switch if this is problematic for existing flows (perhaps the name change will break scripts that assume the default naming schemes). Alternatively, and perhaps better, would be to add a flag to disable this which we could have anybody who depends on the old naming scheme use.


rsetaluri (commented):
Is it possible to guard this with a flag? I think it would be nice to have this be an optional feature, like inferred/auto/smart names or something

leonardt:
Guarded with debug mode, but I do think this should be on by default because it makes things like error messages much easier to understand

leonardt:
Turns out guarding behind the global debug mode flag and enabling that for the tests doesn't work since it breaks some of the coreir tests, for now we can introduce a separate flag

rsetaluri:
I was more concerned about reviewing this change with all the changes to tests and source in one PR. Would it be possible just to separate out the new implementation and perhaps a small test suite that tests that and then turn it on for default in a follow up?

magma/circuit.py
```diff
@@ -392,12 +469,18 @@ def set_debug_info(self, debug_info):
...
+            name = f"AnonymousCircuitInst{id(self)}"
+        if type(self) is AnonymousCircuitType:
+            interface = ", ".join(
+                f"{name}: {type(value)}"
+                for name, value in self.interface.ports.items()
+            )
+            name += f"<{interface}>"
+        else:
+            name += f"<{type(self)}>"
+        return name
```
leonardt, line 483: This was an artifact for the updated tests since it improves the "stringification" logic for named anon circuit instances.

magma/interface.py
```diff
@@ -114,22 +114,18 @@ def _make_wires(value, wired):
...
-        if (driver, value) in wired:
-            break
-        s += _make_wire_str(driver, value, wired)
-        if not driver.driven():
-            break
-        value = driver
-        driver = driver.value()
+    s = _make_wire_str(driver, value, wired)
+    if driver.driven():
+        s += _make_wires(driver, wired)
```
leonardt, line 128: Similarly this was updated for the tests, but is cosmetic (for repr)

conftest.py
```diff
@@ -1,11 +1,13 @@
...
-import magma.config
 from magma.config import config as magma_config
 from magma.util import reset_global_context
 
 
 def pytest_configure(config):
     magma_config.compile_dir = 'callee_file_dir'
+    # TODO: Enable this globally for testing
+    # revert a3a2452168f2251277a14548d053a9ca7a45bff7 for golds
+    # magma_config.use_namer_dict = True
```
rsetaluri, line 10:
```suggestion
    # TODO(leonardt): Enable this globally for testing.
    # Revert a3a2452168f2251277a14548d053a9ca7a45bff7 for golds.
    #
    # magma_config.use_namer_dict = True
```

magma/circuit.py
```diff
@@ -204,14 +205,78 @@ def _get_intermediate_values(value):
     return values
 
 
+class NamerDict(dict):
```
rsetaluri, line 208: should inherit `collections.UserDict`?

magma/circuit.py
```diff
@@ -204,14 +205,78 @@ def _get_intermediate_values(value):
...
+        value: Union[Type, 'Circuit'],
+    ):
+        """If key has been seen more than once, "uniquify" the names by append
+           _{i} to them."""
+        values = self._inferred_names.setdefault(key, [])
+        if len(values) == 0:
+            self._set_name(key, value)
+        elif any(value is x for x in values):
+            # Already uniquified
+            return
```
rsetaluri, line 230: nit
```suggestion
            return  # already uniquified
```

magma/circuit.py
```diff
@@ -204,14 +205,78 @@ def _get_intermediate_values(value):
...
+        else:
+            if len(values) == 1:
+                # Make the first value consistent by append _0
```
rsetaluri, line 233: nit
```suggestion
                # Make the first value consistent by appending _0.
```

magma/circuit.py
```diff
@@ -204,14 +205,78 @@ def _get_intermediate_values(value):
...
+                self._set_name(f"{key}_0", values[0])
+            self._set_name(f"{key}_{len(values)}", value)
+        values.append(value)
+
+    def _set_value_name(self, key: str, value: Type):
+        if not hasattr(value, "name"):
+            return  # Interface object is a Type without a name.
```
rsetaluri, line 240: nit
```suggestion
            return  # interface object is a Type without a name
```

magma/circuit.py
```diff
@@ -204,14 +205,78 @@ def _get_intermediate_values(value):
...
+        else:
+            self._check_unique_name(value.name, value)
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+        if isinstance(value, (Type, MagmaProtocol)):
+            self._set_value_name(key, value)
+        elif isinstance(type(value), CircuitKind):
+            # NOTE: we check type(value) because this code is run in the Circuit
+            # class creation pipeline (so Circuit may not be defined yet).
```
rsetaluri, line 258:
```suggestion
            # NOTE(leonardt): we check type(value) because this code is run in the Circuit
            # class creation pipeline (so Circuit may not be defined yet).
```

magma/circuit.py
```diff
@@ -204,14 +205,78 @@ def _get_intermediate_values(value):
...
+            self._set_inst_name(key, value)
+
+    def __hash__(self):
+        return hash(tuple(sorted(self.items())))
+
+
+config.register(use_namer_dict=RuntimeConfig(False))
```
rsetaluri, line 265: nit: let's move this to top of file to be consistent with other files registering configs

rsetaluri (approved):

PR #1205, opened 2022-12-02
Change seq2 get/set attr

cdonovick:
use the decorated cls's methods instead of object
    enables PEak's peak poke protocol

### DIFF
--- a/magma/syntax/sequential2.py
+++ b/magma/syntax/sequential2.py
@@ -418,7 +418,8 @@ def sequential2(pre_passes=[], post_passes=[],
 
         def sequential_getattribute(self, key):
             """
-            Wrap lists so we can use the setattr interface with lists of registers
+            Wrap lists so we can use the setattr interface with lists of
+            registers
             """
             result = cls_getattribute(self, key)
             if isinstance(result, list):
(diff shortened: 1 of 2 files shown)


magma/syntax/sequential2.py
```diff
@@ -448,6 +413,42 @@ def seq_inner(cls):
...
         io_args = build_io_args(_annotations, output_port_names)
 
+        cls_setattr = cls.__setattr__
+        cls_getattribute = cls.__getattribute__
+
+        def sequential_getattribute(self, key):
+            """
+            Wrap lists so we can use the setattr interface with lists of registers
+            """
+            result = cls_getattribute(self, key)
```
leonardt, line 423: Would calling super() work here?

cdonovick: No.  There is no inheritance going on.  If we did something like the follow could use super but its more of a change:
```Python
class new_cls(cls): 
     def __setattr__(self, attr):
            ... # body of seq_setattr
      
      def __getattribute__(self, attr, value): 
           ... # similar
```

leonardt: Ah yes, I see now, it's being "overridden" via monkey patch rather than subclassing.

leonardt (approved):

leonardt (approved):

Issue #1211, opened 2022-12-20
Does len() on magma enumerated type return bit length of member?


rkshthrmsh:
It appears that len() on magma enumerated type returns bit length of members. This differs from the native Python Enum, which returns the number of members. Bit length of member is useful, however, is there also a way to access the number of members?

```python
# Magma Enum
class State(m.Enum):
        IDLE=0
        CONFIG=1
        OPERATE=8

# len(State)
# 4
```
```python
# Python Enum
class State(Enum):
        IDLE=0
        CONFIG=1
        OPERATE=8

# len(State)
# 3
```

leonardt:
Added support for this via https://github.com/phanrahan/magma/pull/1212

rkshthrmsh:
Thanks! With this change, would State.size or State.N be the recommended way of accessing the bit length of the members?

leonardt:
It will expose access to an attribute "fields" which will contain a dictionary mapping to name to value, here's a simple example for what you're looking for:

```
    class MyEnum(m.Enum):
        x = 0
        y = 1

    assert len(MyEnum.fields) == 2
```

rkshthrmsh:
Alright. So, `len(MyEnum)` will return the bit length of any member, and `len(MyEnum.fields)` will return the number of members.

PR #1212, opened 2022-12-20
[Enum] Add API to access Enum fields

leonardt:
Addresses https://github.com/phanrahan/magma/issues/1211 by providing a way for the user to have access to the enum field dictionary.

### DIFF
--- a/tests/test_type/test_enum.py
+++ b/tests/test_type/test_enum.py
@@ -43,3 +43,11 @@ def test_reserved():
         class MyEnum(m.Enum):
             a = 0
             N = 2
+
+
+def test_enum_fields():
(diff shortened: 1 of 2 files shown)


tests/test_type/test_enum.py
```diff
@@ -43,3 +43,11 @@ def test_reserved():
...
             a = 0
             N = 2
+
+
+def test_enum_fields():
+    class MyEnum(m.Enum):
+        x = 0
+        y = 1
+
+    assert len(MyEnum.fields) == 2
```
rsetaluri, line 53: would just check that `fields` is the exact data structure we expect, instead of just checking the length

rsetaluri (approved):

PR #1216, opened 2022-12-22
[MLIR] Source location additions

rsetaluri:


magma/backend/mlir/mlir.py
```diff
@@ -72,14 +72,34 @@ class MlirLocation(MlirAttribute):
...
     # dialect for the default (unknown) location, we need to use a dynamic
     # import. Fortunately, this is used in a dataclass field factory function
     # anyway, and we can cache it so it will be little to no overhead.
     from magma.backend.mlir.builtin import builtin
-    return builtin.UnknownLoc()
+    location_stack = get_location_stack()
+    try:
+        return location_stack.peek()
+    except IndexError:
+        return builtin.UnknownLoc()
```
leonardt, line 102:
```suggestion
    location_stack = get_location_stack()
    return location_stack.peek_default(builtin.UnknownLoc())
```

magma/backend/mlir/sv.py
```diff
@@ -286,13 +296,17 @@ def print(self, printer: PrinterBase, opts: PrintOpts):
...
+                printer.print_line(self.location.emit())
             return
         printer.print(" else {")
         printer.flush()
         printer.push()
         self._else_block.print(printer, opts)
         printer.pop()
         printer.print_line("}")
+        if opts.print_locations:
+            printer.print_line(self.location.emit())
```
leonardt, line 309: Seems like we always check `opts` and print after the function returns? Perhaps we can just wrap these in a decorator that does that at the end of the function (rather than adding the code at the various return points).

leonardt (approved):
Minor changes requested.

PR #1218, opened 2022-12-28
[SmartBits] Add concat signedness determinator visitor

rsetaluri:


leonardt (approved):
LGTM, can we add a test covering this?

PR #1220, opened 2023-01-04
[MLIR] Add option to emit muxes as if-then-else

rsetaluri:


magma/backend/mlir/hardware_module.py
```diff
@@ -160,16 +160,49 @@ def get_module_interface(
...
+            return
+        cond = ctx.new_value(builtin.IntegerType(1))
+        const = ctx.new_value(select.type)
+        hw.ConstantOp(value=index, results=[const])
+        comb.ICmpOp(predicate="eq", operands=[select, const], results=[cond])
+        curr = sv.IfOp(operands=[cond])
+        with push_block(curr.then_block):
+            sv.BPAssignOp(operands=[wire, data[index]])
+        with push_block(curr.else_block):
+            _process()
```
leonardt, line 186: Will the lack of `else` logic here be problematic for downstream tools? I could imagine a potential latch inference being triggered if the tool doesn't reason about all possible values being matched.  In fact, if there is a non power of two number of inputs then there is a potential for a latch (e.g. if the select logic is generated incorrectly w.r.t. to the number of mux inputs, it's possible for the select value to be equal to a constant that's greater than len(data)).  IIRC, MLIR would normally emit the mux as an array create/select.  Perhaps in Verilog this is okay because the "extra" select values are ignored?

rsetaluri: Finalizing need with customer. Will report back what codegen is desired.

PR #1224, opened 2023-01-09
[When] Enforce when condition to be Bit type

leonardt:


magma/when.py
```diff
@@ -219,7 +226,9 @@ def get_all_blocks(
 
 @dataclasses.dataclass
 class _WhenBlockInfo:
-    condition: Any  # TODO(rsetaluri): Should actually be bit
+    # NOTE(leonardt): we can't use Bit here because of circular dependency, so
+    # instead we enforce it in the public constructors (e.g. m.when).
```
rsetaluri, line 230: nit
```suggestion
    # NOTE(leonardt): We can't use Bit here because of circular dependency, so
    # instead we enforce it in the public constructors (e.g. m.when).
```

rsetaluri (approved):

PR #1226, opened 2023-01-17
[Debug] Add magma helper function annotation

rsetaluri:
This introduces the `@magma_helper_function` annotation (decorator) which can be used to mark functions which instantiate magma circuits. Such functions will then defer to the caller to retrieve debug information (for source location annotations in generated code).

### DIFF
--- a/magma/debug.py
+++ b/magma/debug.py
@@ -1,11 +1,13 @@
-from dataclasses import dataclass
-import uinspect
 import collections
+import dataclasses
+import functools
+import uinspect
 
+from magma.common import Stack
 from magma.config import config, RuntimeConfig
 
 
-@dataclass
+@dataclasses.dataclass
 class _DebugInfo:
     filename: str
     lineno: int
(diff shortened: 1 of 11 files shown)


magma/debug.py
```diff
@@ -44,3 +59,28 @@ def unwire(i, o=None, debug_info=None, keep_wired_when_contexts=False):
...
+        # additive.
+        try:
+            offset = extra_frames_to_skip_stack.peek()
+        except IndexError:
+            offset = 0
+        # NOTE(rsetaluri): We need to skip an extra frame (1 + 1 = 2) since we
+        # are returning a wrapped function (decorated).
+        extra_frames_to_skip_stack.push(2 + offset)
+        # TODO(rsetaluri): Catch potential exceptions.
+        ret = fn(*args, **kwargs)
```
rsetaluri, line 79: @leonardt what's the right way to catch exceptions here?

```python
try:
    ret = fn(*args, **kwargs)
except Exception as e:
    extra_frames_to_skip_stack.pop()
    raise e from None
```
?

Seems like there's some better pattern to do this?

magma/debug.py
```diff
@@ -18,7 +20,20 @@ class _DebugInfo:
...
+def _get_extra_frames_to_skip_stack():
+    global _extra_frames_to_skip_stack
+    return _extra_frames_to_skip_stack
+
+
 def get_debug_info(frames_to_skip):
+    try:
+        extra_frames_to_skip = _get_extra_frames_to_skip_stack().peek()
+    except IndexError:
+        extra_frames_to_skip = 0
```
leonardt, line 35: Could be useful to add a `safe_peek` method to the `Stack` which takes a default argument
```suggestion
    extra_frames_to_skip = _get_extra_frames_to_skip_stack().safe_peek(default=0)
```

leonardt (approved):

PR #1227, opened 2023-01-19
[MLIR] Add option to disallow local variables

rsetaluri:


magma/backend/mlir/compile_to_mlir_opts.py
```diff
@@ -16,3 +16,4 @@ class CompileToMlirOpts:
     location_info_style: str = "none"
     explicit_bitcast: bool = False
     disallow_expression_inlining_in_ports: bool = False
+    disallow_local_variables: bool = False
```
leonardt, line 19: Should we consider making this True by default until the issue is resolved?

leonardt (approved):

magma/backend/mlir/compile_to_mlir_opts.py
rsetaluri: I think no: (a) would possibly require updating the general set of gold verilog; (b) we still need to test on customer code, so it's easier to do it in 2 steps (adding flag & setting flag).

PR #1229, opened 2023-01-25
[MLIR] Use worklist style for graph builder

leonardt:
Avoids recursion error for sufficiently large graphs


leonardt:
It looks like https://github.com/phanrahan/magma/pull/1231 this fixes the root cause of the issue and avoids the graph size problem here, but we may want to still do this at some point.

PR #1232, opened 2023-01-31
[Types] Fix bug in digital.is_wireable logic

leonardt:
The existing logic did not handle the case when the undirected types were subclases of each other.  This fixes it by simplifying the logic to just check whether the undirected types are compatible.

### DIFF
--- a/magma/digital.py
+++ b/magma/digital.py
@@ -136,9 +136,8 @@ class DigitalMeta(ABCMeta, Kind):
             return True
         rhs = magma_type(rhs)
         # allows undirected types to match (e.g. for temporary values)
-        return (issubclass(rhs.flip(), cls) or issubclass(cls, rhs.flip()) or
-                cls.qualify(Direction.Undirected) is rhs or
-                rhs.qualify(Direction.Undirected) is cls)
+        return (issubclass(rhs.undirected_t, cls.undirected_t) or
+                issubclass(cls.undirected_t, rhs.undirected_t))
 
     def is_bindable(cls, rhs):
         return issubclass(cls, magma_type(rhs))
--- a/magma/digital.py
+++ b/magma/digital.py
@@ -136,8 +136,10 @@ class DigitalMeta(ABCMeta, Kind):
             return True
         rhs = magma_type(rhs)
         # allows undirected types to match (e.g. for temporary values)
-        return (issubclass(rhs.undirected_t, cls.undirected_t) or
-                issubclass(cls.undirected_t, rhs.undirected_t))
+        return (
+            issubclass(rhs.undirected_t, cls.undirected_t)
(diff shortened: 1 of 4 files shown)


leonardt:
I realized that we need to keep the direction specific logic because we don't want in/in matching, so this changes the `is` check to a subclass check, which covers the Digital/Bit case without changing the explicitly directed case.

leonardt:
Actually, it turns out the original change was okay, and the test output needs to be updated.  That is because is_wireable only needs to check the compatibility of the types, and the qualification is checked in another part of the pipeline (leading to the different expected output in the tests for the qualifier specific error message)

magma/digital.py
```diff
@@ -136,9 +136,8 @@ def is_wireable(cls, rhs):
             return True
         rhs = magma_type(rhs)
         # allows undirected types to match (e.g. for temporary values)
-        return (issubclass(rhs.flip(), cls) or issubclass(cls, rhs.flip()) or
-                cls.qualify(Direction.Undirected) is rhs or
-                rhs.qualify(Direction.Undirected) is cls)
+        return (issubclass(rhs.undirected_t, cls.undirected_t) or
+                issubclass(cls.undirected_t, rhs.undirected_t))
```
rsetaluri, line 140: nit
```suggestion
        return (
            issubclass(rhs.undirected_t, cls.undirected_t)
            or issubclass(cls.undirected_t, rhs.undirected_t)
        )
```

tests/test_type/test_product.py
```diff
@@ -0,0 +1,10 @@
+import magma as m
+
+
+def test_anon_product_literal():
+    class Foo(m.Circuit):
+        io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit))
+
+        x = m.AnonProduct[{"I": m.Bit, "CE": m.Bit}]()
+        x @= m.product(I=io.I, CE=False)
+    assert Foo.x.CE.value() is m.GND
```
rsetaluri, line 10: nit
```suggestion
def test_anon_product_literal():

    class Foo(m.Circuit):
        io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit))
        x = m.AnonProduct[{"I": m.Bit, "CE": m.Bit}]()
        x @= m.product(I=io.I, CE=False)

    assert Foo.x.CE.value() is m.GND
```

actually do we need to have an enclosing circuit here? could we just do this with free values? e.g.

```python
a = m.Bit(name="a")
x = m.AnonProduct({"u": m.Bit, "v": m.Bit}](name="x")
x @= m.product(u=a, v=False)

assert x.v.value() is m.GND
```

rsetaluri (approved):
Few comments

PR #1236, opened 2023-02-15
[when] hotfix wiring to array with delayed elaboration

leonardt:
Issue: the latest when bit-blasting changes made an assumption that the output resolve had already happened if it was going to happen, so it doesn’t work when the output (something assigned inside the when) is resolved/elaborated after the when logic is complete

Fix: need to add logic to handle elaborated when output, the essence of this is removing the guards in the when logic that allowed the child logic to only work on inputs, however this also means the output (hardware_module) needs to also assign the final output if it's elaborated. When doing this, I added concat logic that handles arrays that are combinations of single element children and slices.  This works by using array create to combine neighboring children and concat to combine the resulting arrays including slices.  This is opposed to the original fix which flattened slices using arrayget ops and a single array concat at the end.

### DIFF
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -28,7 +28,7 @@ from magma.backend.mlir.magma_common import (
 )
 from magma.backend.mlir.mlir import (
     MlirType, MlirValue, MlirSymbol, MlirAttribute, MlirBlock, push_block,
-    push_location,
+    push_location, get_block_stack
 )
 from magma.backend.mlir.scoped_name_generator import ScopedNameGenerator
 from magma.backend.mlir.sv import sv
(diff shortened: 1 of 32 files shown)


magma/backend/mlir/hardware_module.py
```diff
@@ -640,6 +703,10 @@ def _visit(value, counter, value_to_index):
             sv.RegOp(results=[wire])
             sv.ReadInOutOp(operands=[wire], results=[result])
 
+        # Track outer_block so array_ref/constants are placed outside the always
+        # comb to reduce code repetition
+        outer_block = get_block_stack().peek()
```
leonardt, line 708: Note, this outer_block logic means that some refs/consts are appended after the always block where they are used, MLIR is okay with this (tested with mlir-verilog compile), but I may look into adding some logic to insert these before the block that references them to make the output more readable

rsetaluri (approved):

PR #1238, opened 2023-02-27
[When] Move MLIR compile logic to separate file

leonardt:
Figured I would separate this from a follow up that will reorganize the logic (so it's easier to see the code changes versus just moving to a separate file)


magma/backend/mlir/when.py
```diff
@@ -0,0 +1,241 @@
+import contextlib
```
rsetaluri, line 1: let's name this file `when_utils.py`

magma/backend/mlir/when.py
```diff
@@ -0,0 +1,241 @@
...
+)
+from magma.backend.mlir.mlir import get_block_stack, MlirValue, push_block
+from magma.backend.mlir.sv import sv
+from magma.common import sort_by_value
+from magma.primitives.when import iswhen
+from magma.ref import ArrayRef, TupleRef
+from magma.value_utils import make_selector
+
+
+def _assert_false(*args, **kwargs):
```
rsetaluri, line 18: can this be removed from the other file? if not, let's move this to `magma/common.py`

magma/backend/mlir/when.py
```diff
@@ -0,0 +1,241 @@
...
+
+        inst = module.module
+        defn = type(inst)
+        assert iswhen(defn)
+        builder = defn._builder_
+
+        input_to_index = {}
+        output_to_index = {}
+
+        def _visit(value, counter, value_to_index):
```
rsetaluri, line 34: I feel like a value of making this a class is to use member variables to avoid a closure? otherwise I think this should just be a free function -- no point in making it a class

rsetaluri (changes requested):

magma/backend/mlir/when.py
rsetaluri: nvm i see the stacked PRs

magma/common.py
```diff
@@ -339,3 +339,7 @@ def wrapped(*args, **kwargs):
         return wrapped
 
     return decorator
+
+
+def _assert_false(*args, **kwargs):
```
rsetaluri, line 344: `assert_false`

PR #1239, opened 2023-02-28
[When] Refactor MLIR code 

leonardt:
Mainly moves helper code into a standard class structure and adds some more documentation, also does some minor improvements to the logic to simplify the code.  Does not make any major architectural changes.


magma/backend/mlir/hardware_module.py
```diff
@@ -277,10 +277,13 @@ def __init__(self, graph: Graph, ctx):
         self._ctx = ctx
         self._visited = set()
 
+    def new_value(self, T):
```
rsetaluri, line 280: let's just expose `ctx` as a property rather than delegate this function up the chain

magma/backend/mlir/hardware_module.py
```diff
@@ -895,8 +898,8 @@ def visit_instance_wrapper(self, module: ModuleWrapper) -> bool:
                 return True
             return self.visit_array_slice(module)
         if inst_wrapper.name.startswith("magma_array_create_op"):
-            self._make_concat(list(reversed(module.operands)),
-                              module.results[0])
+            self.make_concat(list(reversed(module.operands)),
+                             module.results[0])
```
rsetaluri, line 902: while we're here, nit:
```suggestion
            self.make_concat(
                list(reversed(module.operands)),
                module.results[0]
            )
```

magma/backend/mlir/when.py
```diff
@@ -15,23 +15,46 @@
...
-    assert False
-
-
 class WhenCompiler:
     def __init__(self, module_visitor, module):
         self._module_visitor = module_visitor
+        self._module = module
+        self._operands = self._module.operands
+        self._flatten_all_tuples = \
+            self._module_visitor._ctx.opts.flatten_all_tuples
```
rsetaluri, line 24: nit
```suggestion
        self._flatten_all_tuples = (
            self._module_visitor._ctx.opts.flatten_all_tuples
        )
```

rsetaluri (changes requested):

rsetaluri (approved):

PR #1242, opened 2023-03-06
[MLIR] Add option to split verilog

rsetaluri:
The `split_verilog` option of the MLIR compiler annotates the output IR to directly output Verilog to specific files using the `hw.output_file` attribute.

This change is in preparation for separating monitor modules into separate Verilog files for bind2.

### DIFF
--- a/magma/backend/mlir/mlir_printer_utils.py
+++ b/magma/backend/mlir/mlir_printer_utils.py
@@ -1,6 +1,6 @@
-from typing import Callable, List, Mapping, Union
+from typing import Any, Callable, List, Mapping, Union
 
-from magma.backend.mlir.mlir import MlirValue
+from magma.backend.mlir.mlir import MlirAttribute, MlirValue
 from magma.backend.mlir.printer_base import PrinterBase
 
 
@@ -44,6 +44,12 @@ def print_signature(
     get_name = get_name_fn(raw_names)
     signatures = (f"{get_name(v)}: {v.type.emit()}" for v in value_list)
     printer.print(", ".join(signatures))
+
+
+def _attr_to_string(value: Any) -> str:
+    if isinstance(value, MlirAttribute):
+        return value.emit()
+    return str(value)
 
 
 def print_attr_dict(attr_dict: Mapping, printer: PrinterBase):
--- a/magma/backend/mlir/translation_unit.py
+++ b/magma/backend/mlir/translation_unit.py
@@ -1,9 +1,10 @@
-from typing import Any, Dict
+from typing import Any, Dict, Iterable
 import weakref
 
 from magma.backend.mlir.builtin import builtin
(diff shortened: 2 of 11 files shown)


magma/backend/mlir/mlir_printer_utils.py
```diff
@@ -46,6 +46,14 @@ def print_signature(
     printer.print(", ".join(signatures))
 
 
+def _attr_to_string(value: Any) -> str:
+    if isinstance(value, MlirAttribute):
+        return value.emit()
```
leonardt, line 51: Could we simply define `__str__` on `MlirAttribute` instead?

magma/backend/mlir/translation_unit.py
```diff
@@ -108,6 +121,15 @@ def compile(self):
...
                 if hardware_module.hw_module:
                     self.set_hardware_module(dep, hardware_module)
+        if self._opts.split_verilog:
+            if self._opts.basename is None or self._opts.suffix is None:
+                raise ValueError(
+                    "Must specify basename and suffix if split_verilog is set"
+                )
+            _prepare_for_split_verilog(
+                self._hardware_modules.values(),
+                f"{self._opts.basename}.{self._opts.suffix}",
```
leonardt, line 131: Since we're just using a specific target file here, is this any different than the current verilog export? Does that support a directory structure in the name? Would simply omitting the basename and just providing a directory here achieve the eventual result desired for bind?

magma/backend/mlir/mlir_printer_utils.py
rsetaluri: We could, but it is nice to have the str/repr inherited from dataclass since most attributes are defined as such. This helps with debugging.

The right thing to do here would be to wrap POD-type attributes in special classes, but I haven't done that yet.

magma/backend/mlir/translation_unit.py
rsetaluri: Good question. The way `--export-verilog`/`--export-split-verilog` work is as follows:

- If `--export-verilog` is specified, all `hw.output_file` attributes are ignored and verilog is output to stdout, but the output verilog is annotated with a comment that separates the various specified output files.
- If `--export-split-verilog` is specified, any module annotated with the `hw.output_file` attribute is written out to the specified file, and any un-annotated module is written out to `<module name>.sv` in the CWD. If `hw.output_file` is a directory, then it is the same as being un-annotated, in that the verilog is written out to `<dir>/<module name>.sv`.

In our case, there will be *no un-annotated modules if `split_verilog` is specified. Currently, the only annotation is for `<basename>.v` (or `.sv`). However, this PR is in preparation for separating monitor files, which requires specifying all modules.

magma/backend/mlir/mlir_printer_utils.py
rsetaluri: See #1224, will rebase

leonardt (approved):

Issue #1248, opened 2023-03-15
Verilator always comb order error


leonardt:
magma

```python
import magma as m


class Foo(m.Circuit):
    io = m.IO(I=m.In(m.Bits[8]), S=m.In(m.Bits[1]), O=m.Out(m.Bits[8]))
    x = m.Bits[8]()

    io.O @= x
    with m.when(io.S[0]):
        x @= io.I
    with m.otherwise():
        x @= ~io.I
        io.O @= ~x


m.compile("build/Foo", Foo, output="mlir-verilog")
```

mlir
```
module attributes {circt.loweringOptions = "locationInfoStyle=none"} {
    hw.module @Foo(%I: i8, %S: i1) -> (O: i8) {
        %0 = comb.extract %S from 0 : (i1) -> i1
        %2 = hw.constant -1 : i8
        %1 = comb.xor %2, %I : i8
        %4 = comb.xor %2, %3 : i8
        %6 = sv.reg : !hw.inout<i8>
        %3 = sv.read_inout %6 : !hw.inout<i8>
        %7 = sv.reg : !hw.inout<i8>
        %5 = sv.read_inout %7 : !hw.inout<i8>
        sv.alwayscomb {
            sv.bpassign %7, %3 : i8
            sv.if %0 {
                sv.bpassign %6, %I : i8
            } else {
                sv.bpassign %6, %1 : i8
                sv.bpassign %7, %4 : i8
            }
        }
        hw.output %5 : i8
    }
}
```

verilog
```verilog
// Generated by CIRCT unknown git version
module Foo(
  input  [7:0] I,
  input        S,
  output [7:0] O);

  reg [7:0] _GEN;
  reg [7:0] _GEN_0;
  always_comb begin
    _GEN_0 = _GEN;
    if (S)
      _GEN = I;
    else begin
      _GEN = ~I;
      _GEN_0 = ~_GEN;
    end
  end // always_comb
  assign O = _GEN_0;
endmodule
```

verilator error
```shell
❯ verilator --lint-only build/Foo.v
%Warning-ALWCOMBORDER: build/Foo.v:12:7: Always_comb variable driven after use: '_GEN'
                                       : ... In instance Foo
   12 |       _GEN = I;
      |       ^~~~
                       ... Use "/* verilator lint_off ALWCOMBORDER */" and lint_on around source to disable this message.
[...]

PR #1249, opened 2023-03-21
[When] Temporary Guard Against ALWCOMBORDER issue

leonardt:
See https://github.com/phanrahan/magma/issues/1248 for context, adds a pass that checks whether this issue would be possible and raises an exception.

Ideally we'd support the ok version of this which is when the variable in the loop is assigned before it is used, but this requires some more complex analysis which we can add incrementally (it's also a significant step towards a more general solution that manages these dependencies).

### DIFF
--- a/magma/backend/mlir/build_magma_graph.py
+++ b/magma/backend/mlir/build_magma_graph.py
@@ -15,8 +15,9 @@ from magma.bits import Bits
 from magma.circuit import DefineCircuitKind
 from magma.compile_exception import UnconnectedPortException
 from magma.digital import Digital
+from magma.primitives.register import Register
 from magma.protocol_type import magma_value as get_magma_value
-from magma.ref import InstRef, DefnRef, AnonRef, ArrayRef, TupleRef, LazyInstRef
+from magma.ref import InstRef, DefnRef, AnonRef, ArrayRef, TupleRef
 from magma.t import Type
 from magma.tuple import Tuple as m_Tuple
 
@@ -173,6 +174,19 @@ def _visit_inputs(
         )
 
 
+def _check_no_when_loops(graph: Graph):
+    """Temporary guard against https://github.com/phanrahan/magma/issues/1248"""
+    for node in graph.nodes():
+        if not getattr(node, '_is_when_builder_', False):
+            continue
+        for predecessor in graph.predecessors(node):
+            if isinstance(type(node), Register):
+                # broken by register
+                break
+            if predecessor is node:
+                raise NotImplementedError("Dependency loop in when")
+
+
 def build_magma_graph(
         ckt: DefineCircuitKind,
         opts: BuildMagmaGrahOpts = BuildMagmaGrahOpts()) -> Graph:
@@ -180,4 +194,5 @@ def build_magma_graph(
     _visit_inputs(ctx, ckt, opts.flatten_all_tuples)
     for inst in ckt.instances:
         _visit_inputs(ctx, inst, opts.flatten_all_tuples)
+    _check_no_when_loops(ctx.graph)
     return ctx.graph
--- a/magma/backend/mlir/build_magma_graph.py
+++ b/magma/backend/mlir/build_magma_graph.py
@@ -2,6 +2,7 @@ import dataclasses
 from typing import Any, Callable, Tuple
 
 from magma.array import Array
+from magma.backend.mlir.errors import MlirWhenCycleError
 from magma.backend.mlir.graph_lib import Graph
 from magma.backend.mlir.magma_common import (
     ModuleLike, visit_value_or_value_wrapper_by_direction, safe_root,
@@ -16,6 +17,7 @@ from magma.circuit import DefineCircuitKind
 from magma.compile_exception import UnconnectedPortException
 from magma.digital import Digital
 from magma.primitives.register import Register
+from magma.primitives.when import iswhen
 from magma.protocol_type import magma_value as get_magma_value
 from magma.ref import InstRef, DefnRef, AnonRef, ArrayRef, TupleRef
 from magma.t import Type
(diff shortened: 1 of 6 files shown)


rsetaluri:
Let's squash all these commits and rebase and force-push. Otherwise the merged and squashed PR's commit message will be pretty messy

rsetaluri (commented):
Is it possible to have this pass run on the magma representation independent of MLIR? It seems like a general pass on circuits with when.

rsetaluri:
> Is it possible to have this pass run on the magma representation independent of MLIR? it seems like a general pass on the representation



> Is it possible to have this pass run on the magma representation independent of MLIR? it seems like a general pass on the representation?



> Is it possible to have this pass run on the magma representation independent of MLIR? it seems like a general pass on circuits with when

It's possible that the graph view is actually also more general, and can be used as a type of pass.

rsetaluri:
> Is it possible to have this pass run on the magma representation independent of MLIR? It seems like a general pass on circuits with when.

It's possible that the graph view we're using for the MLIR backend is actually a generally useful representation for passes. We could extract it out and have this check run on that graph.

leonardt:
I do think the graph logic can and should be generic, but

1. this check is specific to the MLIR backend because that's where there's a problem with compilation
2. this is temporary until compilation is fixed

So I'd advocate for not blocking it by the refactoring of the graph logic, which could seems like an orthogonal change to the issue being addressed.

rsetaluri:
> 1. this check is specific to the MLIR backend because that's where there's a problem with compilation
It seems like this *could* cause problems with *any* backend?  That is to say, it represents a mismatch in our translation to Verilog more generally?

> 2. this is temporary until compilation is fixed
> 
> So I'd advocate for not blocking it by the refactoring of the graph logic, which could seems like an orthogonal change to the issue being addressed.
Fine with this.

magma/backend/mlir/build_magma_graph.py
```diff
@@ -173,11 +174,25 @@ def _visit_inputs(
...
+
+
 def build_magma_graph(
         ckt: DefineCircuitKind,
         opts: BuildMagmaGrahOpts = BuildMagmaGrahOpts()) -> Graph:
     ctx = ModuleContext(Graph(), opts)
     _visit_inputs(ctx, ckt, opts.flatten_all_tuples)
     for inst in ckt.instances:
         _visit_inputs(ctx, inst, opts.flatten_all_tuples)
+    _check_no_when_loops(ctx.graph)
```
rsetaluri, line 197: note need a special subclass of `MlirCompilerError` (in `magma/backend/mlir/errors.py`) called `MlirWhenCycleError` (maybe better phrasing is needed :D )

```suggestion
def _check_for_when_cycles(graph: Graph):
    """Temporary guard against https://github.com/phanrahan/magma/issues/1248."""
    for node in graph.nodes():
        if not is_when_builder(node):
            continue
        for predecessor in graph.predecessors(node):
            if isinstance(type(node), Register):  # cycles are broken by registers
                break
            if predecessor is node:
                raise MlirWhenCycleError()


def build_magma_graph(
        ckt: DefineCircuitKind,
        opts: BuildMagmaGrahOpts = BuildMagmaGrahOpts()) -> Graph:
    ctx = ModuleContext(Graph(), opts)
    _visit_inputs(ctx, ckt, opts.flatten_all_tuples)
    for inst in ckt.instances:
        _visit_inputs(ctx, inst, opts.flatten_all_tuples)
    _check_for_when_cycles(ctx.graph)
```

magma/primitives/when.py
```diff
@@ -139,6 +139,7 @@ def __init__(self, block: WhenBlock):
         self._set_definition_attr(_ISWHEN_KEY, True)
         self._set_definition_attr("_builder_", self)
         self._is_when_builder_ = True
+        self._set_definition_attr("_is_when_builder_", True)
```
rsetaluri, line 142: let's add a separate free function `is_when_builder` which is just `def is_when_builder(obj) -> bool: return getattr(obj, "_is_when_builder_", False)`

can also additionally abstract `"_is_when_builder_"` key into `_IS_WHEN_BUILDER_KEY` constant at the top.

tests/test_when.py
```diff
@@ -465,8 +465,10 @@ class _Test(m.Circuit):
         io.O0 @= x
 
     basename = "test_when_recursive_non_port"
-    m.compile(f"build/{basename}", _Test, output="mlir")
-    assert check_gold(__file__, f"{basename}.mlir")
+    # TODO(leonardt): Remove this when proper analysis is added
```
rsetaluri, line 468: nit
```suggestion
    # TODO(leonardt): Remove this when proper analysis for when cycles is added.
```

tests/test_when.py
```diff
@@ -1456,6 +1458,23 @@ class test_when_tuple_as_bits_resolve(m.Circuit):
...
+        x = m.Bits[8]()
+
+        io.O @= x
+        with m.when(io.S[0]):
+            x @= io.I
+        with m.otherwise():
+            x @= ~io.I
+            io.O @= ~x
+
+    with pytest.raises(NotImplementedError):
```
rsetaluri, line 1473: will need to change this error

tests/test_when.py
```diff
@@ -1456,6 +1458,23 @@ class test_when_tuple_as_bits_resolve(m.Circuit):
     assert check_gold(__file__, "test_when_tuple_as_bits_resolve.mlir")
 
 
+def test_when_alwcomb_order():
+    class test_when_alwcomb_order(m.Circuit):
```
rsetaluri, line 1462: nit
```suggestion
def test_when_alwcomb_order():

    class test_when_alwcomb_order(m.Circuit):
```

tests/test_when.py
```diff
@@ -1456,6 +1458,23 @@ class test_when_tuple_as_bits_resolve(m.Circuit):
...
+        io.O @= x
+        with m.when(io.S[0]):
+            x @= io.I
+        with m.otherwise():
+            x @= ~io.I
+            io.O @= ~x
+
+    with pytest.raises(NotImplementedError):
+        m.compile("build/test_when_alwcomb_order", test_when_alwcomb_order,
+                  output="mlir")
```
rsetaluri, line 1475:
```suggestion
        m.compile(
            "build/test_when_alwcomb_order",
            test_when_alwcomb_order,
            output="mlir"
        )
```

rsetaluri (changes requested):

rsetaluri (approved):

PR #1255, opened 2023-03-23
[Bind2] Refactor XMR references

rsetaluri:
Avoids requiring setting instances as attributes on circuits to refer to them (similarly with `self.` for Generator2's). Instead getting the attribute will retrieve the instance by name.

For example,

```python
class Foo(m.Circuit):
    my_placeholder_var = Bar(name="my_instance_name")

assert Foo.my_placeholder_var is Foo.my_instance_name
```

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -27,6 +27,7 @@ except ImportError:
     pass
 
 from magma.clock import is_clock_or_nested_clock, Clock, ClockTypes
+from magma.common import only, IterableException
 from magma.config import get_debug_mode, set_debug_mode, config, RuntimeConfig
 from magma.definition_context import (
     DefinitionContext,
@@ -40,7 +41,6 @@ from magma.logging import root_logger, capture_logs
 from magma.protocol_type import MagmaProtocol
 from magma.ref import TempNamedRef, AnonRef
 from magma.t import In, Type
-from magma.view import PortView
 from magma.wire_container import WiringLog, AggregateWireable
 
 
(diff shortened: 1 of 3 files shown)


rsetaluri:
@leonardt my main concern is the performance cost of doing this lookup every time `getattr` is called. We could add an acceleration data structure (map from names to instances) to speed this up. Although, currently we don't enforce uniqueness of instance names, and that would implicitly enforce that. Open to suggestions here.

magma/circuit.py
```diff
@@ -735,6 +735,13 @@ def __new__(metacls, name, bases, dct):
 
         return self
 
+    def __getattr__(self, attr):
+        try:
+            return only(filter(lambda i: i.name == attr, self.instances))
+        except IterableException:
+            pass
+        return object.__getattribute__(self, attr)
```
leonardt, line 743: Could we invert this, do `__getattribute__` by default, and check instances list on AttributeError? That would optimize the common case (save a level of indirection in the function call stack)

leonardt:
Alternatively, we could try patching the instance name attributes at close time? Assuming the .open() pattern is used for modifications, this should be safe?

magma/circuit.py
leonardt: I think the above behavior is expected given normal Python class/object semantics so I'd advocate for preserving that and having a get API specifically for instances when there is this issue of collisions.

rsetaluri: Good point. See #1262

PR #1257, opened 2023-03-27
[InlineVerilog] Add support for outputs

leonardt:
* Updates instance logic to add output driver
* Updates MLIR compile to handle outputs
  * Emits register to hold instance output value
  * Adds output  value in verbatim substitution

### DIFF
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -831,7 +831,9 @@ class ModuleVisitor:
             input_counter = itertools.count()
             output_counter = itertools.count()
             for value in references.values():
-                if value.is_input():
+                if isinstance(value, PortView):
+                    value = value.port
+                if value.is_output():
                     operand = module.operands[next(input_counter)]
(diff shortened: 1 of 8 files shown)


rsetaluri:
I'm not sure what the intention of this change is. Is it to be able to drive outputs from inline_verilog? Is it to be able to use the value of outputs before they're driven (e.g. where `trace()` currently returns `None`)?

In the generated Verilog for the updated test (below), it looks like `O` has a hanging output -- unless I'm misunderstanding the semantics of `assert` and it can drive values as well.

```Verilog
  reg _GEN;
  assert property (@(posedge CLK) I |-> ##1 _GEN);
  assign O = _GEN;
```

If we want to be able to drive values from outside, we should use something like `InlineVerilogExpression` (more importantly we should use `sv.verbatim.expr`).

leonardt:
Oops, yes the test should be updated to actually drive the output rather than just refer to it.  I'll see about using verbatim.expr instead.

Issue #1252, opened 2023-03-22
[MLIR] Tuple (integer) key names not allowed


leonardt:
```python
import magma as m


class Foo(m.Circuit):
    T = m.Tuple[m.Bit, m.Bits[8]]
    io = m.IO(I=m.In(T), O=m.Out(T))
    io.O[0] @= ~io.I[0]
    io.O[1] @= io.I[1] ^ 0xFF


m.compile("build/Foo", Foo, output="mlir-verilog")
```

```shell
<stdin>:2:35: error: expected valid keyword
    hw.module @Foo(%I: !hw.struct<0: i1, 1: i8>) -> (O: !hw.struct<0: i1, 1: i8>) {
                                  ^
```

CIRCT doesn't support integer key names for generating verilog

rsetaluri:
Fixed by #1260

PR #1260, opened 2023-04-06
[MLIR] Add support for Tuples with integer keys

rsetaluri:
See #1252.

Fixes this issue by prefixing tuple field keys with "_". This must be in sync between:
* The emission of struct types
* The emission of struct extraction

Note that struct creation does not need to be aware of this.

We use a common `tuple_key_to_str()` function (still specific to MLIR) to achieve this.


magma/backend/mlir/hardware_module.py
```diff
@@ -138,9 +138,21 @@ def magma_type_to_mlir_type(type: Kind) -> MlirType:
...
+            for k, t in type.field_dict.items()
+        }
+        return hw.StructType(tuple(fields.items()))
     if issubclass(type, m_Tuple):
-        fields = {str(k): magma_type_to_mlir_type(t)
-                  for k, t in type.field_dict.items()}
+        fields = {
+            f"_{i}": magma_type_to_mlir_type(t)
+            for i, t in enumerate(type.field_dict.values())
+        }
```
leonardt, line 155: How about using something like a `_tuple_key_to_str` like:

```suggestion
        def _tuple_key_to_str(k, type):
            return str(k) if issubclass(type, Product) else f"_{k}"

        fields = {_tuple_key_to_str(k, type): magma_type_to_mlir_type(t)
                  for k, t in type.field_dict.items()}
```

leonardt (approved):
Looks good, but I wonder if we could somehow encapsulate the retrieval of a safe index string into the type? E.g. add an API to tuple/product to get a "safe_index" for each field (where tuple would prepend the underscore).   This would avoid having to deal with the Product vs Tuple check in the backend and allows for future extensibility (e.g. a user could subclass Tuple and define custom behavior, but they may need the ability to provide a "safe_index").

magma/backend/mlir/hardware_module.py
rsetaluri: :+1:

PR #1265, opened 2023-04-21
[Circuit] Add get_instance method

rsetaluri:
Addresses #1262.

### DIFF
--- a/magma/circuit.py
+++ b/magma/circuit.py
@@ -27,7 +27,12 @@ except ImportError:
     pass
 
 from magma.clock import is_clock_or_nested_clock, Clock, ClockTypes
-from magma.common import only, IterableException
+from magma.common import (
+    only,
+    IterableException,
+    EmptyIterableException,
+    NonSingletonIterableException,
+)
 from magma.config import get_debug_mode, set_debug_mode, config, RuntimeConfig
 from magma.definition_context import (
     DefinitionContext,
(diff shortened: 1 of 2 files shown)


magma/circuit.py
```diff
@@ -758,6 +763,15 @@ def is_definition(self):
     def instances(self):
         return self._context_.placer.instances()
 
+    def get_instance(self, name: str) -> 'AnonymousCircuitType':
+        instances = self.instances
+        try:
+            return only(filter(lambda i: i.name == name, instances))
```
leonardt, line 769: Could refactor this into a helper function to share code with https://github.com/phanrahan/magma/pull/1263/files#diff-889a16f7db04ed3ceb8ae4a89693b806f06d3e0c3894022ed50999a1831ce874R752

magma/circuit.py
```diff
@@ -758,6 +763,15 @@ def is_definition(self):
...
+        except EmptyIterableException:
+            raise KeyError(name) from None
+        except NonSingletonIterableException:
+            raise KeyError(f"Found multiple instances with name '{name}'")
```
leonardt, line 773: Would these get "uniquified" in the downstream verilog? Otherwise should we make this illegal in magma? Then we could consider changing the instances container to be a map.  Alternatively, we could uniquify the names at the magma level to avoid this issue?

leonardt (changes requested):

magma/circuit.py
rsetaluri: As of now, nothing in the magma representation requires the constraint that instances have unique names. In fact, it is even valid MLIR to have duplicate instance names, e.g.

```mlir
    hw.module @Bar(%I: i1, %CLK: i1) -> (O: i1) {
        %0 = hw.instance "foo" @Foo(I: %I: i1) -> (O: i1)
        %1 = hw.instance "foo" @Foo(I: %0: i1) -> (O: i1)
        hw.output %1 : i1
    }
```

magma/circuit.py
rsetaluri: Good idea - but I kind of like that we can use these functions in an almost functional pattern, and it's a single line. Think refactoring would increase the total number of lines :).

leonardt (approved):

PR #1267, opened 2023-05-04
[When] Add support for flatten_all_tuples=False

### LINKED ISSUE #1253: [MLIR] when support for flatten_all_tuples
```diff
--- a/tests/test_when.py
+++ b/tests/test_when.py
@@ -342,7 +342,7 @@ def test_memory(T, bits_to_fault_value):
             io.out @= m.from_bits(T, m.Bits[8](0xFF))
 
     m.compile(f"build/test_when_memory_{T_str}", test_when_memory,
-              output="mlir", flatten_all_tuples=True)
+              output="mlir")
 
     if check_gold(__file__, f"test_when_memory_{T_str}.mlir"):
         return
@@ -377,7 +377,6 @@ def test_memory(T, bits_to_fault_value):
     tester.compile_and_run("verilator", magma_output="mlir-verilog",
                            directory=os.path.join(os.path.dirname(__file__),
                                                   "build"),
-                           magma_opts={"flatten_all_tuples": True},
                            flags=['-Wno-UNUSED'])
 
     update_gold(__file__, f"test_when_memory_{T_str}.mlir")
```

This first needs a fix in the tuple pipeline
```diff
--- a/magma/backend/mlir/magma_ops.py
+++ b/magma/backend/mlir/magma_ops.py
@@ -8,11 +8,11 @@ from magma.digital import DigitalMeta
 from magma.generator import Generator2
 from magma.interface import IO
 from magma.t import In, Out
-from magma.tuple import TupleMeta, ProductMeta
+from magma.tuple import TupleMeta, AnonymousProductMeta


 def _get_tuple_field_type(T: TupleMeta, index: Union[int, str]):
-    if isinstance(T, ProductMeta):
+    if isinstance(T, AnonymousProductMeta):
         return T.field_dict[index]
     index = int(index)
     return T.field_dict[index]
```

Then we run into

```
magma/compile.py:104: in compile
    result = compiler.compile()
magma/backend/mlir/mlir_compiler.py:58: in compile
    compile_to_mlir(
magma/backend/mlir/compile_to_mlir.py:24: in compile_to_mlir
    translation_unit.compile()
magma/backend/mlir/translation_unit.py:145: in compile
    hardware_module.compile()
magma/backend/mlir/hardware_module.py:1195: in compile
    self._hw_module = self._compile()
magma/backend/mlir/hardware_module.py:1254: in _compile
    visitor.visit(self._magma_defn_or_decl)
magma/backend/mlir/hardware_module.py:950: in visit
    self._process_magma_module(module)
magma/backend/mlir/hardware_module.py:939: in _process_magma_module
    self.visit(predecessor)
magma/backend/mlir/hardware_module.py:951: in visit
    self.visit_module(ModuleWrapper.make(module, self._ctx))
magma/backend/mlir/common.py:11: in wrapped
    ret = fn(*args, **kwargs)
magma/backend/mlir/hardware_module.py:929: in visit_module
    return self.visit_instance(module)
magma/backend/mlir/common.py:11: in wrapped
    ret = fn(*args, **kwargs)
magma/backend/mlir/hardware_module.py:850: in visit_instance
    return self.visit_primitive(module)
magma/backend/mlir/common.py:11: in wrapped
    ret = fn(*args, **kwargs)
magma/backend/mlir/hardware_module.py:677: in visit_primitive
    return self.visit_when(module)
magma/backend/mlir/common.py:11: in wrapped
    ret = fn(*args, **kwargs)
magma/backend/mlir/hardware_module.py:656: in visit_when
    return WhenCompiler(self, module).compile()
magma/backend/mlir/when_utils.py:282: in compile
    self._process_when_block(self._builder.block)
magma/backend/mlir/when_utils.py:265: in _process_when_block
    self._process_connections(block)
magma/backend/mlir/when_utils.py:246: in _process_connections
    self._make_assignments(connections)
magma/backend/mlir/when_utils.py:235: in _make_assignments
    for wire, value in self._build_wire_map(connections).items():
magma/backend/mlir/when_utils.py:174: in _build_wire_map
    operand = self._get_operand(driver_elt)
magma/backend/mlir/when_utils.py:46: in _get_operand
    return self._operands[self._get_input_index(value)]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <magma.backend.mlir.when_utils.WhenCompiler object at 0x7fa1d1258400>, value = Memory_inst0.RDATA.x

    def _get_input_index(self, value):
>       return self._input_to_index[value]
E       KeyError: Memory_inst0.RDATA.x
```

leonardt:
Fixes #1253


leonardt (commented):
@rsetaluri I needed this change to get xmr working with unflattened tuples:  

The assert is no longer valid because the paths are flattened but the module outputs aren't (we could compare to the flattened results length).

The path separator changes because the tuples values aren't flattened anymore (need to do a struct reference).

Does this make sense or am I misunderstanding what's going on? I will update the path separator logic to based on the flatten_all_tuples option 
The path string should use `.` instead of `_` since the fields aren't flattened.

I'm wondering:

leonardt:
^ this comment is in reference to https://github.com/phanrahan/magma/pull/1267/commits/c6d802d97496bdc6b854460c86935e9a4be2e977

leonardt:
Updated inline_verilog2 test to check both cases, skipped inline_verilog since we're deprecating that anyways

magma/backend/mlir/hardware_module.py
```diff
@@ -143,7 +143,7 @@ def magma_type_to_mlir_type(type: Kind) -> MlirType:
     if issubclass(type, m_Tuple):
         fields = (
             (magma_tuple_key_to_str(k, type), magma_type_to_mlir_type(t))
-            for k, t in type.field_dict.items()
+            for k, t in type.sorted_field_dict_items()
```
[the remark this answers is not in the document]
leonardt: I was running into an issue where struct fields were being generated in a different order in different places in the generated MLIR.  Perhaps there some places where we're transferring the data through some structure that changes the order or something, I can see if I can trace down the root cause location.

rsetaluri:
> I can see if I can trace down the root cause location.

Looks good modulo finding this. Let me know what you find.

magma/backend/mlir/hardware_module.py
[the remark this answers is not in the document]
leonardt: So I think the issue is not that we create any incompatible orderings internally,  but that the user could produce wireable types with an unordered data structure, and since we do structural wiring, this is okay.  Here's some example code that works
```
x = {"x": 0, "y": 1}
y = {"y": 0, "x": 1}

assert list(x.keys()) != list(y.keys())
assert x.keys() == y.keys()


import magma as m

T0 = m.AnonProduct[{"x": m.Bit, "y": m.Bits[8]}]
T1 = m.AnonProduct[{"y": m.Bits[8], "x": m.Bit}]


class Foo(m.Circuit):
    io = m.IO(I=m.In(T0), O=m.Out(T0))
    io.O @= io.I


m.compile("build/Foo", Foo, output="mlir-verilog")
```

Now MLIR/CIRCT on the other hand, requires the tuple fields to be consistent in order across the IR code, so while in magma we may wire types with different orderings, we need to sort them for MLIR.

I will add a test case that covers this.

leonardt:
Found an issue related to key names, currently in magma it's okay to do:
```python
T = m.AnonProduct[{"0": m.Bit, "1": m.Bits[8]}]


class product_key_sanitize(m.Circuit):
    io = m.IO(I=m.In(T), O=m.Out(T))
    io.O @= io.I
```

However, since this is using the product code path, the integer key names aren't sanitized.

```
module attributes {circt.loweringOptions = "locationInfoStyle=none"} {
    hw.module @product_key_sanitize(%I: !hw.struct<0: i1, 1: i8>) -> (O: !hw.struct<0: i1, 1: i8>) {
        hw.output %I : !hw.struct<0: i1, 1: i8>
    }
}
```

```
tests/test_backend/test_mlir/golds/product_key_sanitize.mlir:2:52: error: expected valid keyword
    hw.module @product_key_sanitize(%I: !hw.struct<0: i1, 1: i8>) -> (O: !hw.struct<0: i1, 1: i8>) {
                                                   ^
```

examples/riscv_mini/tests/test_unflattened_tuples.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_riscv_mini_unflattened_tuples():
+    magma_config.compile_dir = 'callee_file_dir'
+    tile = Tile(32)
+    m.compile("build/test_riscv_mini_unflattened_tuples", tile,
+              output="mlir-verilog", disallow_local_variables=True,
+              disallow_packed_struct_assignments=True)
+    assert check_gold(__file__,
+                      "test_riscv_mini_unflattened_tuples.v")
```
rsetaluri, line 17:
```suggestion
    m.compile(
        "build/test_riscv_mini_unflattened_tuples",
        tile,
        output="mlir-verilog",
        disallow_local_variables=True,
        disallow_packed_struct_assignments=True,
    )
    assert check_gold(__file__, "test_riscv_mini_unflattened_tuples.v")
```

magma/backend/mlir/mlir_compiler.py
```diff
@@ -45,7 +45,11 @@ def run_pre_uniquification_passes(self):
         # could introduce more conditional logic, or they could
         # trigger elaboration on values used in existing coditiona
         # logic (which modifies the when builder).
-        run_when_passes(self.main, self.opts.get("emit_when_assertions", False))
+        run_when_passes(
+            self.main,
```
rsetaluri, line 49: btw, we should just use `self._compile_to_mlir_opts.{flatten_all_tuples,emit_when_assertions}` here. same above (L42). Can do this in a separate PR if desired.

magma/backend/mlir/translation_unit.py
```diff
@@ -29,6 +29,8 @@ def _set_module_attrs(mlir_module: builtin.ModuleOp, opts: CompileToMlirOpts):
         lowering_options.append("disallowLocalVariables")
     if opts.omit_version_comment:
         lowering_options.append("omitVersionComment")
+    if opts.disallow_packed_struct_assignments:
```
rsetaluri, line 32: let's factor this out into a separate PR to add this option

magma/backend/mlir/when_utils.py
```diff
@@ -135,7 +139,11 @@ def __init__(self):
                 self.index = tuple()
 
             def __getitem__(self, idx):
-                self.index += (idx, )
```
rsetaluri, line 138: let's move this class outside of the function so that it doesn't get constructed on every function call. can modify the class name if it could collide with something else.

magma/backend/mlir/xmr_utils.py
```diff
@@ -54,7 +55,8 @@ def _ascend_to_leaf(value, leaves):
     ref = value.name
     if isinstance(ref, TupleRef):
         path = _ascend_to_leaf(ref.tuple, leaves)
-        return path + [f".{ref.index}"]
+        key = tuple_key_to_str(ref.index, type(ref.tuple))
+        return path + [f".{key}"]
```
rsetaluri, line 59:
```suggestion
        return path + [f".{tuple_key_to_str(ref.index, type(ref.tuple))}"]
```

magma/backend/mlir/xmr_utils.py
```diff
@@ -91,25 +93,27 @@ def get_xmr_paths(ctx: 'HardwareModule', xmr: PortView) -> List[Tuple[str]]:
     #     leaf to the value referred to, joined with ".", since these are not
     #     flattened.
     root = xmr.port.name.root().value()
+    flatten_all_tuples = ctx.opts.flatten_all_tuples
```
rsetaluri, line 96: no need for this for just 2 references
```suggestion
```

magma/backend/mlir/xmr_utils.py
```diff
@@ -91,25 +93,27 @@ def get_xmr_paths(ctx: 'HardwareModule', xmr: PortView) -> List[Tuple[str]]:
...
     tree = magma_value_or_value_wrapper_to_tree(
-        root, flatten_all_tuples=ctx.opts.flatten_all_tuples)
+        root, flatten_all_tuples=flatten_all_tuples)
```
rsetaluri, line 98:
```suggestion
        root, flatten_all_tuples=ctx.opts.flatten_all_tuples)
```

magma/backend/mlir/xmr_utils.py
```diff
@@ -91,25 +93,27 @@ def get_xmr_paths(ctx: 'HardwareModule', xmr: PortView) -> List[Tuple[str]]:
...
     assert tree.has_node(root)
 
+    separator = "_" if flatten_all_tuples else "."
```
rsetaluri, line 101:
```suggestion
    separator = "_" if ctx.opts.flatten_all_tuples else "."
```

magma/passes/when.py
```diff
@@ -16,8 +16,12 @@ def process_when_builder(self, builder, defn):
 
 
 class EmitWhenAsserts(WhenPass):
+    def __init__(self, main, flatten_all_tuples):
+        super().__init__(main)
+        self.flatten_all_tuples = flatten_all_tuples
```
rsetaluri, line 21:
```suggestion
        self._flatten_all_tuples = flatten_all_tuples
```

magma/passes/when.py
```diff
@@ -16,8 +16,12 @@ def process_when_builder(self, builder, defn):
...
+
     def process_when_builder(self, builder, defn):
-        builder.emit_when_assertions()
+        builder.emit_when_assertions(self.flatten_all_tuples)
```
rsetaluri, line 24:
```suggestion
        builder.emit_when_assertions(self._flatten_all_tuples)
```

magma/when.py
```diff
@@ -565,13 +565,14 @@ def _get_builder_port(value, builder):
     return port
 
 
-def _emit_when_assert(cond, drivee, driver, builder, assignment_map):
-    if _contains_tuple(type(drivee)):
+def _emit_when_assert(cond, drivee, driver, builder, assignment_map,
+                      flatten_all_tuples):
+    if _contains_tuple(type(drivee)) and flatten_all_tuples:
```
rsetaluri, line 570:
```suggestion
def _emit_when_assert(
    cond, drivee, driver, builder, assignment_map, flatten_all_tuples
):
    if _contains_tuple(type(drivee)) and flatten_all_tuples:
```

magma/when.py
```diff
@@ -565,13 +565,14 @@ def _get_builder_port(value, builder):
...
         # Since tuples are elaborated in verilog, we emit an assert for the
         # leaf values.
-        # TODO(leonardt): Update this when we remove flatten_all_tuples.
         for x, y in zip(drivee, driver):
-            _emit_when_assert(cond, x, y, builder, assignment_map)
+            _emit_when_assert(cond, x, y, builder, assignment_map,
+                              flatten_all_tuples)
```
rsetaluri, line 575:
```suggestion
            _emit_when_assert(
                cond, x, y, builder, assignment_map, flatten_all_tuples
            )
```

magma/when.py
```diff
@@ -607,19 +608,21 @@ def _make_else_cond(block, precond):
     return ~block.condition
 
 
-def _emit_default_driver_asserts(block, builder, assignment_map):
+def _emit_default_driver_asserts(block, builder, assignment_map,
+                                 flatten_all_tuples):
```
rsetaluri, line 612:
```suggestion
def _emit_default_driver_asserts(
    block, builder, assignment_map, flatten_all_tuples
):
```

magma/when.py
```diff
@@ -607,19 +608,21 @@ def _make_else_cond(block, precond):
...
     for wire in list(block.root.default_drivers()):
         port = _get_builder_port(wire.drivee, builder)
         assigned_conds = assignment_map[port]
         if not assigned_conds:
             continue
         cond = ~functools.reduce(operator.or_, assigned_conds)
         _emit_when_assert(
-            cond, wire.drivee, wire.driver, builder, defaultdict(list)
+            cond, wire.drivee, wire.driver, builder, defaultdict(list),
+            flatten_all_tuples
```
rsetaluri, line 621:
```suggestion
            cond,
            wire.drivee,
            wire.driver,
            builder,
            defaultdict(list),
            flatten_all_tuples
```

magma/when.py
```diff
@@ -607,19 +608,21 @@ def _make_else_cond(block, precond):
...
         )
 
 
-def emit_when_assertions(block, builder, precond=None,
+def emit_when_assertions(block, builder, flatten_all_tuples, precond=None,
                          assignment_map=defaultdict(list)):
```
rsetaluri, line 626:
```suggestion
def emit_when_assertions(
    block,
    builder,
    flatten_all_tuples,
    precond=None,
    assignment_map=defaultdict(list)
):
```

magma/when.py
```diff
@@ -634,18 +637,22 @@ def emit_when_assertions(block, builder, precond=None,
...
-            assignment_map
+            assignment_map,
+            flatten_all_tuples
         )
 
     for child in block.children():
         # Children inherit this block's cond as a precond.
-        emit_when_assertions(child, builder, cond, assignment_map)
+        emit_when_assertions(child, builder, flatten_all_tuples, cond,
+                             assignment_map)
```
rsetaluri, line 647:
```suggestion
        emit_when_assertions(
            child, builder, flatten_all_tuples, cond, assignment_map
        )
```

magma/when.py
```diff
@@ -634,18 +637,22 @@ def emit_when_assertions(block, builder, precond=None,
...
     else_block = _get_else_block(block)
     if else_block:
         else_cond = _make_else_cond(block, precond)
-        emit_when_assertions(else_block, builder, else_cond, assignment_map)
+        emit_when_assertions(else_block, builder, flatten_all_tuples,
+                             else_cond, assignment_map)
```
rsetaluri, line 652:
```suggestion
        emit_when_assertions(
            else_block, builder, flatten_all_tuples, else_cond, assignment_map
        )
```

magma/when.py
```diff
@@ -634,18 +637,22 @@ def emit_when_assertions(block, builder, precond=None,
...
     if block is block.root:
-        _emit_default_driver_asserts(block, builder, assignment_map)
+        _emit_default_driver_asserts(block, builder, assignment_map,
+                                     flatten_all_tuples)
```
rsetaluri, line 655:
```suggestion
        _emit_default_driver_asserts(
            block, builder, assignment_map, flatten_all_tuples
        )
```

tests/test_verilog/test_inline.py
```diff
@@ -103,7 +103,8 @@ class Main(m.Circuit):
             ready_out=delay.inner_delay.inner_inner_delay.INPUT[1].ready,
         )
 
-    m.compile(f"build/test_inline_tuple", Main, output="mlir")
+    m.compile(f"build/test_inline_tuple", Main, output="mlir",
```
rsetaluri, line 106: can we have this parametrized for both flatten_all_tuples = False, True?

tests/test_when.py
```diff
@@ -11,6 +11,10 @@
 from magma.backend.mlir.errors import MlirWhenCycleError
 
 
+# Change this to validate mlir->verilog compile doesn't error
+_OUTPUT_TYPE = "mlir"
```
rsetaluri, line 15: not sure what the purpose of this is? we should either be testing mlir or mlir-verilog. if we need to test the latter then it should just be that

rsetaluri (changes requested):

magma/backend/mlir/hardware_module.py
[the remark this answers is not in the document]
rsetaluri: Got it, thanks for hunting this down.

I feel like there should be something more explicit here. Ideally what is allowed in magma is consistent with what is allowed by the backend.

Here are some options (all of which btw, I think should be implemented in a separate PR fixing this):

1. Disallow magma from wiring tuples with keys in different orders. So in the above case `io.O @= io.I` would itself raise an error -- the same way that `m.Bit() @= m.Bits[8]()` would even raise an error -- just two different types that can't be wired.
2. All tuple keys should be inherently sorted. So `T0` and `T1` are the *same* type inherently (but multiple different constructor calls could lead to the same type). That way MLIR is agnostic to this. Right now the backend *silently* re-orders/sorts the keys. The keys should either always be considered ordered or not -- not half and half.

Don't have a strong opinion on this. (2) seems easier but it will likely break a lot of things, since it will re-order all keys. (1) seems somewhat difficult to detect...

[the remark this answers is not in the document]
rsetaluri: See #1314

tests/test_verilog/test_inline.py
leonardt: Don't think this is necessary since it's going to be deprecated for inline_verilog2?

leonardt: (The test is parametrized for inline_verilog2)

tests/test_when.py
leonardt: I use it locally to ensure the compilation doesn't fail, this allows me to change the output for the entire file with a single variable, but we leave it on mlir for CI for speed

magma/backend/mlir/translation_unit.py
leonardt: I'm not sure if this makes sense.  We need to compile with flatten_all_tuples=False to test it, otherwise the option isn't useful (since we wouldn't generate any packed struct assignments), so while we could add it as a separate PR, we wouldn't be able to have a meaningful test.  Adding it as a follow up PR also doesn't quite make sense either since this feature requires the option to be tested with certain tools.

magma/backend/mlir/hardware_module.py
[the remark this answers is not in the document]
leonardt: On one hand, OCAML does not care about the order of a record type (https://cs3110.github.io/textbook/chapters/data/records_tuples.html#:~:text=The%20order%20of%20the%20fi%3Ati%20inside%20a%20record%20type,could%20be%20used%20without%20definition.)

On the other hand, C does (since it allows you to control the layout).

I think 2. would actually provide better compatibility since existing code would still work (but the ordering in the backend might be different).  The tests here seem to suggest that's not a major issue for our internal test cases at least (we'd just be changing "when" the sorting happens to construction time rather than in the backend).  I think this option would mirror ocaml (there's no difference between orderings, and they are "normalized" via sorting the keys).

1. would provide more explicit control over layout (e.g. if we want fields to be flattened into specific "index"), which could be useful for performance control.

A middle ground might be to provide 1. as a low-level interface and define a more abstract type system (2.) on top of 1.

tests/test_when.py
rsetaluri: ok, let's make it into a structured environment variable then

PR #1270, opened 2023-05-10
[WIP][MLIR] Use CIRCT python API for Verilog generation

rsetaluri:


rsetaluri:
See for #1275 for non-draft

PR #1271, opened 2023-05-15
[MLIR] Improve expected version error message

leonardt:
Report the expected git hash for CIRCT version


leonardt:
Deprecated by CIRCT API

PR #1272, opened 2023-05-15
[CI] Add cached CIRCT

leonardt:
Also adds cached CIRCT compilation to the github workflow.


rsetaluri:
All looks good. Being overly picky here, but can we separate out the change to add CIRCT cache vs. migrating the examples into separate PRs? It will make it easier to track, especially when we want to revert the caching once we build the real python dependency.

rsetaluri (approved):

leonardt:
Note it looks like caches are restricted to resorting from the current branch or master: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache so once a cached version of master is built we'll avoid having to rebuild for each branch.

PR #1274, opened 2023-05-23
[Examples] Add riscv_mini

leonardt:


examples/riscv_mini/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
```
rsetaluri, line 1: do conftest.py's compose hierarchically? i.e. how does this file interact with top-level conftest.py?

rsetaluri (approved):
Looks good. A few concerns just to keep an eye on:

1. Repo size/download speed. It might be good just to use this PR as a check-in to profile these two metrics. Don't think we need to do this regularly, but would be a good time to find this out now.
2. Test/CI speed. Good idea to have riscv_mini tests be separate. Wondering if we should move the riscv_mini tests to a separate action (rather than step of one job), and/or change the trigger conditions. We could make it more of a weekly integration test, rather than run on every PR/commit.

examples/riscv_mini/conftest.py
leonardt: Yes, a local conftest will run only for the tests in a sub-directory: https://docs.pytest.org/en/6.2.x/writing_plugins.html#conftest-py-local-per-directory-plugins (so rather than thinking about it interacting with the top-level conftest, it may be more useful to think of it as "scoping" configuration for sub-directories).

leonardt:
* `circt-ci` - `290.312` kb
```
~/repos/magma circt-ci
base ❯ du -s .
290312	.
```
* `riscv-mini` - `292.688` kb
```
~/repos/magma riscv-mini
base ❯ du -s .
292688	.
```

So a nominal 0.82% increase in disk usage.  There's other things to consider such as the structure of the git information itself.

```
~/repos/magma riscv-mini
base ❯ git-sizer --verbose
Processing blobs: 15099
Processing trees: 19742
Processing commits: 6702
Matching commits to trees: 6702
Processing annotated tags: 29
Processing references: 604
| Name                         | Value     | Level of concern               |
| ---------------------------- | --------- | ------------------------------ |
| Overall repository size      |           |                                |
| * Commits                    |           |                                |
|   * Count                    |  6.70 k   |                                |
|   * Total size               |  2.22 MiB |                                |
| * Trees                      |           |                                |
|   * Count                    |  19.7 k   |                                |
|   * Total size               |  22.1 MiB |                                |
|   * Total tree entries       |   574 k   |                                |
| * Blobs                      |           |                                |
|   * Count                    |  15.1 k   |                                |
|   * Total size               |   149 MiB |                                |
| * Annotated tags             |           |                                |
|   * Count                    |    29     |                                |
| * References                 |           |                                |
|   * Count                    |   604     |                                |
|     * Branches               |   135     |                                |
|     * Tags                   |   228     |                                |
|     * Remote-tracking refs   |   240     |                                |
|     * Git stash              |     1     |                                |
|                              |           |                                |
| Biggest objects              |           |                                |
| * Commits                    |           |                                |
|   * Maximum size         [1] |  2.86 KiB |                                |
|   * Maximum parents      [2] |     2     |                                |
| * Trees                      |           |                                |
|   * Maximum entries      [3] |   208     |                                |
| * Blobs                      |           |                                |
|   * Maximum size         [4] |  14.2 MiB | *                              |
|                              |           |                                |
| History structure            |           |                                |
| * Maximum history depth      |  3.69 k   |                                |
| * Maximum tag depth      [5] |     1     |                                |
|                              |           |                                |
| Biggest checkouts            |           |                                |
| * Number of directories  [6] |   100     |                                |
| * Maximum path depth     [7] |     6     |                                |
| * Maximum path length    [8] |   127 B   | *                              |
| * Number of files        [6] |  1.42 k   |                                |
| * Total size of files    [9] |  36.1 MiB |                                |
| * Number of symlinks    [10] |     1     |                                |
| * Number of submodules  [11] |     2     |                                |

[1]  46efb7e36f4d82749c32ad0e0ff10968412f868d
[2]  412f84a2271eb907d935595123eb3e009585a02f
[3]  93578d1f8ae609dc2ac0eebe16096b8a383815c2 (refs/remotes/origin/when-formal:tests/test_backend/test_mlir/golds)
[4]  ba83b553b940312bb156172ad6591bcb53e946f1 (8c6cbe5201fde1d1741d21fd0dd630675efa44b0:bin/rhea-hls)
[5]  82cea98f7c220345f90692ce7d754405464ea1fb (refs/tags/v0.1.20)
[6]  dad655c50805932d25115bd1d7ecca410a3f1841 (7479d5a371084f7f55bd9b2198d0c57457cca323^{tree})
[7]  9352a649a9f669f60103a23865a44a1d6075ab74 (57a13018a8a29c0a5cc9b44d1be3fefc079542e9^{tree})
[8]  13f8670911d6f7c360bb0bdacd2805daaf4dd2d6 (refs/heads/circt-ci^{tree})
[9]  2560d5f4b30fb5665d46adafe0fcba6d9f17bf8a (8c6cbe5201fde1d1741d21fd0dd630675efa44b0^{tree})
[10] ade6608d4f3283de668641d08177ba5d4c24f89a (refs/heads/circt-ci:docs)
[11] 2f3dda3e20a23a14ba6f33f36858a02497d96d18 (91ee03ab855afb46c0e0742f732f9e7423ac5aa1:magma/backend/mlir)
```

Doesn't seem to look like anything too problematic, [they mention](https://github.com/github/git-sizer) keeping things under 1GB, which it seems like we are well under.

```
base ❯ git gc
git count-objects -vH
Enumerating objects: 41718, done.
Counting objects: 100% (41718/41718), done.
Delta compression using up to 16 threads
Compressing objects: 100% (12875/12875), done.
Writing objects: 100% (41718/41718), done.
Total 41718 (delta 31939), reused 37564 (delta 28464), pack-reused 0
count: 96
size: 428.00 KiB
in-pack: 41718
packs: 1
size-pack: 22.37 MiB
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```

leonardt:
The riscv-mini tests take `2m 43s` which doesn't seem like a major barrier to action speed.  One option might be to run them only on master and PRs to master, but I don't think they're so slow that it's worth delaying when we're notified that it's broken (e.g. if we merge a couple PRs during the week, we'll have to determine which one broke the weekly check).  Locally it's simple enough for the user to skip running the tests by just point pytest to the `tests` directory, so I think having it on branch builds would still be useful feedback (useful to be aware that they've broken something before even submitting the PR).

rsetaluri:
@leonardt thanks for the analysis!

PR #1275, opened 2023-05-25
[MLIR] Use python API

rsetaluri:


rsetaluri:
@leonardt note that we still need CIRCT to make a release with macosx support. Hopefully that will happens soon :)

leonardt (approved):

PR #1278, opened 2023-06-15
[MLIR] Add locations to HW module signatures

rsetaluri:
Due to https://github.com/llvm/circt/pull/4540, module ports now are also given a location. We need to explicitly set their location (as we do for all op's) so that they don't print `<stdin>...`.


leonardt (approved):

tests/test_backend/test_mlir/golds/simple_structural_location_info_style_wrapInAtSquareBracket.mlir
```diff
@@ -1,5 +1,5 @@
 module attributes {circt.loweringOptions = "locationInfoStyle=wrapInAtSquareBracket"} {
-    hw.module @simple_structural(%a: i16, %b: i16, %c: i16, %CLK: i1) -> (y: i16, z: i16) {
+    hw.module @simple_structural(%a: i16 loc(unknown), %b: i16 loc(unknown), %c: i16 loc(unknown), %CLK: i1 loc(unknown)) -> (y: i16 loc(unknown), z: i16 loc(unknown)) {
```
leonardt, line 2: Should we include a test that has a known location?

rsetaluri: we currently don't ever emit valid locations for ports. we need to add that feature; noted in the PR

rsetaluri:
It should be noted that in this PR, we only add the plumbing for module ports to emit a location -- we don't actually emit valid locations, just the default `loc(unknown)` for now. This brings us to parity before the https://github.com/llvm/circt/pull/4540 change. Soon, we should actually emit real locations of port declarations to enhance the annotations.

PR #1279, opened 2023-06-22
[When] Revert hotfix handling of wiring directly to when output (#1237)

leonardt:
This change duplicated when output targets when a user directly wires the output of a when primitive.  This avoided an inconsistency that occured when a when output was directly wired by the user and then resolved after the direct wiring.  Since the reference to the direct wiring wasn't stored, the direct wiring wasn't updated to match the resolved value.  This was a problem when resolving values elaborated the children as separate primitive outputs (rewiring the conditionally driven value to new outputs).

Now that we are elaborating when outputs using the children of the output (instead of creating new outputs for each child) we no longer have the issue of needing to track additional values driven by a when output (the standard resolve logic now handles this since we maintain the origin when output as the driver).

This simplifies the generated code by avoiding duplication in the generated verilog and reusing the same target variable (although it does introduce some extracts where in previous cases it was a bit-blasted assignment inside the always block)

### DIFF
--- a/magma/wire_container.py
+++ b/magma/wire_container.py
@@ -217,7 +217,6 @@ class Wireable:
         self._enclosing_when_context = get_curr_when_block()
         self._wired_when_contexts = []
         self._parent = None
-        self.is_when_port = False
 
     @property
     def parent(self):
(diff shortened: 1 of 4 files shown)


rsetaluri (approved):
Looks good, we should just ensure that this PR goes into a release by itself

PR #1280, opened 2023-06-27
[When] Add support for emitting when assertions for translation validation

leonardt:
For each conditional wire in a when context, we emit an assert that states if the condition is true, the drivee must equal the driver.

Note: the inline verilog code path requires recursion for tuples (but not arrays).  Once we support flatten_all_tuples=False, we can remove this requirement which would allow better validation of tuple elaboration logic in code generation.

I ended up enabling the assertion generation for all the when tests to ensure that all our existing test cases were covered in assertion testing, which seemed like a better idea than effectively duplicating all the tests with assertions enabled.


rsetaluri (commented):
My main feedback here is that it would be much preferred to refactor this into an orthogonal, self-contained pass rather than an option that touches many parts of the code. It seems possible to insert these inline_verilog statements as a pass? and then wire the appropriate registers into them? we could even add wire objects so that even the MLIR compilation doesn't need to know of them, it just takes in a wire.

leonardt:
Refactored the logic to be in a separate pass.

The wire naming is a bit more subtle.  The issue is that the MLIR value assigned inside the sv.if (which is created inside the when compiler) needs to be named so that MLIR doesn't optimize names out.

Here's the issue, suppose we have a simple assignment
```
if (x)
    _GEN_0 = y;
else
    _GEN_0 = z;
```
now immediately after we might use that value
```
```
_GEN_1 = _GEN_0
if (a)
    _GEN_1 = ~_GEN_0
```

MLIR will try to simplify that code to reduce names, and in this case it could "back propogate" the `_GEN_1` to avoid introducing extra _GEN variables, so we'd get something like:

```
if (x)
    _GEN_1 = y;
else
    _GEN_1 = z;
if (a)
    _GEN_1 = ~_GEN_0
```

Now it's not sufficent to simply insert a `_WHEN_WIRE` that references `_GEN_0`, because then the generated verilog would just be updated to read `_GEN_1` by the MLIR logic (still structurally correct).  However, this breaks the assertion logic, because we generate an assertion based on the first if/else logic, which fails because the second if statement may change the `_WHEN_WIRE` value when `a` is True (even though this should have no bearing on the intermediate value).

We need to insert a name for the `if` statement target to avoid MLIR doing this name simplification.  I'm not sure if there's a way to do this without adding the logic into the backend code.

File: magma/backend/mlir/when_utils.py [hunk 1]
```diff
@@ -13,6 +14,7 @@
 from magma.primitives.when import iswhen
 from magma.ref import ArrayRef, TupleRef, DerivedRef
 from magma.value_utils import make_selector
+from magma.config import config
```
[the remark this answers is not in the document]
leonardt: See my response above re: code path, I'm open to an alternative that addresses the issue but this is only the way that I could come up with.

leonardt:
I'll try it out

rsetaluri:
> I'll try it out

@leonardt any luck?

leonardt:
Resolved the issue with the temporary wire logic, ready for final review now.

magma/when.py
```diff
@@ -1,14 +1,20 @@
 import abc
+from collections import defaultdict
 import contextlib
 import dataclasses
 import functools
 import itertools
+import operator
 from typing import Any, Iterable, Optional, Set, Tuple, Union
 
+from magma.config import config, RuntimeConfig
```
rsetaluri, line 10: per the above comment, we can remove dependency on `config`

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
     return latches
 
 
+# Use a global counter so assertion names are unique across the design
+# (not just per module)
+_WHEN_ASSERT_COUNTER = itertools.count()
```
rsetaluri, line 551: we should just use `assert` or `when_assert` rather than mix the two. locally to this file can just be `assert`

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+                    " ({drivee} == {driver}));")
+
+
+def _get_builder_port(value, builder):
+    """
+    Get reference to corresponding builder output port.  With the bulk
+    reconstruction logic, this may refer to a child of a builder output.
+    """
+    port = builder._check_existing_derived_ref(
+        value, builder._output_to_name, builder._output_to_index)
```
rsetaluri, line 562:
```suggestion
    port = builder._check_existing_derived_ref(
        value, builder._output_to_name, builder._output_to_index
    )
```

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+    if port is None:
+        port = getattr(builder, builder._output_to_name[value])
+    return port
+
+
+def _emit_when_assert(cond, drivee, driver, builder, assignment_map):
+    if _contains_tuple(type(drivee)):
+        # Since tuples are elaborated in verilog, we emit an assert for the
+        # leaf values.
+        # TODO(leonardt): update this when we remove flatten_all_tuples
```
rsetaluri, line 572:
```suggestion
        # TODO(leonardt): Update this when we remove flatten_all_tuples.
```

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+    if not port.wired():
+        return  # port was discarded, do not need to emit assert
+
+    # Track the conditions a value is assigned in for the default driver logic.
+    assignment_map[port].append(cond)
+
+    port = builder.get_when_assert_wire(port)
+
+    _inline_verilog2(_ASSERT_TEMPLATE, cond=cond, drivee=port, driver=driver,
+                     id=next(_WHEN_ASSERT_COUNTER))
```
rsetaluri, line 587:
```suggestion
    # Track the conditions a value is assigned in for the default driver logic.
    assignment_map[port].append(cond)
    port = builder.get_when_assert_wire(port)
    _inline_verilog2(
        _ASSERT_TEMPLATE,
        cond=cond,
        drivee=port,
        driver=driver,
        id=next(_WHEN_ASSERT_COUNTER),
    )
```

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+def _make_cond(block, precond):
+    """Prepends the precond to the block condition if it exists."""
+    if precond is not None:
+        cond = precond
+        if block.condition is not None:
+            cond &= block.condition
+    else:
+        assert block.condition is not None
+        cond = block.condition
+    return cond
```
rsetaluri, line 599:
```suggestion
    if precond is not None:
        if block.condition is not None:
            return precond & block.condition
        return precond
    assert block.condition is not None
    return block.condition
```

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+
+
+def _make_else_cond(block, precond):
+    """Invert the current condition, if there's a precond, append it."""
+    else_cond = ~block.condition
+    if precond is not None:
+        else_cond = precond & else_cond
+    return else_cond
```
rsetaluri, line 607:
```suggestion
    if precond is not None:
        return precond & ~block.condition
    return ~block.condition
```

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+
+def _emit_default_driver_asserts(block, builder, assignment_map):
+    for wire in list(block.root.default_drivers()):
+        port = _get_builder_port(wire.drivee, builder)
+        assigned_conds = assignment_map[port]
+        if not assigned_conds:
+            continue
+        cond = ~functools.reduce(operator.or_, assigned_conds)
+        _emit_when_assert(cond, wire.drivee, wire.driver, builder,
+                          defaultdict(list))
```
rsetaluri, line 618:
```suggestion
        _emit_when_assert(
            cond, wire.drivee, wire.driver, builder, defaultdict(list)
        )
```

magma/when.py
```diff
@@ -526,6 +546,108 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+        # Children inherit this block's cond as a precond.
+        emit_when_asserts(child, builder, cond, assignment_map)
+
+    else_block = _get_else_block(block)
+    if else_block:
+        else_cond = _make_else_cond(block, precond)
+        emit_when_asserts(else_block, builder, else_cond, assignment_map)
+
+    if block is block.root:
+        _emit_default_driver_asserts(block, builder, assignment_map)
```
rsetaluri, line 648:
```suggestion
    cond = _make_cond(block, precond)
    for conditional_wire in block.conditional_wires():
        _emit_when_assert(
            cond, conditional_wire.drivee, conditional_wire.driver, builder, assignment_map
        )
    for child in block.children():
        # Children inherit this block's cond as a precond.
        emit_when_asserts(child, builder, cond, assignment_map)
    else_block = _get_else_block(block)
    if else_block:
        else_cond = _make_else_cond(block, precond)
        emit_when_asserts(else_block, builder, else_cond, assignment_map)
    if block is block.root:
        _emit_default_driver_asserts(block, builder, assignment_map)
```

rsetaluri (changes requested):

magma/backend/mlir/hardware_module.py
```diff
@@ -1349,3 +1349,6 @@ def _add_module_parameters(self, hw_module: hw.ModuleOpBase):
             type = python_type_to_mlir_type(type)
             param = hw.ParamDeclAttr(name, type)
             hw_module.parameters.append(param)
+
+    def gen_scoped_name(self, name: Optional[str] = None, force: bool = False):
```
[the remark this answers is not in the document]
leonardt: Actually with the pass logic, we don't need this API for now (was used in the when_utils compiler)

magma/primitives/when.py
```diff
@@ -121,6 +123,15 @@ def _add_default_drivers_to_register_inputs(
         latches.remove(drivee)
 
 
+def _rewire_driven_value_or_values(value, temp):
```
rsetaluri, line 126: `value` -> `value_or_values`

magma/when.py
```diff
@@ -526,6 +542,109 @@ def find_inferred_latches(block: _BlockBase) -> Set:
...
+    For each drivee in a conditional wire, track the conditions where it is
+    assigned using assignment_map.  This is used for the default driver logic
+    (emit an assert for the case when none of the relevant conditions are
+    true).
+    """
+    cond = _make_cond(block, precond)
+    for conditional_wire in block.conditional_wires():
+        _emit_when_assert(
+            cond, conditional_wire.drivee, conditional_wire.driver, builder,
+            assignment_map
```
rsetaluri, line 634:
```suggestion
            cond,
            conditional_wire.drivee,
            conditional_wire.driver,
            builder,
            assignment_map
```

tests/gold/test_when_2d_array_assign.mlir
```diff
@@ -15,6 +15,23 @@ module attributes {circt.loweringOptions = "locationInfoStyle=none"} {
         %3 = hw.array_get %I[%2] : !hw.array<2xi2>, i1
         %4 = hw.constant 1 : i1
         %5 = hw.array_get %I[%4] : !hw.array<2xi2>, i1
-        hw.output %0 : !hw.array<2xi2>
+        %8 = hw.array_get %0[%2] : !hw.array<2xi2>, i1
```
rsetaluri, line 18: is the `emit_when_asserts` option on by default? I don't see where tests are turning it on

leonardt: It is an argument to `m.compile`, e.g. https://github.com/phanrahan/magma/pull/1280/files#diff-b516e2f8f40209812d4f740191a456abcfc72d567ce20bd41720437cb92b793dR40

magma/passes/when.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class FinalizeWhens(WhenPass):
+    def process_when_builder(self, builder, defn):
+        assert not builder._finalized
+        defn._context_._placer.place(builder.finalize())
+
+
+infer_latches = pass_lambda(InferLatches)
+do_emit_when_asserts = pass_lambda(EmitWhenAsserts)
```
rsetaluri, line 30: i think we should name these as `xxx_pass`. then we can get rid of `do_emit_...` (assuming the do_ is to avoid confusion with the emit_when_asserts bool option)

tests/gold/test_when_2d_array_assign.mlir
rsetaluri: got it...but where in this PR do we pass it to `m.compile()` as `True`? just not seeing it here

rsetaluri: oops my bad, i didn't realize `tests/test_when.py` was folded so ctrl-f didn't see it. got it :)

magma/primitives/when.py
```diff
@@ -121,6 +123,15 @@ def _add_default_drivers_to_register_inputs(
         latches.remove(drivee)
 
 
+def _rewire_driven_value_or_values(value_or_values, temp):
+    if isinstance(value_or_values, list):
```
rsetaluri, line 127: temp -> new_value_or_values

leonardt: Since temp is an instance of `type(value)`, we can just call it `new_value` (shouldn't be a list)

tests/test_backend/test_mlir/golds/inline_verilog_output.mlir
```diff
@@ -0,0 +1,15 @@
+module attributes {circt.loweringOptions = "locationInfoStyle=none"} {
```
rsetaluri, line 1: these files look spurrious? don't see a new test added

rsetaluri (approved):
Looks good. Last request ( :) ). Can we rename the option and action to `emit_when_assertions`? The emitted verilog can still be `WHEN_ASSERT`, but just in the code.

Also, let's try to rebase master rather than merge, when we merge the PR.

PR #1289, opened 2023-07-20
[MLIR] Fix array of bit constant creation

### LINKED ISSUE #1288: [Primitives] `Register(m.Array[N, m.Bit])` doesn't compile

rsetaluri:
Fixes #1288.


magma/array.py
```diff
@@ -636,6 +636,14 @@ def const(self):
                        for _, child in self._enumerate_children())
         return False
 
+    def __int__(self):
+        if not self.const():
+            raise TypeError("Can't call __int__ on a non-constant")
+        if not issubclass(self.T, Digital):
+            raise TypeError("Can only call __int__ on array of bit values")
+        bits = [int(v) for _, v in self._enumerate_children()]
```
leonardt, line 644: Perhaps it's safer to use `bool` instead of `int` here since this assumes the children are "bit-like" (evaluate to True or False), which in turn would require a `__bool__` method?  Otherwise a child type that evaluates to an integer could work here, but probably would error downstream in the BitVector logic (unless it forces the bit values to true or false).  This might catch the error earlier.

leonardt (approved):

Issue #1291, opened 2023-07-26
[CompileGuard] Compile guard select issues


rsetaluri:
- [ ] Support only 1 key
- [ ] Support in MLIR backend

PR #1292, opened 2023-07-26
[CompileGuard] Fix compile guard select key parsing

rsetaluri:
See #1291. Fixes two-key case (previously unsupported), and adds and refactors tests.


rsetaluri:
Closing in favor of #1294 which should encapsulate all fixes.

PR #1294, opened 2023-07-27
[CompileGuard Select] Fix issues

rsetaluri:


magma/compile_guard.py
```diff
@@ -140,44 +140,44 @@ def _is_simple_type(T: Kind) -> bool:
 
 class _CompileGuardSelect(Generator2):
     def __init__(self, T: Kind, keys: Tuple[str]):
-        # NOTE(rsetaluri): We need to add this check because the implementation
+        assert keys  # not empty
```
leonardt, line 143:
```suggestion
        assert keys, "Expected not empty keys"
```

tests/test_compile_guard.py
```diff
@@ -253,20 +254,50 @@ class _Top(m.Circuit):
...
+    class _Top(m.Circuit):
+        io = m.IO(I=m.In(m.Bit), O=m.Out(m.Bit)) + m.ClockIO()
+        x = m.Register(m.Bit)()(io.I ^ 1)
+        io.O @= m.compile_guard_select(COND1=x, default=io.I)
+
+    basename = "test_compile_guard_select_two_keys"
+    m.compile(f"build/{basename}", _Top, inline=True)
+    assert m.testing.check_files_equal(
+        __file__, f"build/{basename}.v", f"gold/{basename}.v"
+    )
```
leonardt, line 282: Should we update these to use mlir/mlir-verilog instead of coreir? Will save us doing it in the future.

leonardt (changes requested):

tests/test_compile_guard.py
rsetaluri: Will do these in a follow up PR.

PR #1296, opened 2023-07-31
[Examples] Add reset for all synchronous components in riscv_mini

leonardt:
Needed for verilog simulators that initialize to `X` rather than 0


examples/riscv_mini/tests/test_imm_gen.py
```diff
@@ -60,6 +60,9 @@ class DUT(m.Circuit):
                   counter.O, imm.sel, imm.O, O)
 
     tester = f.Tester(DUT, DUT.CLK)
+    tester.poke(DUT.RESET, 1)
```
rsetaluri, line 63: should we use a `ResetTester`?

rsetaluri (approved):

PR #1299, opened 2023-08-08
[MLIR] Make minor fixes and improvements

rsetaluri:
Some notes:
* Adds a tool to convert an MLIR file to Verilog with the same options we have in the top down flow. Useful for debugging (e.g. hand-modify .mlir file, recompile to Verilog).
* Always checks Verilog compilation for backend tests by default. Previously, this was turned off since (a) CI didn't always have `circt` available; (b) `circt-opt` wasn't standardized; and (c) running `circt-opt` was slow. Now all of these issues are addressed with v2.3.0.


magma/backend/mlir/mlir_to_verilog.py
```diff
@@ -33,6 +33,6 @@ def mlir_to_verilog(
         )
         pm.run(module.operation)
     if opts.split_verilog:
-        circt.export_split_verilog(module, "")
+        circt.export_split_verilog(module, ".")
```
leonardt, line 36: Should add a test to cover this case

tools/mlir_to_verilog_main.py
```diff
@@ -0,0 +1,78 @@
...
+            f"enabled"
+        )
+
+    with open(args.infile, "r") as f_in:
+        mlir_to_verilog(f_in, args.outfile, opts)
+
+
+if __name__ == "__main__":
+    exit(main())
+
```
leonardt, line 78: Let's add a test to cover this script

leonardt: Is it worth adding this script to the package (like https://github.com/phanrahan/magma/blob/master/setup.py#L15) so that it's easier to use? That way we wouldn't have to manually add it to the path or find it relative to current location.  We could prefix with `magma_mlir_to_verilog` or something so it doesn't conflict in the path.  Also, perhaps there's a way to only install it in dev mode unless we think the user might find it useful.  I know there's `python setup.py develop`, not sure if there's a pip equivalent.

tools/mlir_to_verilog_main.py
```diff
@@ -0,0 +1,78 @@
...
+    args = parser.parse_args(prog_args)
+    opts = slice_opts(vars(args), MlirToVerilogOpts)
+    
+    logging.debug(f"Running with opts: {opts}")
+    if opts.split_verilog and args.outfile is not sys.stdout:
+        logging.warning(
+            f"outfile ({args.outfile.name}) likely ignored with split_verilog "
+            f"enabled"
+        )
```
leonardt, line 70: Is there actually a case where it isn't ignored? If so we should check and not issue a warning so we don't issue spurious warnings that are ignored.  Otherwise we should make the language stronger, e.g. "will be ignored"

tools/mlir_to_verilog_main.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def _field_to_argument_params(field: dataclasses.Field) -> Dict:
+    if field.default_factory is not dataclasses.MISSING:
+        raise TypeError(field)
+    params = {
+        "required": field.default is dataclasses.MISSING,
+    }
+    if field.type is bool and not params["required"] and not field.default:
+        params.update({"action": "store_true"})
```
leonardt, line 22:
```suggestion
        params["action"] = "store_true"
```

tools/mlir_to_verilog_main.py
```diff
@@ -0,0 +1,78 @@
...
+        return params
+    if not params["required"]:
+        params.update({"default": field.default})
```
leonardt, line 25:
```suggestion
        params["default] = field.default
```

tools/mlir_to_verilog_main.py
leonardt: Could just use key assignment for these?

https://stackoverflow.com/questions/15456158/python-dict-update-vs-subscript-to-add-a-single-key-value-pair

tools/mlir_to_verilog_main.py
```diff
@@ -0,0 +1,78 @@
...
+    params.update({
+        "action": "store",
+        "type": field.type,
+    })
```
leonardt, line 29:
```suggestion
    params["action"] = "store"
    params["type"] = field.type
```

leonardt (changes requested):

tools/mlir_to_verilog_main.py
rsetaluri: good point, it is always ignored. the output file always comes from the annotations in the .mlir file (or are automatically determined from module names)

tools/mlir_to_verilog_main.py
rsetaluri: That's a good idea. So I've intended everything in `tools/` to be run as `python tools/{tool}.py`. I'm ok leaving that as is since it's a developer only thing (develop setup mode nonetheless noted) and it's just the filepath and they're all in tools.

magma/backend/mlir/mlir_to_verilog.py
rsetaluri: done. also added an option to pass in the directory since that might be useful

leonardt (approved):

PR #1301, opened 2023-08-08
[Mantle] Init ice40 primitives

leonardt:
Provides an example of latest magma style for @phanrahan who will start extending this platform for the new mantle library.

Adds a new API `get_param_types` for circuits that allows us to begin to deprecate the `coreir_config_param_types` API with a simpler `param_types` API


phanrahan:
Thanks. I will review ASAP.

I have also created two public repos mantle2 and loam2 so I can extend and test the parts not in magma.

phanrahan:
I tried running the test test_ice40 .  It fails because coreir is not installed. Can we rebqse on the depreciate coreir branch?

leonardt:
Rebased on deprecate-coreir

leonardt:
Just pushed another rebase, I needed to first rebase `deprecate-coreir` on `master`, then I rebased this branch again onto `deprecate-coreir` again to resolve some issues in the diff

phanrahan:
It now complains about no module circt.

leonardt:
I would try a `pip install -e .` to reinstall magma, we added a python wheel for circt so we no longer need users to separately compile their own binary (their should be an m-series mac version available)

phanrahan:
I tried that

ERROR: Could not find a version that satisfies the requirement circt==1.48.1.dev34 (from magma-lang) (from versions: none)
ERROR: No matching distribution found for circt==1.48.1.dev34

leonardt:
@rsetaluri I was looking at https://github.com/llvm/circt/blob/main/.github/workflows/uploadWheels.yml#L25-L30, did we not add support for m-series?

Issue #1269, opened 2023-05-09
[MLIR] Bug in always comb ordering


leonardt:
```python
class test_when_alwcomb_order_2(m.Circuit):
    io = m.IO(I=m.In(m.Bits[8]), S=m.In(m.Bits[2]), O=m.Out(m.Bits[8]))

    a = m.Register(m.Bits[8])(name="a")
    b = m.Register(m.Bits[8])(name="b")
    b.I @= a.O

    x = m.Bits[8]()
    x @= a.O

    with m.when(io.S[0]):
        x @= io.I
        b.I @= io.I

    with m.when(io.S[1]):
        a.I @= 0
    with m.otherwise():
        a.I @= x

    io.O @= b

m.compile(
    "build/test_when_alwcomb_order_2",
    test_when_alwcomb_order_2,
    output="mlir-verilog"
)
```

produces

```verilog
// Generated by CIRCT circtorg-0.0.0-1773-g7abbc4313
module test_when_alwcomb_order_2(
  input  [7:0] I,
  input  [1:0] S,
  input        CLK,
  output [7:0] O
);

  reg [7:0] _GEN;
  reg [7:0] a;
  reg [7:0] _GEN_0;
  reg [7:0] _GEN_1;
  always_comb begin
    if (S[1])
      _GEN = 8'h0;
    else
      _GEN = _GEN_0;
    _GEN_0 = a;
    _GEN_1 = a;
    if (S[0]) begin
      _GEN_0 = I;
      _GEN_1 = I;
    end
  end // always_comb
  reg [7:0] b;
  always_ff @(posedge CLK) begin
    a <= _GEN;
    b <= _GEN_1;
  end // always_ff @(posedge)
  initial begin
    a = 8'h0;
    b = 8'h0;
  end // initial
  assign O = b;
endmodule

```

Notice that `_GEN_0` (`x` in the source python) is referenced before it is assigned later in the always comb block.  The issue has to do with the ordering of module compilation, i.e. the dependencies between the when primitive instances isn't not reflected in the order of the module compilation.

MLIR for reference
```
module attributes {circt.loweringOptions = "locationInfoStyle=none"} {
    hw.module @test_when_alwcomb_order_2(%I: i8, %S: i2, %CLK: i1) -> (O: i8) {
        %0 = comb.extract %S from 0 : (i2) -> i1
        %1 = comb.extract %S from 1 : (i2) -> i1
        %2 = hw.constant 0 : i8
[...]

PR #1309, opened 2023-08-17
[When] Split when blocks with cycles

leonardt:

Note the generated verilog can have empty blocks since some of the assignments are being emitted in a split block, but this seems to be okay for verilator.  Since this is likely to be an uncommon case, I don't think we should worry too much about trying to cleanup the generated verilog.


tests/test_when.py
```diff
@@ -1835,12 +1815,136 @@ class test_when_alwcomb_order(m.Circuit):
...
+    if check_gold(__file__, "test_when_alwcomb_order.v"):
+        return
+    verilator_path = os.path.join(
+        os.path.dirname(__file__),
+        "build",
+        "test_when_alwcomb_order.v"
+    )
+    assert not os.system(
+        f"verilator --lint-only {verilator_path}"
+    )
```
rsetaluri, line 1834:
```suggestion
    assert not os.system(f"verilator --lint-only {verilator_path}")
```

tests/test_when.py
```diff
@@ -1835,12 +1815,136 @@ class test_when_alwcomb_order(m.Circuit):
...
+    if check_gold(__file__, "test_when_alwcomb_order_complex.v"):
+        return
+    verilator_path = os.path.join(
+        os.path.dirname(__file__),
+        "build",
+        "test_when_alwcomb_order_complex.v"
+    )
+    assert not os.system(
+        f"verilator --lint-only {verilator_path}"
+    )
```
rsetaluri, line 1871:
```suggestion
    assert not os.system(f"verilator --lint-only {verilator_path}")
```

tests/test_when.py
```diff
@@ -1835,12 +1815,136 @@ class test_when_alwcomb_order(m.Circuit):
...
+    if check_gold(__file__, "test_when_alwcomb_order_nested_2.v"):
+        return
+    verilator_path = os.path.join(
+        os.path.dirname(__file__),
+        "build",
+        "test_when_alwcomb_order_nested_2.v"
+    )
+    assert not os.system(
+        f"verilator --lint-only {verilator_path}"
+    )
```
rsetaluri, line 1946:
```suggestion
    assert not os.system(f"verilator --lint-only {verilator_path}")
```

tests/test_when.py
```diff
@@ -1835,12 +1815,136 @@ class test_when_alwcomb_order(m.Circuit):
             x @= ~io.I
             io.O @= ~x
 
-    with pytest.raises(MlirWhenCycleError):
-        m.compile(
-            "build/test_when_alwcomb_order",
-            test_when_alwcomb_order,
-            output=_OUTPUT_TYPE
-        )
+    m.compile(
```
rsetaluri, line 1818: wondering if all these tests following the same structure could be run using a test harness -- or simpler just a helper function like `run_test()`

magma/when.py
```diff
@@ -680,3 +680,71 @@ def otherwise():
...
+    driver_map: for each context that drivee is driven, store the driver
+    """
+    driver_map = {}
+    for ctx in drivee._wired_when_contexts:
+        wires = ctx.get_conditional_wires_for_drivee(drivee)
+        driver_map[ctx] = (wire.driver for wire in wires)
+    return driver_map
+
+
+def split_when_cycles(builder, defn):
```
rsetaluri, line 742: if these are all free functions, let's move them to the pass area since theyre not fundamental to the basic when construct

rsetaluri (changes requested):

magma/passes/split_when_utils.py
```diff
@@ -0,0 +1,76 @@
...
+    for _elsewhen in curr_block.elsewhen_blocks():
+        _emit_new_when_assign(value, driver_map, _elsewhen)
+    if curr_block.otherwise_block:
+        _emit_new_when_assign(value, driver_map, curr_block.otherwise_block)
+
+
+def _build_driver_map(drivee):
+    """
+    driver_map: for each context that drivee is driven, store the driver
+    """
```
rsetaluri, line 60:
```suggestion
    # driver_map stores the driver, for each context that drivee is driven.
```

not really a docstring

rsetaluri (approved):
Final minor comment

PR #1311, opened 2023-08-24
[CIRCT] Bump to 1.51.1.dev75

leonardt:
* https://github.com/llvm/circt/pull/5880 needed to fix issue with disallowPackedStructs flag


rsetaluri (approved):
We should consider making `omit_version_comment = True` by default so we don't have all these lines to change everytime. We can do it an a separate PR if desired.

PR #1312, opened 2023-08-24
[CoreIR] Make coreir an optional dependency

leonardt:


leonardt:
@rsetaluri this is ready for review now, I updated to be a minimal change that:
* Removes the dependency from the setup.py
* Updates the internals to only import coreir on demand (so importing magma doesn't fail when coreir isn't present)

I started down the road of updating the tests, but realized it might be simpler to just do a simple sanity check for the optional dependency for now (rather than updating the tests to skip coreir if it's not present, I think it's better to migrate or remove tests we don't want anymore in a follow up PR).

PR #1315, opened 2023-08-31
[Bind2] Add support for compile guard around bind

rsetaluri:
Adds support for emitting `ifdef` around `bind` calls. The syntax for this is:

```python
with m.compile_guard(...):
    m.bind2(...)
```

which emits MLIR like:

```MLIR
sv.ifdef ... {
  sv.bind ...
}
```

```Verilog
`ifdef ...
  bind ...
`endif
```

One important thing to note is that in the case of split verilog, the output file attribute annotation has to be applied up the compile guard stack, i.e. the `sv.ifdef` surrounding the `sv.bind` also needs the `hw.OutputFileAttr` attached.

### DIFF
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -45,11 +45,7 @@ from magma.backend.mlir.scoped_name_generator import ScopedNameGenerator
 from magma.backend.mlir.sv import sv
 from magma.backend.mlir.when_utils import WhenCompiler
 from magma.backend.mlir.xmr_utils import get_xmr_paths
-from magma.bind2 import (
-    get_bound_instance_info,
-    is_bound_instance,
-    UnboundInstanceError,
-)
+from magma.bind2 import maybe_get_bound_instance_info, is_bound_instance
 from magma.bit import Bit
 from magma.bits import Bits, BitsMeta
 from magma.bitutils import clog2, clog2safe
@@ -1140,9 +1136,8 @@ class NativeBindProcessor(BindProcessorInterface):
             instance = hw.InnerRefAttr(defn_sym, sym)
             sv.BindOp(instance=instance)
         for instance in self._defn.instances:
-            try:
-                bound_instance_info = get_bound_instance_info(instance)
-            except UnboundInstanceError:
+            bound_instance_info = maybe_get_bound_instance_info(instance)
+            if bound_instance_info is None:
                 continue
             inst_sym = self._ctx.parent.get_mapped_symbol(instance)
             ref = hw.InnerRefAttr(defn_sym, inst_sym)
--- a/magma/backend/mlir/hardware_module.py
+++ b/magma/backend/mlir/hardware_module.py
@@ -45,7 +45,11 @@ from magma.backend.mlir.scoped_name_generator import ScopedNameGenerator
 from magma.backend.mlir.sv import sv
 from magma.backend.mlir.when_utils import WhenCompiler
 from magma.backend.mlir.xmr_utils import get_xmr_paths
-from magma.bind2 import is_bound_instance
+from magma.bind2 import (
+    get_bound_instance_info,
+    is_bound_instance,
+    UnboundInstanceError,
+)
 from magma.bit import Bit
 from magma.bits import Bits, BitsMeta
 from magma.bitutils import clog2, clog2safe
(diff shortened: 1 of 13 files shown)


magma/backend/mlir/hardware_module.py
```diff
@@ -1135,11 +1139,20 @@ def postprocess(self):
...
             instance = hw.InnerRefAttr(defn_sym, sym)
             sv.BindOp(instance=instance)
-        bound_instances = list(filter(is_bound_instance, self._defn.instances))
-        for bound_instance in bound_instances:
-            inst_sym = self._ctx.parent.get_mapped_symbol(bound_instance)
+        for instance in self._defn.instances:
+            try:
+                bound_instance_info = get_bound_instance_info(instance)
+            except UnboundInstanceError:
+                continue
```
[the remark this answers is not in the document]
leonardt: Also, reading this: https://stackoverflow.com/questions/903130/hasattr-vs-try-except-block-to-deal-with-non-existent-attributes

And doing the bench locally for python 3.8
```

~
base ❯ python -mtimeit -s 'class C(object): a = 4
c = C()' 'hasattr(c, "nonexistent")'

5000000 loops, best of 5: 42.9 nsec per loop

~
base ❯ python -mtimeit -s 'class C(object): a = 4
c = C()' 'hasattr(c, "a")'
5000000 loops, best of 5: 44.2 nsec per loop

~
base ❯

~
base ❯ python -mtimeit -s 'class C(object): a = 4
c = C()' 'try:
 c.a
except:
 pass'

10000000 loops, best of 5: 20.4 nsec per loop

~
base ❯ python -mtimeit -s 'class C(object): a = 4
c = C()' 'try:
 c.nonexistent
except:
 pass'
1000000 loops, best of 5: 229 nsec per loop
```
it suggests we might want to use `getattr` with default here instead of try/except because I think being a bound instance is the less common case?  The more general pattern seems to be: if expected to exist, use try/catch, if expected to not exist, use hasattr or getattr

magma/backend/mlir/mlir.py
```diff
@@ -197,6 +197,15 @@ def new_region(self) -> MlirRegion:
...
         self.parent = weakref.ref(parent)
 
+    def parent_op(self) -> Optional['MlirOp']:
+        block = maybe_dereference(self.parent)
+        if block is None:
+            return None
+        region = maybe_dereference(block.parent)
+        if region is None:
+            return None
+        return maybe_dereference(region.parent)
```
leonardt, line 207: Since we're only testing Python 3.8 and above, perhaps we can start using the walrus operator
```suggestion
        if not (block := maybe_dereference(self.parent)):
            return None
        if not (region := maybe_dereference(block.parent)):
            return None
        return maybe_dereference(region.parent)
```

magma/backend/mlir/mlir_passes.py
```diff
@@ -0,0 +1,25 @@
...
+        self._root = root
+        self._callable = callable(self)
+
+    def _run_on_op(self, op: MlirOp):
+        for region in op.regions:
+            for block in region.blocks:
+                for op in block.operations:
+                    yield from self._run_on_op(op)
+                    if self._callable:
+                        yield self(op)
```
leonardt, line 17: Would this be better as an abstact base class with __call__ required to be defined? Otherwise this logic seems not useful (runs the traversal but doesn't do anything with the traversed elements)

magma/compile_guard.py
```diff
@@ -17,6 +19,38 @@
...
+def get_active_compile_guard_info() -> Iterable['CompileGuardInfo']:
+    stack = _get_compile_guard_builder_stack()
+    offset = 0
+    while True:
+        try:
+            builder = stack.peek(offset=offset)
+        except IndexError:
+            break
+        yield builder.info
+        offset += 1
```
leonardt, line 46: Perhaps we could add an `Iterable` interface to the stack and use it here?

leonardt (changes requested):

magma/backend/mlir/hardware_module.py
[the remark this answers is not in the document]
rsetaluri: :+1:

magma/backend/mlir/mlir_passes.py
rsetaluri: :+1:

magma/compile_guard.py
rsetaluri: :+1:

leonardt (approved):

Issue #1316, opened 2023-09-14
[Types] Tuple field wiring check needs to happen recursively


leonardt:
https://github.com/phanrahan/magma/commit/4387b6436d1290ce38214533203a9a3f2c6b21fb is unsufficient to catch the case in which the mismatched field order is contained within a recursive type that is not explicitly wired (e.g. when the parents are bulk wired)

PR #1317, opened 2023-09-14
[Types] Add recursive wiring key ordering check

leonardt:
Fixes https://github.com/phanrahan/magma/issues/1316

Ensures that recursive types check tuple key ordering

### DIFF
--- a/magma/tuple.py
+++ b/magma/tuple.py
@@ -163,7 +163,8 @@ class TupleKind(TupleMeta, Kind):
         if not isinstance(rhs, TupleKind) or len(cls.fields) != len(rhs.fields):
             return False
         for idx, T in enumerate(cls.fields):
-            if not T.is_wireable(rhs[idx]):
+            T = magma_type(T)
+            if not T.is_wireable(magma_type(rhs[idx])):
                 return False
         return True
 
@@ -172,7 +173,8 @@ class TupleKind(TupleMeta, Kind):
         if not isinstance(rhs, TupleKind) or len(cls.fields) != len(rhs.fields):
             return False
         for idx, T in enumerate(cls.fields):
-            if not T.is_bindable(rhs[idx]):
+            T = magma_type(T)
+            if not T.is_bindable(magma_type(rhs[idx])):
                 return False
         return True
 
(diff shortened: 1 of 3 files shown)


magma/tuple.py
```diff
@@ -330,6 +332,18 @@ def wire(self, o, debug_info):
...
+
+        for i, k in enumerate(self.keys()):
+            if not type(self).fields[i].is_wireable(type(o).fields[i]):
+                _logger.error(
+                    WiringLog(f"Cannot wire {{}} (type={type(o)}, to "
+                              f" {{}} (type={type(self)})"
+                              f"because the key {k} is not wireable",
+                              o, self),
+                    debug_info=debug_info
+                )
```
rsetaluri, line 344:
```suggestion
                    WiringLog(
                        f"Cannot wire {{}} (type={type(o)}, to "
                        f" {{}} (type={type(self)})"
                        f"because the key {k} is not wireable",
                        o,
                        self
                    ),
                    debug_info=debug_info
                )
```

magma/tuple.py
```diff
@@ -585,23 +599,25 @@ def is_wireable(cls, rhs):
         rhs = magma_type(rhs)
         if (
             not isinstance(rhs, AnonProductKind) or
-            len(cls.fields) != len(rhs.fields)
+            list(cls.field_dict.keys()) != list(rhs.field_dict.keys())
         ):
             return False
         for k, v in cls.field_dict.items():
-            if k not in rhs.field_dict or not v.is_wireable(rhs.field_dict[k]):
+            v = magma_type(v)
```
rsetaluri, line 606: it seems like we are doubly calling magma_type here and inside is_wireable? the interface to that function should be clear whether it accepts a type or (type | magma_protocol). and in that case just pass in the raw "type" instead of calling magma_type out here.

rsetaluri (changes requested):

magma/tuple.py
```diff
@@ -585,23 +601,24 @@ def is_wireable(cls, rhs):
...
         if (
             not isinstance(rhs, AnonProductKind) or
-            len(cls.fields) != len(rhs.fields)
+            list(cls.field_dict.keys()) != list(rhs.field_dict.keys())
         ):
             return False
         for k, v in cls.field_dict.items():
-            if k not in rhs.field_dict or not v.is_bindable(rhs.field_dict[k]):
+            v = magma_type(v)
+            if not v.is_bindable(magma_type(rhs.field_dict[k])):
```
rsetaluri, line 621: same question re: magma_type here

rsetaluri (approved):
Looks good modulo magma_type comment for is_bindable

PR #1322, opened 2023-09-22
[When] Split when blocks with cycles

rsetaluri:


rsetaluri:
Note the generated verilog can have empty blocks since some of the assignments are being emitted in a split block, but this seems to be okay for verilator. Since this is likely to be an uncommon case, I don't think we should worry too much about trying to cleanup the generated verilog.

rsetaluri:
@leonardt this is essentially a merge with master of #1309.

PR #1324, opened 2023-09-23
[When] Add support for unflattened tuples

rsetaluri:
Allows compiling when blocks to MLIR with `flatten_all_tuples=False`.


rsetaluri:
@leonardt this is basically #1267 rebased on #1323.

rsetaluri:
Allows compiling when blocks to MLIR with `flatten_all_tuples=False`.

PR #1325, opened 2023-09-23
Add bitlen utility

rkshthrmsh:
- bitlen(x) provides number of bits in binary representation of x.

Although `clog2` is provided as a utility function for approximating bit lengths from values, it fails in edge cases where the value itself is a power of 2. For example, `clog2(16) = 4`, but 5 bits are needed to store the value 16. The right formula here would be: `floor(log_2(16)) + 1`. Although most hardware designers might be aware of this, a helper function for bit length of any generic value could be helpful.

### DIFF
--- a/magma/bitutils.py
+++ b/magma/bitutils.py
@@ -34,3 +34,6 @@ def clog2safe(x: int) -> int:
     if x == 1:
         return 1
     return clog2(x)
+
+def bitlen(x: int) -> int:
+    return math.floor(math.log2(x)) + 1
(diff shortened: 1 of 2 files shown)


rsetaluri:
Thanks. Can you add some tests to `tests/test_bitutils.py`? You can follow the pattern of the other tests in that file.

rkshthrmsh:
Added tests.

rsetaluri (approved):

rsetaluri:
@rkshthrmsh let's not include 2608825 in this PR, and do it separately

cdonovick:
Worth noting that `bit_length` is already a method on `int`.

```Python3
assert (16).bit_length() == 5
```

So probably no need to provide this as a utility.

rkshthrmsh:
@cdonovick, thanks for pointing this out. @rsetaluri, I think this PR can be closed in favor of the `bit_length` method on `int`.

PR #1331, opened 2023-09-26
[MLIR] Update MLIR to Verilog test

cdonovick:


cdonovick:
Context: my OS doesn't have the specific version of CIRCT specified by the setup.py so my version string looks different.  Other tests run without issue for me.

PR #1333, opened 2023-10-11
[MLIR] Fix issues with tuple compilation

leonardt:
* Fixes problem with `.value()` where `iswhole` would incorrectly return True when two tuples with different key orderings were wired (it's ok to wire two tuples when there keys don't match if you wire the children up explicitly, but `.value()` shouldn't return the source tuple because that's not a compatible type, instead it should construct a new flipped version).
* Fixes issue with tuple compile keys (need to prepend `_` to integer keys)
* Fixes sorting issue in when struct create (did not revert when other sorting code was reverted)
* Update names/comments in when recursive code (is general across array/tuples, but some code still referred to array)

### DIFF
--- a/magma/tuple.py
+++ b/magma/tuple.py
@@ -405,7 +405,7 @@ class Tuple(Type, Tuple_, AggregateWireable, metaclass=TupleKind):
                 j = keys[i]
             except IndexError:
                 return False
-            # elements should be in correct key order
+            # Elements should be in correct key order
             if ts[i].name.index != j:
                 return False
 
--- a/magma/tuple.py
+++ b/magma/tuple.py
@@ -375,8 +375,7 @@ class Tuple(Type, Tuple_, AggregateWireable, metaclass=TupleKind):
     def wired(self):
         return all(t.wired() for t in self)
 
-    @staticmethod
-    def _iswhole(ts, keys):
+    def _iswhole(self, ts):
 
         for i in range(len(ts)):
             if ts[i].anon():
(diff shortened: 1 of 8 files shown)


magma/tuple.py
```diff
@@ -392,19 +391,28 @@ def _iswhole(ts, keys):
...
+
+        if keys != list(self.keys()):
+            return False
+
+        for i in range(len(ts)):
+            try:
+                j = keys[i]
+            except IndexError:
+                return False
+            # elements should be in correct key order
```
rsetaluri, line 408: nit
```suggestion
            # Elements should be in correct key order.
```

tests/test_type/test_tuple.py
```diff
@@ -408,3 +408,14 @@ class tuple_key_ordering(m.Circuit):
...
+
+def test_tuple_same_key_value():
+    T0 = m.AnonProduct[{"x": m.Bit, "y": m.Bits[8]}]
+    T1 = m.AnonProduct[{"y": m.Bits[8], "x": m.Bit}]
+
+    class tuple_same_key_value(m.Circuit):
+        io = m.IO(I=m.In(T0), O=m.Out(T1))
+        io.O.x @= io.I.x
+        io.O.y @= io.I.y
+        assert isinstance(io.O.value(), m.Out(T1)), type(io.O.value())
```
rsetaluri, line 421: we can probably do all this without an enclosing circuit right?

```suggestion
    I = m.Out(T0)
    O = m.In(T1)
    O.x @= I.x
    O.y @= I.y
    assert isinstance(O.value(), m.Out(T1)), type(O.value())
```

rsetaluri (approved):
few minor comments, otherwise looks good!

PR #1336, opened 2023-10-27
[MLIR] Update module printing syntax

leonardt:
Needed for upstream changes https://github.com/llvm/circt/pull/6205


magma/backend/mlir/hw.py
```diff
@@ -111,25 +110,29 @@ class ModuleOpBase(MlirOp):
...
+            for value in self.results:
+                yield f"out {value.raw_name}: {value.type.emit()}"
+            return
+        for value in self.operands:
+            yield f"in {value.name}: {value.type.emit()} {value.location.emit()}"
+        for value in self.results:
+            yield (
+                f"out {value.raw_name}: {value.type.emit()} "
+                f"{value.location.emit()}"
+            )
```
leonardt, line 126: How about another helper like:

```suggestion
    def _yield_term(self, prefix, value, print_locations):
        assert prefix in {"in", "out"}
        name = value.name if prefix == "in" else value.raw_name
        type = value.type.emit()

        s = f"{prefix} {name}: {type}"

        if print_locations:
            s += f" {value.location.emit()}"

        yield s

    def _get_terms(self, print_locations: bool) -> Iterable[str]:
        for value in self.operands:
            yield from self._yield_term("in", value)
        for value in self.results:
            yield from self._yield_term("out", value)
```

PR #1342, opened 2023-12-05
[v3.0.0] Export v2 objects as core objects

rsetaluri:
Exports Generator2 as Generator, inline_verilog2 as inline_verilog and bind2 as bind.

For Generator and inline_verilog, both versions still exist internally, but this is a breaking change with respect to the external API. For example, m.Generator is now what m.Generator2 used to be, which has a different interface.

For bind, Circuit.bind and Generator.bind no longer exist, and only m.bind remains. However, the internal code backing Circuit.bind still exists.

Additionally, some tests have been deleted or skipped for convenience since they cover non-public API (previously deprecated).

A future change should purge the internal remnants of the old versions. However, such a change will not require a major upgrade, since the external API will not break.


magma/__init__.py
```diff
@@ -78,9 +78,11 @@ def set_mantle_target(t):
 from magma.family import get_family
 _Family_ = get_family()
 
-from .generator import Generator, Generator2, DebugGenerator2
+from .generator import Generator2, DebugGenerator2
+Generator = Generator2
+DebugGenerator = DebugGenerator2
+
```
leonardt, line 84:
```suggestion
from .generator import (
    Generator2 as Generator, 
    DebugGenerator2 as DebugGenerator
)
```

magma/__init__.py
```diff
@@ -132,6 +134,7 @@ def set_mantle_target(t):
 import magma.smart
 from magma.compile_guard import compile_guard, compile_guard_select
 from magma.inline_verilog2 import inline_verilog2
+inline_verilog = inline_verilog2
```
leonardt, line 137:
```suggestion
from magma.inline_verilog2 import inline_verilog2 as inline_verilog
```

magma/__init__.py
```diff
@@ -144,6 +147,7 @@ def set_mantle_target(t):
 from magma.when import when, elsewhen, otherwise
 from magma.value_utils import fill
 from magma.bind2 import bind2, make_bind_ports
+bind = bind2
```
leonardt, line 150:
```suggestion
from magma.bind2 import (
    bind2 as bind, 
    make_bind_ports
)
```

leonardt (approved):
LGTM, minor nits on import (maybe using `as` is better since it forces the renaming to be inline with the import so it doesn't get lost)

rsetaluri:
> LGTM, minor nits on import (maybe using `as` is better since it forces the renaming to be inline with the import so it doesn't get lost)

This is a good point, but actually a lot of internal code still uses the `<name>2` versions. I figure this can be an internal only change after the fact? And have duplicates exposed.

rsetaluri:
> > LGTM, minor nits on import (maybe using `as` is better since it forces the renaming to be inline with the import so it doesn't get lost)
> 
> This is a good point, but actually a lot of internal code still uses the `<name>2` versions. I figure this can be an internal only change after the fact? And have duplicates exposed.

I'll add comments to make this clear.

PR #1350, opened 2024-01-24
[Examples] Add register_file

leonardt:


examples/tests/test_register_file.py
```diff
@@ -0,0 +1,27 @@
+import random
+import pytest
```
rsetaluri, line 2: nit
```suggestion
import pytest
import random
```

examples/tests/test_register_file.py
```diff
@@ -0,0 +1,27 @@
...
+
+from ..register_file import RegisterFile
```
rsetaluri, line 4: any chance there's a better way to do this?

examples/tests/test_register_file.py
```diff
@@ -0,0 +1,27 @@
...
+
+import magma as m
+import fault as f
+from hwtypes import BitVector
```
rsetaluri, line 8: nit
```suggestion
import fault as f
from hwtypes import BitVector
import magma as m
```

rsetaluri (approved):

examples/tests/test_register_file.py
leonardt: I'm going to change the examples format to be in the form a jupyter notebook, which would put the test code in the same "file" and avoid the import issue (also making it easier to associate the test/design code in one place).

